PR #1, opened 2012-08-29
Removed magic number

Triang3l:
Replaced magic number in server/util_md5.c with constant arithmetic.


Triang3l:
oh, it's a mirror

PR #2, opened 2012-09-09
Fixed IE10's DNT setting being ignored

OscarGodson:


mxriverlynn:
+1 

the arrogance and absurdity of Apache ignoring IE10 users is truly mind-boggling

zacparker:
+1

liquidboy:
+1

hizanberg:
-1 

Expecting a lot of small minded devs to vote +1 here, instead of petitioning a change to IE10's policy to abide by the intent of the standards.

OscarGodson:
@hizanberg Actually we all _like_ the way IE10 is handling it.

interscaperob-duplicate:
+1

Expecting many self-righteous devs to vote -1 here, because they don't like the way the standard is being interpreted by browser vendors, and think injecting code in a message execution engine is the right way to fix the problem. 

There is no possible way a company, for example, would never want to override the will of the individual users and not allow their employees to be tracked for advertising purposes. And the people that aren't idiots, that actually took the time to read whet the "Express" settings said and still chose to click it anyways... we should just throw their choice out the window, right?

hizanberg:
@OscarGodson Who cares what you think it should be? You have no more right to set the defaults preference of all IE10 users anymore than Microsoft does.

OscarGodson:
@hizanberg You realize thats why we're all upset right? Because Apache is making the choice for IE10 users...? Sort of an ironic comment you just made. Also, IE10 isn't setting a default. It's just a preselected value. The user ultimately has to make a choice to accept the preselected value or not. So, in this case, only Apache is trying to set a default for everyone on IE10. Not me, us, or Microsoft.

interscaperob-duplicate:
@hizanberg Your statement assumes that all IE10 users are morons, and that no one will read the Express settings, and that people just click next until prompted otherwise. Some will, but not all, and to throw away the choice for the people that were OK with the default and chose it, just because some people click "next" without understanding the consequences, is foolish. It's like throwing out the results of a Presidential Election because some people didn't vote the way you wanted them to.

hizanberg:
@OscarGodson yes you're ironically upset at a change of a default server setting in Apache (that doesn't do anything yet) to render's IE10's DNT meaningless since most of the times it wont have been a conscience user preference.

@interscaperob I'm assuming defaults will always get selected regardless of what they are, hence the user populace isn't making the choice the way it is. You're delusional if you think defaults aren't important or aren't a way for vendors to abuse for their own anti-competitive purposes. If you don't care what the defaults should be, you should be petitioning IE to change their policy so you are able to opt-in to DNT and it will actually get respected.

interscaperob-duplicate:
@hizanberg "render's IE10's DNT meaningless since most of the times it wont have been a conscience user preference."

This is based on an assumption of user behavior and not any actual empirical data. if there was no text on the setup page, that would be one thing. It is there, it is clear, and it satisfies the requirements. And the part that you fail to understand is that setting it the other way IS STILL A DEFAULT. It's just one that advertisers just happen to like.

hizanberg:
@interscaperob  An explicit choice is never made when its bundled as part of the Default settings.
For various reasons (i.e. being the easiest option) regardless of what the default options are, most users will go with the defaults, at which point the browser vendor is making the decision.

It doesn't even matter what text is on the setup page, since most users wont even read it. It needs to be on a stand-alone screen (or dialog) with only that question only and no default option pre-selected for it to be a conscience user decision. Otherwise going to user preferences and checking a checkbox works too, since at that point most users will be deciding for themselves then.

interscaperob-duplicate:
You understand that if the original value is DNT:0, that is still a default, right? It's a bit, and it is not nullable. It is either on or off. If it ships off... IT IS STILL A DEFAULT.

Correction: The header can not be sent. No header still equals off, which still equals "track me". Having the set value is still the result of an explicit user interaction. The spec does not say it has to be the only option on the page. It does however say that the header should not be interfered with by third parties.

covener:
When there's no preference expressed, there's no header. That's the third state.

imanavg:
@hizanberg - Please don't spread lies. It is proven over and over again that IE is actually compliant, since it is giving the option to user to make a choice i.e. whether choose express settings where it shows that DNT would be enabled or select customize to modify the settings. The user IS making the CHOICE.

So please don't spread FUD or try to create a political war due to your personal bias on the issue.

PS: Your point about user not reading the text is as stupid as it can get. The standard asks for an option to be provided and it is provided. It also lists that an implementation may bundle DNT as a part of other option such as privacy high and MS is doing that, so I fail to understand your point. Please keep your personal belief out of OSS, it would serve you well.

The text from the draft:
"For example, a user might select a check-box in their user agent's configuration, install a plug-in or extension that is specifically designed to add a tracking preference expression, or <B>make a choice for privacy that then implicitly includes a tracking preference (e.g., Privacy settings: high)</B>. Likewise, a user might install or configure a proxy to add the expression to their own outgoing requests. For each of these cases, we say that a tracking preference is enabled."

OscarGodson:
@mythz 

> There is no proof that IE is compliant (and no you saying so doesn't make it so).

Is this enough proof they're making it an option that is clearly marked? Not only that, you're free to change it at this time as well.


> the fact that the person who helped form the standards, is saying that they're not compliant should at least ring some warning bells

We don't care if they are or are not. That's not the issue. Although we all agree their definition of compliance agrees with Microsofts implementation, we have more of an issue that they're _overriding user's choice in IE10 as a form of "punishment" to Microsoft_. That's not how open source and open standards works and it's wrong. Open standards are not laws, that's the entire point of them. They're supposed to be followed and grown organically, not forcefully.

sschocke:
+1

As many have said, the web server is not the place for this. Also, the fact that it is basically hidden away as a default in Apache's configuration means those same idiots that don't read the "Express Settings" page will also use Apache with it's defaults on, and that damages the standard.

I have no problem with the idea behind this... Write a blog post or something about Microsoft's evil choice in your opinion... include in it a sample of how you can force DNT=0 for IE10 User Agents, and advertise it to everyone that will listen... Make forcing DNT=0 an _EXPLICIT_ sysadmins choice!

But then again, @royfielding, @hizanberg, etc are all too scared no sysadmin will actually do it, and user privacy may actually get respected... or at least, that's how it seems from my point of view.

EvilHom3r:
+1

gabranth:
@OscarGodson isn't that image showing that Microsoft is complying with the DNT draft as its not set its only turned on when the end users chooses with the express setting that clearly stats that it will be enabled or turned off with customized settings

BrazilianJoe:
+1.

I can see it has been shown that MS complies to the standard and provides the necessary choice for the user. It also presents it in a way that it is easier for the user to have less invasive settings, which is largely seen as a good thing by the consumers themselves.

As a user, sysadmin and web developer, working with Windows, Mac and Linux, I can't see how this patch saves any trouble from any part involved, much on the contrary. It needlessly meddles with the information sent by the browser. The amount of comments already shows that this was a bad move/commit, and politically motivated.

If MS had it set off by default, there would be screaming about how MS does not protect the users by 'hiding DNT in obscure custom settings'. Damned if you do, damned if you don't. All this Microsoft-bashing in open source is getting old already.

londospark:
IE 10 is compliant with the spec and should not be treated in any other manner. MS has made it clear that DNT will be switched on and has made it clear that one may opt out and how. I hope this change is committed.

Fever905:
+1
Grow a freaking brain wouldja?  And this was written by someone who works for Adobe whose clients and partners BENEFIT from the Do Not Track being disabled?  Huge conflict of interest here.  The choice is made in the express setup.  Its clearly marked there.

mamund:
Here is the key sentence in the draft:

_"The basic principle is that a tracking preference expression is only transmitted when it reflects a deliberate choice by the user."_[1]

Here are some possible ways to model this choice:
1. Tracking is allowed. Do you want to change this? yes( ) no ( ) (optional)
2. Tracking is now allowed. Do you want to change this? yes( ) no ( ) (optional)
3. Do you want to allow tracking? yes ( ) no ( ) (required)

Some here assert that granting a user the ability to skip taking an action ("We will set option Z to value X unless you tell us otherwise") meets the definition of "deliberate choice by the user." @royfielding (and others) do not see it the same way.

Until this point is cleared up, the work is stalled. My suggestion is to modify the document to define "deliberate choice by the user" in a way that removes the current ambiguity and still allows vendors freedom to craft their UX as they wish while knowing they are within the proposed standard.

[1] http://www.w3.org/2011/tracking-protection/drafts/tracking-dnt.html#determining

imanavg:
@mamund - You are focusing only on partial text. Here is the detailed text at the same link:

"<B>We do not specify how tracking preference choices are offered to the user or how the preference is enabled</B>: each implementation is responsible for determining the user experience by which a tracking preference is enabled. For example, a user might select a check-box in their user agent's configuration, install an extension or add-on that is specifically designed to add a tracking preference expression, or <B>make a choice for privacy that then implicitly includes a tracking preference (e.g., Privacy settings: high)</B>. The user-agent might ask the user for their preference during startup, perhaps on first use or after an update adds the tracking protection feature. Likewise, a user might install or configure a proxy to add the expression to their own outgoing requests."

kristiaanvandeneynde:
@imanavg You just quoted why you're wrong. They key words are "The user-agent", not "The operating system". Unless the screenshot that @OscarGodson showed above is shown in Internet Explorer 10 on first use, **Microsoft is ignoring the spec.**

**You cannot know** that the person who installed the OS on a computer will also be **the actual user of that pc.**
Therefore, the DNT settings set during OS install **cannot ever be trusted to be the actual user's preference**.

That being said, the code that was added to Apache to block out IE10 also disrespects the DNT standard.

In short:
Yes, IE10 should be ignored when DNT is set by the OS.
No, IE10 shouldn't be ignored as a whole. Find a better way to solve this.

unikitty37:
+1.

Apache has no way of knowing whether or not that DNT: 1 is set because somebody handed me a pre-installed system where they just clicked through the options, or whether I went into the settings and verified that Do Not Track was set as I expected.

Just as Google had no right to override my "no third-party cookies" setting on Safari just because it happened to be the default, Apache has no right to override my "do not track" setting on Internet Explorer just because _it_ happened to be the default.

imanavg:
@kristiaanvandeneynde - What are you talking about, how many users do you think have more than one user in case of PC? Yes for may be a few scenarios it would result in an administrator selected default, but that does not mean that 90% of users who have single account should be penalized especially since they made the choice to have DNT on.

This is highly irresponsible on the part of Apache foundation. Shame on them for abusing OSS for their personal agenda.

kristiaanvandeneynde:
@imanavg You'd be surprised how many people buy preconfigured computers or have their IT-savvy son/daughter/brother/dog configure it for them because they don't (want to) know the first thing about computers.

And do you have any factual data to back up your claim that 90% of the population configures its own pc? I think your numbers are way off...

ghost:
-1, until MS revert their code, this PR should not be merged.

ghost:
Seems you are all quite ignorant about how Roy arrived at the patch! It was discussed and agreed on by the powers that be in he Apache project. It was not a one-man-decision.

imanavg:
@drak - it was Roy's decision, a stupid one at best, not a cumulative one. The title of his commit proves his political stand. His patch and his politics has no place in OSS. And you keep your MS hatred to yourself, OSS is a better place without people filled with hatred similar to Islamic Jihadist.

ghost:
@imanavg it is you who is behaving like a extremist, and you who is transferring your hatred and whatever else onto those who disagree with you. I never said I hate MS, I in fact use it over linux for my desktop. However, their browser leaves a lot to be desired and in a free world, I am able to says so. In your world, evidently those who disagree with you are terrorists who deserve the sharp end of an Apache helicopter attack? I support Apache's decision, and I would ask you to respect that fact without slinging mud.

imanavg:
@drak - All your are doing is coming here and sharing your opinion without any knowledge of the subject or any logical points as to why you think the original patch was the right choice.

On the other hand, we have proven over and over again that IE is actually compliant, since it is giving the option to user to make a choice i.e. whether choose express settings where it shows that DNT would be enabled or select customize to modify the settings. The user IS making the CHOICE.

What Apache is doing is ignoring the choice I as a user would make by selecting Express Setup option and you support that? Ask yourself, would you have supported the same commit, if it was against Chrome or FF? Sorry, your comment history here shows your bias, no matter how objective you pretend to be.

The text from the draft:
"For example, a user might select a check-box in their user agent's configuration, install a plug-in or extension that is specifically designed to add a tracking preference expression, or <B>make a choice for privacy that then implicitly includes a tracking preference (e.g., Privacy settings: high)</B>. Likewise, a user might install or configure a proxy to add the expression to their own outgoing requests. For each of these cases, we say that a tracking preference is enabled."

ziyan:
+1

kristiaanvandeneynde:
@imanavg You haven't proven jack. Please stop patronizing people for expressing their opinion. Ad hominem attacks are the tool of those that have no real case to defend.

Because the IE10 DNT setting is set during Windows 8 install, you can **never** truly know whether a user intended to switch on DNT or not. Ergo Apache is making the right move here, albeit hypocritical. It's about sending a message.

If Microsoft's move is left unchecked, then what stops advertisers from simply not respecting DNT at all? After all, they could use IE10's settings to prove their right, saying: "We're willing to cooperate, but Microsoft forced our hand."

imanavg:
@kristiaanvandeneynde Yes right, go read the damn spec and then argue. If I install windows 8 and I click express setting, why is Apache ignoring it? This proves that it is Apache that is in violation not respecting my settings than anything else.

Stop being a hater in life and you would feel much better.

kristiaanvandeneynde:
@imanavg There's a reason your posts got deleted from https://github.com/apache/httpd/commit/a381ff35fa4d50a5f7b9f64300dfd98859dee8d0.

You're so full of yourself that you just insult anyone who disagrees with you.
There is a term for that: nerd rage. Please crawl back under the bridge from whence you came, troll.

I choose to no longer waste my time on you.
Blocking you and unfollowing this thread based on https://github.com/apache/httpd/commit/95b9a0fa04107515c2928cc844de50ad21ddc7bf.
Also reporting your comment in which you compare anyone who disagrees with you with jihadists.

PR #4, opened 2013-10-02
Add ws:// and wss:// as valid absolute URIs

cjh1:
Add case to is_absolute_uri(...) to match ws:// and
wss:// so that rewrite rules involving absolute websockets
URIs will not be treated as file paths.

See the following issue for more details:

https://issues.apache.org/bugzilla/show_bug.cgi?id=55598


cjh1:
Has been merged into trunk, Thanks!

PR #7, opened 2014-03-24
Use new OpenSSL methods for fixing broken chains when available

ekasper:
OpenSSL >= 1.0.2 has SSL_CTX_build_cert_chain for automatically building certificate chains. Calling this with SSL_BUILD_CHAIN_FLAG_CHECK | SSL_BUILD_CHAIN_FLAG_NO_ROOT will catch and fix some obvious certificate chain configuration errors such as misordered chains or chains that include the self-signed root.

If the certificate is self-signed or building a chain from the provided certificates fails for some other reason, this operation is a no-op, i.e., the original configuration will be preserved.

Note this patch only applies to chains configured via SSLOpenSSLConfCmd or SSLCertificateFile, and does nothing for the legacy SSLCertificateChainFile configuration.


pquerna (changes requested):
This needs to be rebased in order to be applied

PR #9, opened 2014-04-08
Revert "Fix our documentation to not show the expression in Require expr...

ghost:
... as quoted."

This reverts commit 407db4e9f389cc7317b065d7e141ea9c8021df06.


rbowen:
I'm very unclear on what you're attempting to accomplish with this PR. And, yes, I know this was a very, very long time ago. But do you recall what problem you were trying to solve here?

PR #14, opened 2015-05-18
Now tracking the response_codes for gnuplot output.

edyesed:


support/ab.c
```diff
@@ -1364,6 +1367,8 @@ static void close_connection(struct connection * c)
             s->ctime     = ap_max(0, c->connect - c->start);
             s->time      = ap_max(0, c->done - c->start);
             s->waittime  = ap_max(0, c->beginread - c->endwrite);
+            // Now save the connection response code over to the stats[i] response code.
+            memcpy(s->response_code, c->c_response_code, 4);
```
mdw123, line 1371: If this is a null-terminated string, you should use strncpy rather than memcpy (weird to use memcpy on null-terminated strings). Also, what is the true size of c->c_response_code? Is it 4 bytes or 4 characters? How big is a character?

edyesed:
closing for inactivity

PR #15, opened 2015-08-03
Enable TCP_FASTOPEN, if available.

ghost:
Note that it doesn't affect clients that do not support it.

### DIFF
--- a/modules/generators/mod_cgid.c
+++ b/modules/generators/mod_cgid.c
@@ -606,6 +606,7 @@ static int cgid_server(void *data)
     server_rec *main_server = data;
     apr_hash_t *script_hash = apr_hash_make(pcgi);
     apr_status_t rv;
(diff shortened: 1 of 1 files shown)


jimjag:
Where is that define enabled or checked for?

ghost:
/usr/include/netinet/tcp.h:#define TCP_FASTOPEN      23 /\* Enable FastOpen on listeners */

jimjag:
I would be best, I think, to check for this at build-time ala autoconf and use the normal HAVE_*. This also requires that APR be aware of it as well, to really be complete.

candrews:
Can this please be reviewed/merged?

candrews:
Reported in bugzilla at https://bz.apache.org/bugzilla/show_bug.cgi?id=63102

HOSTED-POWER:
Is it still not implemented? Would be great if it was, very easy to fix apparently :| :)

PR #16, opened 2015-08-03
Enable TCP_FASTOPEN, if available.

ghost:
Note that this does not affect clients that do not have this functionality.
Second note: failure is also harmless. We could log something but this will annoy sysops with an apache compiled for a more recent glibc and an older kernel.


candrews:
Can this please be reviewed/merged?

candrews:
Reported in bugzilla at https://bz.apache.org/bugzilla/show_bug.cgi?id=63102

PR #18, opened 2016-01-12
wrong return value while create a new ticket

oknet:
from openssl online document: https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_tlsext_ticket_key_cb.html

The return value of the cb function is used by OpenSSL to determine what further processing will occur. The following return values have meaning:

2
This indicates that the ctx and hctx have been set and the session can continue on those parameters. Additionally it indicates that the session ticket is in a renewal period and should be replaced. The OpenSSL library will call cb again with an enc argument of 1 to set the new ticket (see RFC5077 3.3 paragraph 2).

1
This indicates that the ctx and hctx have been set and the session can continue on those parameters.

0
This indicates that it was not possible to set/retrieve a session ticket and the SSL/TLS session will continue by by negotiating a set of cryptographic parameters or using the alternate SSL/TLS resumption mechanism, session ids.

If called with enc equal to 0 the library will call the cb again to get a new set of parameters.

less than 0
This indicates an error.

### DIFF
--- a/modules/ssl/ssl_engine_kernel.c
+++ b/modules/ssl/ssl_engine_kernel.c
@@ -2284,7 +2284,7 @@ int ssl_callback_SessionTicket(SSL *ssl,
                       "TLS session ticket key for %s successfully set, "
                       "creating new session ticket", sc->vhost_id);
 
-        return 0;
+        return 1;
     }
     else if (mode == 0) {
         /* 


oknet:
parent link to https://github.com/apache/trafficserver/pull/400

ylavic:
Committed in https://svn.apache.org/r1736186

Please note that this change has no real effect with the current OpenSSL (and forks) usage, they check only for >= 0 return value when requesting for a new ticket (parameter enc is 1).

See:
https://github.com/openssl/openssl/blob/master/ssl/statem/statem_srvr.c#L3017
https://github.com/libressl-portable/openbsd/blob/master/src/lib/libssl/src/ssl/s3_srvr.c#L2520
https://boringssl.googlesource.com/boringssl/+/master/ssl/s3_srvr.c#2046

Thanks for the report, returning 1 is probably the right thing to do should the SSL libs later change their check on the return value.

oknet:
you are welcome.

I'm a ATS code reader, just follow the OpenSSL API document to understand
the SSLNETVConnection, and found the `bug' in ATS code.
bryancall ( a committer of ATS ) tell me the code same as the apache httpd
project.

Thanks for your analysis in detail.

2016-03-22 21:45 GMT+08:00 ylavic notifications@github.com:

puiterwijk:
Please note that after https://github.com/openssl/openssl/commit/5c753de668322bf9903a49ba713b2cbc62667571 got merged into OpenSSL, it no longer just checks for >=0, but the ==0 is actually enforced as documented.
So any OpenSSL with that patch (1.1.0 and higher) will just not inject a ticket key without this return value fix.

ylavic:
Thanks for the note, the patch is currently being proposed for backport to the next 2.4.x, one more vote and it's in.

ylavic:
Backported in https://svn.apache.org/r1811742, will be part of 2.4.29.

PR #19, opened 2016-02-26
Fix QualifyRedirectURL explanation text

tjikkun:
Text was copied from the line above, should have been changed


ylavic:
Thanks, fixed in http://svn.apache.org/r1734835.

PR #20, opened 2016-06-01
scheme support in server name directive

keywan-ghadami-oxid:
translated missing paragraph about scheme support in server name directive


docs/manual/mod/core.xml.de
```diff
@@ -2924,6 +2924,13 @@ selbst zu identifizieren</description>
     Hostname im <code>Host:</code>-Header der Anfrage auftauchen muss,
     damit sie diesem virtuellen Host zugeordnet wird.</p>
 
+    <p>Möglicherweise steht der Server hinter einem Gerät, dass SSL verarbeitet steht,
```
resmo: This sentence does not make sense, there is duplicate 'steht'.

PR #21, opened 2016-08-12
Followup to commit 64e605c2d2d81bd7288aa19e028287fae5ee998b

YoshikiHigo:
We inserted if statements to check return values of apr_get_table because the same change was conducted in a past commit.

https://github.com/apache/httpd/commit/64e605c2d2d81bd7288aa19e028287fae5ee998b#diff-5db53ddc1c86d78f9a3be70c49036281

Yoshiki, Shinpei, Hideaki, and Mei


jimjag:
The placement of the 'const char *ct' line is problematic, since it is a c99ism

YoshikiHigo:
Thanks for your feedback

PR #27, opened 2017-02-17
mod_status: extended text-mode

drzraf:
Currently, consuming extended server-status depends upon a HTML parser.
Most script/monitoring tools prefers a structured text input.
This introduce a ?short URL parameter which behave as the current ?auto:
It it shows stripped down information.
But using ?auto, the full extended status is shown in text-mode.

NB: some 3rd-party modules hooking mod_status many needs further adaptation.


drzraf:
Dear @asfgit , would you mind explaining yourself? Aren't you a bot?
If yes, then you too should concerned about having **extended** status be readable without bringing your *html-parsers* friends in.

PR #28, opened 2017-02-28
Do not hard-define empty env var

benrubson:
Hello,

This PR solves issue #59371 :
https://bz.apache.org/bugzilla/show_bug.cgi?id=59371

Thank you 👍 

Ben

### DIFF
--- a/docs/manual/env.xml
+++ b/docs/manual/env.xml
@@ -100,6 +100,13 @@
         module="mod_rewrite">RewriteRule</directive> which uses the
         <code>[E=...]</code> option to set environment variables.</p>
 
+        <p>A per-request variable name must not be found in the environment
+        variables defined before the server starts, otherwise the value from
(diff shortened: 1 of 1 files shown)


benrubson:
I Would also be very glad to see this backported to 2.4.x.
Many thx 👍

benrubson:
Hi,
Could we think about merging this ?
Thank you very much 👍 
Ben

covener:
I am not sure this is safe to backport, it might be reasonable for people to be using empty native environment variables, e.g. when things are deactivated.  It seems better to address the overlap between the different kinds of interpolation

benrubson:
Thank you for your feedback Eric 👍 
Mmmmh you think that some may intentionally use `MYVAR=` "instead" of just running `unset MYVAR`.
Regarding the overlap issue you are talking about, to be honest, I'm not sure I'll be able to correct it :|

PR #29, opened 2017-03-22
#60899 - new private utility function to return server's SSL_CTX object

ghost:
* there exist valid use case for custome modules to inspect the
    SSL_CTX for security compliance reasons
  * accessing the private structures is dangerous as they could change
  * user beware -- callers are responsable for knowing what ssl library
    the ctx is from and responding appropriately


notroj:
The SSL_CTX can be accessed through the hooks exposed by mod_ssl_openssl.h, which should satisfy this requirement.

```
APR_DECLARE_EXTERNAL_HOOK(ssl, SSL, int, init_server,
                          (server_rec *s, apr_pool_t *p, int is_proxy, SSL_CTX *ctx))
```

notroj:
Per comment above, this should be possible already.

PR #30, opened 2017-03-22
PVS-Studio: fixed weaknesses

VasilievSerg:
We have found and fixed weaknesses CWE-476 (NULL Pointer Dereference) and CWE-570 (Expression is Always True) using PVS-Studio tool.

Analyzer warnings: [V528](https://www.viva64.com/en/w/V528/), [V595](https://www.viva64.com/en/w/V595/), [V560](https://www.viva64.com/en/w/V560/), [V581](https://www.viva64.com/en/w/V581/).

PVS-Studio is a static code analyzer for C, C++ and C#.

This PR is related to the [issue #60903](https://bz.apache.org/bugzilla/show_bug.cgi?id=60903). I also mentioned there one more warning of PVS-Studio that wasn't fixed, but it seemed important to me. 

Thanks.

### DIFF
--- a/modules/generators/mod_cgid.c
+++ b/modules/generators/mod_cgid.c
@@ -1194,9 +1194,7 @@ static int log_script(request_rec *r, cgid_server_conf * conf, int ret,
                 apr_file_puts(argsbuffer, f);
             apr_file_puts("\n", f);
         }
-    }
-
-    if (script_err) {
+        
         apr_file_close(script_err);
     }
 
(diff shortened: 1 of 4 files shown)


notroj:
These all look fixed or no longer relevant for current trunk.  Are you able to retest with current code?

VasilievSerg:
Hi @notroj,

I've used the PVS-Studio static analyzer for the searching of suspicious code fragments. I suggest that you also try it for finding weaknesses / potential errors. I think if you check the project yourself it will be better than if I do it myself since you know the project better and as a result, you can select more useful warnings from the analyzer output. :)

Here is the license key:

>**Name**: apache-httpd
**Key**: PX7G-4YYY-1U9G-XR84
License type: Team
License Valid Thru: 4 Feb 2021

If you have questions about the analyzer, please, don't hesitate to ask. :)

notroj:
Thanks for the offer!  I don't have time for this, so you could offer it to the mailing list:

https://httpd.apache.org/lists.html#http-dev

I'm going to close this out since all the issues appear to be fixed or no longer relevant.

PR #33, opened 2017-04-25
Add host header override for proxy health check 2.5.x

manusfreedom:
Not able to test currently, but can be useful with backends using host-header.


jimjag:
I am curious regarding which use cases would require something like this?

manusfreedom:
Architecture:
LVS -> Apache (Reverse proxy + balancer) -> IIS (can be another app like Apache with VirtualHost) using exclusively host header and only one IP.
Adding this feature you can have some apps load balanced on the same IIS farm.
And health check each app independently.

PR #43, opened 2017-12-22
[AB] Replace 'HTML transferred' by 'Body transferred'

lyrixx:
Because the server can return something different from HTML.
Like JSON


lyrixx:
Is this the right place to contribute ? /cc @ylavic

ylavic:
The best place to contribute on httpd is the development mailing list (see http://httpd.apache.org/lists.html#http-dev).
The team barely monitors pull requests on gh (repository mirrored only), and should probably more, thus patches posted on the mailing list or httpd's bugzilla (improvement/fixes) are more likely to be considered (and quickly too).

lyrixx:
Thanks

PR #48, opened 2018-03-14
* modules/filters/mod_xml2enc.c (xml2enc_ffunc): do not remove error buckets

pemi66:
Error bucket is a special kind of meta bucket that is not part of apr but added in httpd and can be created via the ap_bucket_error_create() call.  This type of bucket is simply destroyed by mod_xml2enc.


bombadil:
I worked with @pemi66 on this issue. The impact of the error bucket getting dropped is that any failure raised by another filter earlier in the output filter chain will get dropped and thus not handled at all.

bombadil (approved):

bombadil:
See also https://bz.apache.org/bugzilla/show_bug.cgi?id=62180

PR #49, opened 2018-04-10
ASF-54339 interleaved logs

reed-sandberg:
Under high load, log output showed interleaving of log
records. When writing to a pipe, any message
greater than PIPE_BUF may interleave with concurrent writers, any
message PIPE_BUF or less is guaranteed to be atomic. On Linux, this
is typically 4K.

* Patch Apache HTTPD Server to break up log messages > PIPE_BUF into
  smaller chunks that are guaranteed to remain intact with multiple
  writers when configured with '|<' (mnemonic, split pipe, rather than
  just '|' for piped log output).
* Patch rotatelogs program to reassemble the chunks.


reed-sandberg:
BZ ref:
https://bz.apache.org/bugzilla/show_bug.cgi?id=54339

PR #51, opened 2018-05-03
Make index file more beautiful

TravorLZH:
The Apache's index page has been ugly for ten years. It's time to make it more HTML5

I use responsive layout in this index file which makes it look more beautiful on phones.

### DIFF
--- a/docs/docroot/index.html
+++ b/docs/docroot/index.html
@@ -1,1 +1,11 @@
-<html><body><h1>It works!</h1></body></html>
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0"/>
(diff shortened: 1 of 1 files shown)


docs/docroot/index.html
```diff
@@ -1 +1,11 @@
-<html><body><h1>It works!</h1></body></html>
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+		<title>Welcome to Apache Server 2.4</title>
```
bombadil, line 5: If you add an opening `<head>` tag, please also add a closing one.

bombadil (changes requested):

docs/docroot/index.html
TravorLZH: DONE

PR #52, opened 2018-06-02
Add Option to generate multiple error logs of different format in Apache 2.4.x

europa502:
Just as there is an option to generate multiple log files of different format, it would be very much convenient if multiple error log files could be made with different format rather then declaring a global ErrorLogFormat.


bombadil:
With more than 250 commits this pull request is not reviewable.

I guess you accidentally picked the wrong target branch. Would you mind to update the target branch of your pull request to branch 2.4.x? Maybe that fixes it.

PR #60, opened 2019-07-22
Fix to log ssl_cipher with mod_log_json

eric-therond:
Hello
a fix to allow the logging of SSL_CIPHER variable (instead of SSL_COPHER)

Regards


tititiou36:
Hi,

This has been fixed in trunk (r1864834). Thanks a lot for spotting and reporting it.

PR #62, opened 2019-08-15
Add missing closing parenthesis in error message

usev6:


usev6:
For the record: Since this has been closed, but I don't see the patch in the SVN repo, I've opened a bug report in the ASF Bugzilla: https://bz.apache.org/bugzilla/show_bug.cgi?id=63806

PR #69, opened 2019-10-27
atomic creation of pid file.

ncarrier:
The original pattern for creating the pid file was:
open_create(pid_file)
write(pid_file, pid)
close(pid_file)

But if a power outage occurs between open_create and write, the file will
be empty and httpd will refuse to start afterwards unless the corrupt pid
file is removed.

This patch uses the pattern:
open_create(temp_pid_file)
write(temp_pid_file)
close(temp_pid_file)
rename(temp_pid_file, pid_file)
which is guaranteed to be atomic, provided that temp_pid_file and pid_file
are located in the same file system, which this patch does by creating
a temporary file name with the pattern:
    pid_file_name + random_suffix

Signed-off-by: Nicolas Carrier <nicolas.carrier@orolia.com>

### DIFF
--- a/server/log.c
+++ b/server/log.c
@@ -1524,6 +1524,8 @@ AP_DECLARE(void) ap_log_pid(apr_pool_t *p, const char *filename)
     pid_t mypid;
     apr_status_t rv;
     const char *fname;
+    char *temp_fname;
+    apr_fileperms_t perms;
 
     if (!filename) {
         return;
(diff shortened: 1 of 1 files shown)


ncarrier:
For servers, it may be acceptable to ask for the user to remove the offending corrupt pid file, but in embedded products, it's not an option.

This patch has been submitted to httpd's bugzilla in February but didn't get a response since then: https://bz.apache.org/bugzilla/show_bug.cgi?id=63140

notroj:
Looks like a good idea, though I think it should catch errors returned by apr_file_rename and apr_file_perms_set.

PR #72, opened 2019-11-11
server/utils: fix Bison directives

akimd:
Non terminal cannot have string aliases (only tokens appear in error
messages).  %token is used to define tokens, and %nterm non terminals.
The hidden %type (which was only recently documented) is meant for
both tokens and non terminals.  Yet

    %type <foo> expr "expression"

is actually more or less equivalent to

    %nterm <foo> expr
    %token <foo> "expression"

which is clearly not the intention of the author here.

* server/util_expr_parse.y: Remove useless string-literal only tokens.
Prefer %nterm to %type to avoid this error.

### DIFF
--- a/server/util_expr_parse.y
+++ b/server/util_expr_parse.y
@@ -98,22 +98,22 @@
 %right  T_OP_NOT
 %right  T_OP_CONCAT
 
-%type   <exVal>   cond              "condition"
-%type   <exVal>   comp              "comparison"
-%type   <exVal>   strfunc           "string function"
-%type   <exVal>   listfunc          "list function"
-%type   <exVal>   list              "list"
-%type   <exVal>   words             "words"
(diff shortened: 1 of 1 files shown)


notroj:
This didn't get tested properly in Travis because buildconf suppresses regeneration of the parser sources by default.  I fixed this in 2dd0046f300cf667d8a473d51e1c097f2715efa4 so if you rebase to current trunk you'll get one CI job testing it properly.

akimd:
Hi @notroj 
Thanks for the help!  I updated the patch, and it seems to pass.
Cheers!

notroj:
Thanks for the contribution @akimd!  I've merged this to trunk, but I'll have to ask you to close the PR since I can't do it myself.

akimd:
Thanks!

PR #73, opened 2019-11-13
Also avoid adding the Set-Cookie header in both r->headers_out and

uhliarik:
r->err_headers_out in ap_cookie_remove and ap_cookie_remove2 functions
to avoid duplication in HTTP response.


uhliarik:
PR 60910

PR #77, opened 2019-11-22
mod_proxy: fix proxy connection cleanup from an n+2 pool.

rpluem:
When connection_destructor() is called after pchild is gone, we can't
dereference worker->cp anymore. This happens is debug/one_process mode only,
if we exit by calling apr_terminate() or clearing the process pool directly.

Fix this by NULL-ing worker->cp in conn_pool_cleanup(), and by registering it
as a pre_cleanup.

git-svn-id: https://svn.apache.org/repos/asf/httpd/httpd/trunk@1822531 13f79535-47bb-0310-9956-ffa450edef68


rpluem:
Merged as 484abff

PR #78, opened 2019-11-22
Delay some memory allocation.

rpluem:
If this handler will not handle the request, no need to waste bytes in the request pool.

git-svn-id: https://svn.apache.org/repos/asf/httpd/httpd/trunk@1829676 13f79535-47bb-0310-9956-ffa450edef68


rpluem:
Merged as 484abff

PR #79, opened 2019-11-22
63503 backport step3

rpluem:
Add error messages and return bad request.

fix incorrect rv. Sorry.

Follow up to r1847232.
There is no point to use "old" numbers in recent commit.

Also avoid number duplication. The messages are the same but in different code path, so having different numbers makes sense.
This also avoids a warning when running:
   make update-log-msg-tags


rpluem:
Merged as 484abff

PR #84, opened 2019-12-11
Replace zeustech.net with version from web archive

schmijos:
Following the link <httpx://zeustech.net> brings you to a website looking like this:

2019-12-11 at 10 58

I don't want to expect from our users to endure this offence.


covener:
Thank you Josua, committed after making sure ASF legal was okay with it.

PR #88, opened 2020-01-09
Ensure that we always pass on all we have in case we see an EOR bucket

rpluem:


rpluem:
Obsoleted by b3110d36afaceab45e083a68b1a8493b8244af2a.

PR #91, opened 2020-01-14
Backport systemd support.

notroj:


notroj:
Fixed in http://svn.apache.org/viewvc?rev=1873519&view=rev

PR #93, opened 2020-01-25
Support for User: header in draft-vanrein-http-unauth-user

vanrein:
According to draft-vanrein-httpd-unauth-user
Offers HTTP support for https://john@example.com
Implemented as equivalent to https://example.com/~john

### DIFF
--- a/docs/manual/env.xml
+++ b/docs/manual/env.xml
@@ -120,6 +120,13 @@
 	the <a href="http://www.ietf.org/rfc/rfc3875">CGI
         specification</a>.</p>
 
+	<p>On top of this, when a <code>User</code> header is included in the
+	request on top of the resource specification provided by <code>Host</code>,
+	then CGI and SSI scripts will see <code>LOCAL_USER</code> set to the
+	unescaped form of this header value.  This header is specified
+	in <code>draft-vanrein-http-unauth-user</code> and is deliberately not
+	the result of any authentication process.</p>
+
     </section>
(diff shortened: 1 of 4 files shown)


covener:
Wouldn't this need to add a Vary response header?

vanrein:
> Wouldn't this need to add a Vary response header?

You are right of course!  I actually wrote it in my draft...

   HTTP caches [RFC7234] need to distinguish requests with different
   User header values.  The Vary header [Section 7.1.4 of [RFC7231]]
   MUST be present in the matching response, and the header MUST either
   be a single "*" star (U+002a) or list the "user" name, for all
   responses whose processing was influenced by the User header.  This
   requirement does not apply to software and configurations that ignore
   the User header.

...but overlooked it in the Userdir.  Will fix!

vanrein:
This adds a token `User` to the `Vary` header, precisely when the `User` header is used.  Thank you for pointing out the oversight; I  was thinking of the equivalent `/~username` form in `mod_userdir.c` and forgot about it.

vanrein:
That's funny, a check failed on a missing package during apt-get on one platform.

I'll upload documentation next, so that'll trigger another round anyway.

vanrein:
> [covener] CGI would already be providing HTTP_USER implicitly. Is there a reason for an addl var?
> [vanrein] Removing URL-style escapes.

The escapes are used to carry UTF-8 content, which is not possible in plain HTTP header syntax.

PR #95, opened 2020-02-06
Backport: Arrange the xml display and make it uses xml.

candux:
It is currently very difficult to use the xml interface of the balancer-manager because it needs a nonce even when you only want to view data. This is fixed in master but not in 2.4. This PR backports this change.
Bugreport: https://bz.apache.org/bugzilla/show_bug.cgi?id=63074
git-svn-id: https://svn.apache.org/repos/asf/httpd/httpd/trunk@1847295 13f79535-47bb-0310-9956-ffa450edef68

### DIFF
--- a/modules/proxy/mod_proxy_balancer.c
+++ b/modules/proxy/mod_proxy_balancer.c
@@ -1225,7 +1225,6 @@ static int balancer_handler(request_rec *r)
          )
         )
        ) {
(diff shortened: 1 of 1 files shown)


notroj:
The nonce is missing in the XML output (proposed for backport in #587 ) and the rest of this is wrong/obsolete.

PR #98, opened 2020-02-26
Fixed Content Type bug

chonchiog:
Fixes BZ bug https://bz.apache.org/bugzilla/show_bug.cgi?id=64140.


covener:
Merged in http://svn.apache.org/viewvc?view=revision&revision=1874616

notroj:
Fixed in trunk per above, thanks for the patch!

PR #101, opened 2020-03-16
Fix git ignore

chonchiog:
This gitignore file was generated from git-svn.


notroj:
Very nice, thank you @velazcorenier

PR #105, opened 2020-03-26
Add SSLProxyOpenSSLConfCmd: Equivalent of SSLOpenSSLConfCmd for proxy HTTPS connections

airtower-luna:
This patch is relevant for reverse proxy configurations with HTTPS connections to backend servers. To let my proxy connect to specific backend servers I need to allow ECC certificate authentication and ECDH using Brainpool curves. On the server side it is already possible to configure them using `SSLOpenSSLConfCmd`:
```
SSLOpenSSLConfCmd Curves brainpoolP384r1:brainpoolP256r1
```
However, with this setting on the backend server the reverse proxy could not connect, because the OpenSSL defaults do not allow Brainpool curves. This patch introduces `SSLProxyOpenSSLConfCmd` to allow the same thing for the client side of reverse proxy connections:
```
SSLProxyOpenSSLConfCmd Curves brainpoolP384r1:brainpoolP256r1
```
With this new setting the proxy connection works fine. Other somewhat unusual reverse proxy configurations might benefit from being able to use `SSL_CONF_cmd`, too.

I'd be happy to add documentation as well, should I just update `docs/manual/mod/mod_ssl.xml` for that?

If possible I'd like to get this patch into the 2.4.x series as well, please let me know if I should take any additional steps for that.


airtower-luna:
Pushed again after rebase on trunk.

airtower-luna:
Rebased on trunk again. Bugzilla issue is here: https://bz.apache.org/bugzilla/show_bug.cgi?id=64352

airtower-luna:
Rebased on trunk again, previous `.gitignore` changes became obsolete.

PR #108, opened 2020-04-07
libcurl dependency version fix for mod_md

xl32:
libcurl dependency version fix: mod_md builds and works fine against libcurl 7.29, so EL7 (CentOS 7, RHEL 7) is really supported.


notroj:
This was merged in 2a8260a069e7f73cadcfd71a9f593b0cf67269d1.  Thanks for the patch, can you close the PR?

PR #109, opened 2020-04-13
WIP : update icons for auto_index

sblondon:
This PR is a work-in-progress about bug report [#45525](https://bz.apache.org/bugzilla/show_bug.cgi?id=45525).

The updated icons come from material-design provided by Google under Apache licence (https://material.io).

You can see the modified icons by watching `icons/README.html`: only some .png files are modified. They can be compared the equivalent .gif files (which are not modified).

I have several questions:
 - are you interested to update the icons?
 - currently the size of the icons are not exactly the same (20\*22 for the older icons, 24\*24 for the newer ones). Is it ok or keeping the same size is mandatory?
 - I don't understand what it is the goal of the black background for the icons in the `README.html`. Has someone some info about it?
 - Materialdesign provides .svg format. I think it would be better to use .svg files instead of bitmap files. However, I guess it's not possible currently to protect backward compatibility. Is it possible for httpd 2.6 release?

Don't merge it now, it needs still work.


sblondon:
This is a preview showing the diffenrence between current icons and material icons. It's a partial screenshot of the `README.html`:
diff_icons

PR #110, opened 2020-04-15
WIP: Test for testing in containers

notroj:


elukey:
Niceee!!! Didn't know about Docker support!

notroj:
@elukey may be nice if I get it working.  Had the script running under podman locally and now I've made it worse I think.

PR #117, opened 2020-04-27
Move FILE buckets on setaside 

ylavic:
This is to prevent the original/underlying apr_file_t->filedes from being
reset to -1, invalidating all other shared buckets (e.g. splitted)
pointing to the same file.

Like for opaque buckets (->length == -1) which are also moved, we now
assume that FILE buckets have the appropriate lifetime, that is the one
of the current request (i.e. until EOR) or at least the one of the
connection. This is the case for all of httpd codebase (AFAICT).


ylavic:
After discussion on dev@, not an issue in the first place.
Case closed.

PR #123, opened 2020-05-06
Drop SSLRandomSeed & PRNG seeding with OpenSSL 1.1.1 and later.

notroj:
```
mod_ssl: Drop SSLRandomSeed implementation with OpenSSL 1.1.1.
Require that OpenSSL is configured with a suitable entropy source,
or fail startup otherwise.

* modules/ssl/ssl_private.h:
  Define MODSSL_USE_SSLRAND for OpenSSL < 1.1.1.
  (SSLModConfigRec): Only define pid, aRandSeed for <1.1.1.
  (ssl_rand_seed): Define as noop if !MODSSL_USE_SSLRAND.

* modules/ssl/ssl_engine_init.c (ssl_init_Module):
  Only initialize mc->pid for MODSSL_USE_SSLRAND.
  Fail if RAND_status() returns zero.
  (ssl_init_Child): Drop getpid and srand for !MODSSL_USE_SSLRAND.

* modules/ssl/ssl_engine_rand.c: ifdef-out for !MODSSL_USE_SSLRAND.
  (ssl_rand_seed): Drop warning if PRNG not seeded (now a startup
  error as above).
  
* modules/ssl/ssl_engine_config.c (ssl_config_global_create): Drop
  aRandSeed initialization.  (ssl_cmd_SSLRandomSeed): Log a warning if
  used w/!MODSSL_USE_SSLRAND.
```


peteroupc:
As shown in https://github.com/syslog-ng/syslog-ng/issues/3124, `RAND_status` can block indefinitely, at least in early boot (see also the pull request https://github.com/syslog-ng/syslog-ng/pull/3271).  It would be good if a future version of OpenSSL had a timeout mechanism for `RAND_status` to avoid indefinite blocking, but in the meantime, see the pull request linked to for one possible workaround.

ylavic:
The best httpd can do if entropy is missing is to wait, IMHO.
Embbedded devices should either not run software that need cryptography (at early boot), or provide a hardware to generate entropy..

notroj:
I would be fine with blocking as well, but it does not look correct to me that RAND_status() can block with the default ENGINE in OpenSSL 1.1.1. I can't see mention of the OpenSSL version in the linked report.  This is what the default RAND_status() does:

https://github.com/openssl/openssl/blob/master/crypto/rand/drbg_lib.c#L1349

ylavic:
I suppose that RAND_status()::rand_drbg_lock(drbg) can block if some other RAND_() method is being called (and blocking) in an other thread, but it's not the case during httpd init anyway.

PR #124, opened 2020-05-07
mod_ldap: Use the LDAP API directly to implement the rebind callback

notroj:
```
mod_ldap: Use the LDAP API directly to implement the rebind callback
for modern versions of OpenLDAP, avoiding the overhead of the apr-util
implementation.

* modules/ldap/util_ldap.c:
  Define USE_APR_LDAP_REBIND if a modern version of OpenSSL is used.
  (uldap_rebind_proc): New function.
  (uldap_rebind_init, uldap_rebind_add): Define, using either the
  callback or the (bad) APR-util versions.
  (uldap_connection_unbind): Clear the rebind pool to remove rebind
  references prior to destroying the LDAP *.
  Omit for !USE_APR_LDAP_REBIND.
  (uldap_connection_init): Use new wrappers, only create the rebind
  pool if USE_APR_LDAP_REBIND.

* include/util_ldap.h: Don't include apr_ldap_rebind.h here.

PR: 64414
```

### DIFF
--- a/include/util_ldap.h
+++ b/include/util_ldap.h
@@ -32,7 +32,6 @@
 #if APR_MAJOR_VERSION < 2
 /* The LDAP API is currently only present in APR 1.x */
 #include "apr_ldap.h"
-#include "apr_ldap_rebind.h"
 #else
 #define APR_HAS_LDAP 0
 #endif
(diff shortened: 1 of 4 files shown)


covener:
I can no longer get referrals to happen on my system but +1

notroj:
Do you mean they have broken, or you can't get the test setup configured @covener? I am going to try with a referrals config here.

covener:
Sorry not broken, my hack test  LDAP server with a referral shoved into the search tree no longer "simulates MSAD" like it used to so no referral goes on the wire.

PR #125, opened 2020-05-10
mod_proxy_http: handle Upgrade request and Switching Protocol

ylavic:
If the request Upgrade header matches the worker parameter "upgrade" and
the backend switches the protocol, do the tunneling in mod_proxy_http.
This allows to keep the protocol to HTTP until the backend really
switches the protocol, to apply usual output filters.


modules/proxy/mod_proxy_http.c
```diff
@@ -1615,6 +1649,62 @@ int ap_proxy_http_process_response(proxy_http_req_t *req)
...
+            /* Set timeout to the lowest configured for client or backend */
+            apr_socket_timeout_get(backend->sock, &backend_timeout);
+            apr_socket_timeout_get(ap_get_conn_socket(c), &client_timeout);
+            if (backend_timeout >= 0 && backend_timeout < client_timeout) {
+                tunnel->timeout = backend_timeout;
+            }
+            else {
+                tunnel->timeout = client_timeout;
+            }
+
```
rpluem, line 1686: Why do we need to align the timeouts and and have a procedure different from mod_proxy_wstunnel? This can be confusing.

modules/proxy/mod_proxy_http.c
```diff
@@ -1615,6 +1649,62 @@ int ap_proxy_http_process_response(proxy_http_req_t *req)
...
+            /* Let proxy tunnel forward everything */
+            status = ap_proxy_tunnel_run(tunnel);
+            if (ap_is_HTTP_ERROR(status)) {
```
rpluem, line 1694: The procedure here looks different from the one in mod_proxy_ws_tunnel:

1. We don't support async MPM's
2. We do not remove the reqtimeout filter

modules/proxy/mod_proxy_http.c
ylavic: In mod_proxy_wstunnel there is also a single (idle_)timeout for the tunneling part, this is how the ap_proxy_tunnel loop works (I don't it makes sense to have two separate timeouts for client and backend once it's not HTTP anymore).

Whether we should use the min, max, always use the backend's, or have a dedicated idle timeout is an open question, though. I'd say the backend's, maybe?

modules/proxy/mod_proxy_http.c
ylavic: 1. Async is coming next (probably a separate PR).
This would not be only for the tunneling part, we can go async for HTTP too when/if waiting for client (body) or backend data.
I need to find the right interface in proxy_util (based on wstunnel pump mechanism) for any proxy module to be able to use it.

2. An oversight, will do it now.

ylavic: Point 2. addressed in 5228c13.

modules/proxy/mod_proxy_http.c
rpluem: I saw that a single timeout is used in `ap_proxy_tunnel` and this is not the point I wanted to tackle. I was more worried that people get confused timeout wise with the setting done by `ProxyWebsocketIdleTimeout` which is `mod_proxy_wstunnel` related.  But this is not backported yet and probably before it is backported / 2.5[6] is released we have all that in `mod_proxy_http` such there is no need for having `ProxyWebsocketIdleTimeout`  ever in a stable release in its current form.
Once we have all that in `mod_proxy_http` we can think of how to handle it with the same parameters (approach) we currently use for timeouts. I think we already use two timeouts in `mod_proxy_wstunnel`in trunk. One for idle timeouts and one (or two different ones for client and server) for blocking timeouts.

ylavic: The plan would indeed be to deprecate proxy_wstunnel and have the upgrades handled by proxy_http, so `ProxyWebsocketIdleTimeout` would become ProxyAsyncTimeout and for the HTTP protocol we'd keep the usual `Timeout` (client side) and `ProxyTimeout` / worker `timeout=` on the origin side (which are the ones used by proxy_wstunnel too until the upgrade).

I'm on the async side of things now for this to possible ;)
Though as you said this is trunk only today in proxy_wstunnel.

covener:
Curious, would you keep the current async method or move the work to a filter?

ylavic:
> Curious, would you keep the current async method or move the work to a filter?

Still the MPM even callback interface like wstunnel, the filters' should_yield() and co would help determine whether we need to go async only.

minfrin:
Backported to v2.4.47.

PR #131, opened 2020-07-07
Check and use gettid() directly with glibc 2.30+.

notroj:
```
* configure.in: Check for gettid() and define HAVE_SYS_GETTID if
  gettid() is only usable via syscall().

* server/log.c (log_tid): Use gettid() directly if available.
```

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -535,22 +535,26 @@ prctl \
 timegm \
 getpgid \
 fopen64 \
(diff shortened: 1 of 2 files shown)


notroj:
Fixed in 33d99cf67f4bbdce86be0a0fdbb3d26f18e0f2a7

PR #133, opened 2020-07-10
In Lua 5.4, there is new lua_resume function definition, which has 4 …

uhliarik:
…params

now (previously had only 3). Pass dummy integer variable as 4th param, if there
is function definition with 4 params.

I didn't have opportunity to run test suite on it, so feel free to review it.

### DIFF
--- a/modules/lua/mod_lua.c
+++ b/modules/lua/mod_lua.c
@@ -342,7 +342,7 @@ static apr_status_t lua_setup_filter_ctx(ap_filter_t* f, request_rec* r, lua_fil
 {
     apr_pool_t *pool;
     ap_lua_vm_spec *spec;
-    int n, rc;
+    int n, rc, nres;
     lua_State *L;
     lua_filter_ctx *ctx;    
     ap_lua_server_cfg *server_cfg = ap_get_module_config(r->server->module_config,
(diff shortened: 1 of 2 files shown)


wrowe:
Something important to note, the logic in mod_lua confused me... we are ignoring the 2nd arg altogether. I'm not clear why this choice was made since the code and the lua docs don't explain. Given that discrepancy, I'm not entirely sure this works in practice.

Second important thing to note, we wouldn't know, because the test suite doesn't exercise lua_resume.

The patch certainly builds, so the PR is certainly interesting, thanks for submitting!

notroj:
Yeah, I also couldn't really understand what the new argument is doing or whether it is really needed here. :(

rainerjung:
From http://www.lua.org/manual/5.4/manual.html#8.3:

The function lua_resume has an extra parameter. This out parameter returns the number of values on the top of the stack that were yielded or returned by the coroutine. (In previous versions, those values were the entire stack.) 

I don't know, what this means for mod_lua, just wanted to provide a link to the param explanation.

notroj:
So the lua filtering API uses a "yield" from the filter code to pass back bucket data.  It seems to assume currently that after a yield  exactly one item is present on the stack, e.g. here https://github.com/apache/httpd/blob/trunk/modules/lua/mod_lua.c#L495

I guess there is some way to determine the stack size with <5.4 but haven't checked.   In #134 I've taken the patch from @uhliarik and added checking that the stack size is exactly one using the 5.4 api.  I think this makes sense but not very familiar with the code.

Also added a test for the output filters in http://svn.apache.org/viewvc?view=revision&revision=1880665 to see if this breaks anything for <5.4 as tested.

PR #136, opened 2020-08-21
Process early errors via a dummy HTTP/1.1 request as well

rpluem:


rpluem:
PR against wrong branch.

PR #144, opened 2020-10-12
Add .opus extension for audio/ogg mimetype, per RFC 7845

rigrig:
RFC 7845: https://tools.ietf.org/html/rfc7845#section-9
> An "Ogg Opus file" consists of one or more sequentially multiplexed
> segments, each containing exactly one Ogg Opus stream.  The
> RECOMMENDED mime-type for Ogg Opus files is "audio/ogg".

> The RECOMMENDED filename extension for Ogg Opus files is '.opus'.

IANA template: https://www.iana.org/assignments/media-types/audio/ogg


notroj:
Thanks for the patch, merged.

PR #148, opened 2020-11-19
HTTP_IN filter to handle 100 continue in blocking mode only

ylavic:
Make HTTP_IN filter send 100 continue in blocking mode only.

When mod_proxy_http prefetches input data it calls the HTTP_IN filter
in nonblocking mode, but since it does not want 100 continue to be sent
for every case (e.g. 100-continue forwarding), it hacks r->expecting_100
(save in req->expecting_100, reset, eventually restore..) all over the
place.

Let's avoid this by making the HTTP_IN filter send 100 continue only
when called in blocking mode (once still), instead of the first time
it's called.

* modules/http/http_filters.c (struct http_filter_ctx): Add the seen_data
  bit and rename eos_sent to at_eos (HTTP_IN does not send any EOS).

* modules/http/http_filters.c (ap_http_filter): Move 100 continue
  handling outside the initialization/once block, and do it in blocking
  mode only. Track in ctx->seen_data whether some data were already
  received, and if so don't send 100 continue per RFC 7231 5.1.1.

* modules/proxy/mod_proxy_http.c: Remove req->expecting_100 (and its
  danse with r->expecting_100) now that reading from the input filters
  does the right thing.

### DIFF
--- a/modules/http/http_filters.c
+++ b/modules/http/http_filters.c
@@ -407,43 +407,41 @@ apr_status_t ap_http_filter(ap_filter_t *f, apr_bucket_brigade *b,
         }
     }
 
-    {
-        /* Since we're about to read data, send 100-Continue if needed.
-         * Only valid on chunked and C-L bodies where the C-L is > 0.
-         *
-         * If the read is to be nonblocking though, the caller may not want to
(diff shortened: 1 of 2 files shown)


ylavic:
Merged in r1883639 and r1883640.

PR #150, opened 2020-12-02
mod_xml2enc: avoid corrupting Microsoft Office documents

jogu:
The Microsoft OOXML format uses xml packaged into a zip file, and has
mimetypes like:

application/vnd.openxmlformats-officedocument.spreadsheetml.sheet

This mimetypes contains 'xml', but is unfortunately not an xml file.

xml2enc processes these files (in particular, when mod_proxy_html is
used), typically resulting in them being corrupted as it seems to
attempt to perform a ISO-8859-1 to UTF-8 conversion on them.

Update the check to look for '+xml', which hopefully reflects only
mimetypes that actually contain xml.

closes https://bz.apache.org/bugzilla/show_bug.cgi?id=64339


notroj:
Thanks for the contribution!

PR #153, opened 2020-12-07
Fix misleading terminology

mpreziuso:
As described in https://bz.apache.org/bugzilla/show_bug.cgi?id=58798, the terminology used here is misleading: hashing (one-way algorithm) is not a form of encryption (two-way algorithm).


ylavic:
Committed to httpd's svn (http://svn.apache.org/r1884207) with some further changes in the dbmmanage[.in] script itself (the .html documentation is generated from the .xml files, so the changes there were non included).

Thanks!

PR #154, opened 2020-12-09
Merge fix for 100-continue deadlock for spooled request bodies,

ylavic:
leading to Request Timeout (408).
PR 63855.


ylavic:
Backported to v2.4.47

PR #160, opened 2020-12-14
fix assignment instead of comparison

ihsinme:
there is a mistake in the code. which is difficult to notice. I suggest the following fix.

### DIFF
--- a/server/util_pcre.c
+++ b/server/util_pcre.c
@@ -425,7 +425,7 @@ AP_DECLARE(int) ap_regname(const ap_regex_t *preg,
             apr_array_push(names);
         }
         if (upper || prefix) {
-            char *name = ((char **) names->elts)[capture] =
+            char *name = ((char **) names->elts)[capture] ==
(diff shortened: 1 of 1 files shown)


covener:
I think it is intentionally an assignment, otherwise  the passed in *names is not filled in

ihsinme:
> I think it is intentionally an assignment, otherwise the passed in *names is not filled in

in my opinion your code may not be correct. I beg your pardon, but where there should be a comparison there cannot be an assignment.

ylavic:
The code is admittedly not the clearest one, but its point is to assign the result of `prefix ? ... : ...` to both `name` **and** `names->elts[capture]` (cast omitted here).
The local variable/pointer `name` is to cache the result and avoid using an indirect addressing (`names->elts[capture]`) twice in the same block (uppercasing below).
That's valid `C` at least...

notroj:
You can also see that this change causes the test cases to fail.

ihsinme:
I'm sorry I wasted your time.

notroj:
No time wasted, thank you for trying to improve the code!

PR #161, opened 2020-12-14
fixing an incorrect buffer size

ihsinme:
there is probably an error in the source code, since memory is allocated with a size equal to the pointer to conn_config.
I suggest fixing it.


ylavic:
Fixed in http://svn.apache.org/r1884431 along with another change.
The `struct conn_config_t` contains only a single pointer field for now, so `sizeof(conn_config) == sizeof(*conn_config)`, but the change is needed should the structure be later extended.

Thanks!

PR #163, opened 2020-12-16
2.4.x: Fix races in pidfile handling

notroj:


minfrin:
Backported to v2.4.47:

http://svn.apache.org/viewvc?rev=1885562&view=rev

PR #164, opened 2020-12-24
mod_session_crypto clenups

Maxime2:
add logging for crypto driver error when apr_crypto_block_dcrypt_finish() fails

add cleanup for crypto block

add cleanup for crypto context


Maxime2:
Why this PR is not merged in yet, - is anything wrong with it?

PR #171, opened 2021-02-19
Update apreq.h

adabadadudu:
multiply the CPU usage

### DIFF
--- a/include/apreq.h
+++ b/include/apreq.h
@@ -42,7 +42,7 @@
  * Read chucks of data in 64k blocks from the request 
  */
 
-#define APREQ_DEFAULT_READ_BLOCK_SIZE   (64  * 1024)
+#define APREQ_DEFAULT_READ_BLOCK_SIZE        65536   
 
 /**
  * Maximum number of bytes mod_apreq2 will send off to libapreq2 for parsing. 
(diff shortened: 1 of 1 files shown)


covener:
Why? It seems more readable beforehand.

adabadadudu:
HI Eric .
This is for the CPU .
multiply and plus.
int x = 5+5; // C
if than faster of
int x = 2*5; // C
and this is for program usage and not for readability .
Thanks .

covener:
Is there some environment where this affects what happens at runtime?  https://en.wikipedia.org/wiki/Constant_folding

adabadadudu:
no
code optimization .
read :
https://stackoverflow.com/questions/6879114/multiplication-optimization

rpluem:
I don't see a benefit here as modern compilers precalculate "static" calculations like this during compilation and in the end the values you calculated manually will be used anyway during runtime. Hence we can use the more readable ones.

adabadadudu:
yes .
I don't know .
My compiler is gcc .
What is Apache httpd ( Apache http server ) compiler ?
gcc support this technology ( I don't know )
Okay .
Thanks

rpluem:
gcc does these precalculations.

PR #180, opened 2021-04-18
::TODO:: Update LICENSE

I-Cat:
- [x]  fix insecure url

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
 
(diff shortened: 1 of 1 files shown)


notroj:
Thanks for the PR, but I would not have the confidence to adjust the Apache License text in even a trivial fashion like this, this should go through an official process of updating https://www.apache.org/licenses/LICENSE-2.0 and we shouldn't touch this per project.

wrowe:
Also, there may be other opportunities to update links in docs and example config to https from http that haven't been previously available, and the project would appreciate any PR to refresh these.

I-Cat:
>>> ok brb
>>>
>>

PR #181, opened 2021-04-23
change 2493 urls [Wait for Travis] 🚧

I-Cat:
Commit:
change 2493 urls from http:// to https://

Signed off by  -- @I-Cat

[Wait for Travis] 🚧


covener:
This may have damaged the korean translations (diff in github looks suspicious) updates all the license/copyright headers for nearly every file,  and hit at least 1 URL that had an explicit port 80 in it. It also means things like /server-status and /server-info examples don't work w/o TLS.  I don't really see the added value to offset this.

I-Cat:
> This may have damaged the korean translations (diff in github looks suspicious) updates all the license/copyright headers for nearly every file, and hit at least 1 URL that had an explicit port 80 in it. It also means things like /server-status and /server-info examples don't work w/o TLS. I don't really see the added value to offset this.
Reply to : @covener covener

> Message:
Yes Thank you for your input.  Maybe changing all the URLs to https:// all at once was a little much. That's odd, I will check the diff and see what happened to the port. Also will close and revert. The only way I could test this was by a pull request. Is there a nother way I can test this before I make a pull request?  Then I can verify a passing build. All advice is appreciated.

Signed-off-by: I-Cat

notroj:
This is not mergeable as-is as @covener said, e.g. touching LICENSE and the license headers is definitely not allowed. If you split this up into small, uncontroversial, mergeable changes (e.g. a PR fixing spelling mistakes in CHANGES) those would be possible to review as separate PRs. I'm closing this since it's not going to be handled, sorry.

PR #182, opened 2021-05-01
mod_cache_disk: handle the failure of storing results to temp file, add error message

byshen:
* modules/cache/mod_cache_disk.c: handle the failure of storing results to temp file, add error message


byshen:
APLOGNO is currently blank for demonstration, let me know if I need to fix it. Thanks!

PR #183, opened 2021-05-01
Add reasons before returning declined in mod_dialup 

byshen:
* modules/test/mod_dialup.c : apr_file_open() may fail in the dialup handler, so leave a log message before simply return declined.

### DIFF
--- a/modules/test/mod_dialup.c
+++ b/modules/test/mod_dialup.c
@@ -181,6 +181,9 @@ dialup_handler(request_rec *r)
                        , 0, r->pool);
 
     if (rv) {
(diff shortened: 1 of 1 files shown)


modules/test/mod_dialup.c
```diff
@@ -181,6 +181,9 @@ dialup_handler(request_rec *r)
                        , 0, r->pool);
 
     if (rv) {
+        ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, r, APLOGNO()
```
notroj: This should log the rv rather than passing 0?

byshen: Updated it. Thanks!

PR #185, opened 2021-05-04
Rewrite of mod_unique_id:

atlesn:
The old implementation has multiple issues also when the latest counter-modification is taken into account.

The old module was a bit difficult to understand with a lot of extra logic doing byte per byte operations "manually". I'm proposing a new structure to the whole module which is easier to modify.

The actual problem with uniqueness cannot be solved as intended by the original authors 20+ years ago. This implementation used IP-addresses inside the unique ID, this has since been removed.

Maybe a configuration parameter could provide a server ID (set manually) which can be used inside the string.

I'm also unsure about this "Thread ID" parameter, I haven't found any reliable source for this parameter. Currently a truncated pointer is used.

- Use Process ID, Thread ID, Timestamp, 16-bit per thread counter
  and 16 random bits
- Use provided library function for encoding base64
- Use packed struct instead of addressing each member manually in
  and index
- Use global counter /only/ if threads are not being used
- General re-structuring of functions
- Initialize counters on demand, remove two init functions
- Straight forward to change the build-up of the unique ID parameters
  after discussion.
- TODO : Add endian conversion, find better thread ID


atlesn:
Which byte swap functions should be used for the fixed width types that work on all platforms and that preferably are used elsewhere in the code?

atlesn:
I've added the configuration parameter **UniqueIdServerId** which is used to ensure uniqueness across different servers. I've also implemented byte swapping and changed the method of retrieving a thread ID.

wjcarpenter:
@atlesn You might be interested in some of the conversation taking place on this bugid: https://bz.apache.org/bugzilla/show_bug.cgi?id=65159

modules/metadata/mod_unique_id.c
```diff
@@ -19,238 +19,285 @@
...
-    unsigned short counter;
-    unsigned int thread_index;
+/* Unique ID structure members must be aligned on 1-byte boundaries */
+
+#pragma pack(push)
+#pragma pack(1)
+
+typedef struct unique_id_rec {
+    apr_uint32_t process_id;
+    apr_uint64_t thread_id;
```
ylavic, line 69: Since the struct is packed the field `thread_id` (uint64_t) is now unaligned, can we access it without faulting (SIGBUS) on platforms like sparc?
Do we really need packing here, or e.g. `process_id` could be moved below `timestamp` to let the struct be naturally aligned (with no holes)?

atlesn:
@wjcarpenter Thanks, I can see that somebody linked in this PR there as well. I've also posted this diff on bugzilla https://bz.apache.org/bugzilla/show_bug.cgi?id=65307

modules/metadata/mod_unique_id.c
atlesn: Yes, so basically order them by size? 64, 64, 32, 16, 16

PR #187, opened 2021-05-12
Common HTTP/1 parsing code between core and h2.

ylavic:
Merge [r1734009, r1734231, r1734281, r1838055, r1838079, r1876664, r1876674,
       r1876784, r1879078, r1881620] from trunk:

core: Split ap_create_request() from ap_read_request()

added AP_DECLARE for new ap_create_request

declaring ap_create_request for external linkage

Add StrictHostCheck 

.. to allow ucnonfigured hostnames to be rejected. 

The checks happen during NVH mapping and checks that the
mapped VH itself has the host as a name or alias.

add ids

core, h2: send EOR for early HTTP request failure.

The core output filters depend on EOR being sent at some point for correct
accounting of setaside limits and lifetime.

Rework ap_read_request() early failure (including in post_read_request() hooks)
so that it always sends the EOR after ap_die().

Apply the same scheme in h2_request_create_rec() which is the HTTP/2 to HTTP/1
counterpart.

core, h2: common ap_parse_request_line() and ap_check_request_header() code.

Extract parsing/validation code from read_request_line() and ap_read_request()
into ap_parse_request_line() and ap_check_request_header() helpers such that
mod_http2 can validate its HTTP/1 request with the same/configured policy.

core: follow up to r1876664: allow ErrorDocument to read body when applicable

Unless ap_read_request() failed to read the request line or header, or
Transfer-Encoding is invalid, we can still provide the request body to custom
error handlers (ErrorDocument) that ask it (e.g. internal redirects to CGI).

So this commit splits early failure path (previously die_early label) in two,
die_unusable_input and die_before_hooks, where the latter preserves input
filters (including HTTP_IN).

Also, the code to apply the connection timeout and r->per_dir_config from the
server is now in a new apply_server_config() helper since it's used multiple
times. Note that apr_socket_timeout_set() is a noop if the new timeout is the
same as the one already in place, so there is no need to cache the old timeout
nor use apr_socket_timeout_get(). Likewise, r->server is initially set to
c->base_server so apply_server_config() is overall a noop when no change is
needed.

Validate request-target per RFC 7230 section 5.3.

RFC 7230 requires that the request-line URI be absolute, besides
"CONNECT authority-form" and "OPTIONS asterisk-form".

Enforce it in ap_parse_request_line().

Process early errors via a dummy HTTP/1.1 request as well

Process early errors via a dummy HTTP/1.1 request as well to ensure
that the request gets logged correctly and possible custom error
pages are considered. The previous way of directly sending a HTTP/2
answer with the HTTP status code appropriate for the error is more
efficient, but does not log the request nor sents a possible custom
error page.

* modules/http2/h2.h: Add http_status to h2_request struct and define
  H2_HTTP_STATUS_UNSET.

* modules/http2/h2_request.c(h2_request_create_rec): Check if
  http_status is set for the request and die with the
  status code it contains if set.

* modules/http2/h2_session.c(on_header_cb): Adjust the error condition
  now that we mark early errors via http_status: Only return an error
  if the status is not success and http_status is not H2_HTTP_STATUS_UNSET.

* modules/http2/h2_stream.c(set_error_response): Set http_status
  on the request instead of creating headers for a response and a
  respective brigade.

Github: closes #137

Submitted by: minfrin, icing, icing, covener, covener, ylavic, ylavic,
              ylavic, ylavic, rpluem

### DIFF
--- a/docs/manual/mod/core.xml
+++ b/docs/manual/mod/core.xml
@@ -5207,6 +5207,42 @@ recognized methods to modules.</p>
 </directivesynopsis>
 
 <directivesynopsis>
+<name>StrictHostCheck</name>
+<description>Controls whether the server requires the requested hostname be
+             listed enumerated in the virtual host handling the request
+             </description>
+<syntax>StrictHostCheck ON|OFF</syntax>
+<default>StrictHostCheck OFF</default>
+<contextlist><context>server config</context><context>virtual host</context>
+</contextlist>
+<compatibility>Added in 2.5.1</compatibility>
+
+<usage>
+    <p>By default, the server will respond to requests for any hostname,
+    including requests addressed to unexpected or unconfigured hostnames. 
+    While this is convenient, it is sometimes desirable to limit what hostnames
+    a backend application handles since it will often generate self-referential
+    responses.</p>
+
+    <p>By setting <directive>StrictHostCheck</directive> to <em>ON</em>,
+    the server will return an HTTP 400 error if the requested hostname
+    hasn't been explicitly listed by either <directive module="core"
+    >ServerName</directive> or <directive module="core"
+    >ServerAlias</directive> in the virtual host that best matches the
+    details of the incoming connection.</p>
+
+   <p>This directive also allows matching of the requested hostname to hostnames
+   specified within the opening <directive module="core">VirtualHost</directive>
+   tag, which is a relatively obscure configuration mechanism that acts like
+   additional <directive module="core">ServerAlias</directive> entries.</p>
+
+   <p>This directive has no affect in non-default virtual hosts. The value
+   inherited from the global server configuration, or the default virtualhost 
+   for the ip:port the underlying connection, determine the effective value.</p>
+</usage>
+</directivesynopsis>
+
+<directivesynopsis>
 <name>MergeSlashes</name>
 <description>Controls whether the server merges consecutive slashes in URLs.
 </description>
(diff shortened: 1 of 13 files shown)


docs/manual/mod/core.xml
```diff
@@ -5206,6 +5206,42 @@ recognized methods to modules.</p>
...
+<directivesynopsis>
+<name>StrictHostCheck</name>
+<description>Controls whether the server requires the requested hostname be
+             listed enumerated in the virtual host handling the request
+             </description>
+<syntax>StrictHostCheck ON|OFF</syntax>
+<default>StrictHostCheck OFF</default>
+<contextlist><context>server config</context><context>virtual host</context>
+</contextlist>
+<compatibility>Added in 2.5.1</compatibility>
```
rpluem, line 5218: Added in 2.4.49?
```suggestion
<compatibility>Added in 2.4.49</compatibility>
```

rpluem (changes requested):

docs/manual/mod/core.xml
ylavic: Thanks, Stefan has taken care of it while backporting.

ylavic:
Backported in https://svn.apache.org/repos/asf/httpd/httpd/branches/2.4.x@1890245 (ecebcc035ccd8d0e2984fe41420d9e944f456b3c)

PR #188, opened 2021-05-18
Try using SSL_CTX_set_dh_auto() which is present in 1.1.0 and later.

notroj:
It looks like the OpenSSL-internal implementation doesn't use a cache
of generated DH * objects, instead calling BN_bin2bn() each invocation.
Not sure whether this performance impact matters.  The
SSL_CTX_set_tmp_dh_callback() API is deprecated in OpenSSL 3.0.

### DIFF
--- a/modules/ssl/ssl_private.h
+++ b/modules/ssl/ssl_private.h
@@ -1150,7 +1150,7 @@ void ssl_init_ocsp_certificates(server_rec *s, modssl_ctx_t *mctx);
 
 #endif
 
-#ifndef MODSSL_USE_OPENSSL_PRE_1_1_API
+#if MODSSL_USE_OPENSSL_PRE_1_1_API
 /* Retrieve DH parameters for given key length.  Return value should
  * be treated as unmutable, since it is stored in process-global
  * memory. */
--- a/modules/ssl/ssl_private.h
+++ b/modules/ssl/ssl_private.h
@@ -1150,10 +1150,12 @@ void ssl_init_ocsp_certificates(server_rec *s, modssl_ctx_t *mctx);
 
 #endif
 
+#ifndef MODSSL_USE_OPENSSL_PRE_1_1_API
 /* Retrieve DH parameters for given key length.  Return value should
  * be treated as unmutable, since it is stored in process-global
  * memory. */
 DH *modssl_get_dh_params(unsigned keylen);
+#endif
 
 /* Returns non-zero if the request was made over SSL/TLS.  If sslconn
  * is non-NULL and the request is using SSL/TLS, sets *sslconn to the
--- a/modules/ssl/ssl_engine_init.c
+++ b/modules/ssl/ssl_engine_init.c
@@ -92,7 +92,6 @@ static int DH_set0_pqg(DH *dh, BIGNUM *p, BIGNUM *q, BIGNUM *g)
 
     return 1;
 }
-#endif
 
 /*
  * Grab well-defined DH parameters from OpenSSL, see the BN_get_rfc*
@@ -172,6 +171,7 @@ DH *modssl_get_dh_params(unsigned keylen)
         
     return NULL; /* impossible to reach. */
 }
+#endif
 
 static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
                                        server_rec *s)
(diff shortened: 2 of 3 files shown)


modules/ssl/ssl_private.h
```diff
@@ -1150,10 +1150,12 @@ void ssl_init_ocsp_certificates(server_rec *s, modssl_ctx_t *mctx);
 
 #endif
 
+#ifndef MODSSL_USE_OPENSSL_PRE_1_1_API
```
rpluem: Shouldn't this be `#if MODSSL_USE_OPENSSL_PRE_1_1_API`?
BTW: I think there are cases where `MODSSL_USE_OPENSSL_PRE_1_1_API` is not defined at all, in case of `LIBRESSL_VERSION_NUMBER < 0x2060000f` https://github.com/apache/httpd/blob/0dbc5ca0769aa35f0b956294b2c85726d3ea6466/modules/ssl/ssl_private.h#L131-L149
Wouldn't that be an issue for all the `#if MODSSL_USE_OPENSSL_PRE_1_1_API` code?
```suggestion
#if MODSSL_USE_OPENSSL_PRE_1_1_API
```

rpluem (changes requested):

modules/ssl/ssl_private.h
notroj: Thanks @rpluem, I pushed a commit to fix that first point.  w.r.t. second point, I have no idea, if it is broken somebody with LibreSSL will have to show us how to fix it.

bigio: Adding MODSSL_USE_OPENSSL_PRE_1_1_API define in the LIBRESSL_VERSION < 0x2060000f case makes it build with libressl-2.5.5. The code builds with Libressl 3.3.3 as well.
```
diff --git a/modules/ssl/ssl_private.h b/modules/ssl/ssl_private.h
index 7eb9a364b0..5ba968be41 100644
--- a/modules/ssl/ssl_private.h
+++ b/modules/ssl/ssl_private.h
@@ -137,6 +137,7 @@
         SSL_CTX_ctrl(ctx, SSL_CTRL_SET_MIN_PROTO_VERSION, version, NULL)
 #define SSL_CTX_set_max_proto_version(ctx, version) \
         SSL_CTX_ctrl(ctx, SSL_CTRL_SET_MAX_PROTO_VERSION, version, NULL)
+#define MODSSL_USE_OPENSSL_PRE_1_1_API (1)
 #elif LIBRESSL_VERSION_NUMBER < 0x2070000f
 /* LibreSSL before 2.7 declares OPENSSL_VERSION_NUMBER == 2.0 but does not
  * include most changes from OpenSSL >= 1.1 (new functions, macros,
```

bigio (approved):

rpluem (approved):

modules/ssl/ssl_private.h
ylavic: Possibly something like this then?
```
Index: modules/ssl/ssl_private.h
===================================================================
--- modules/ssl/ssl_private.h	(revision 1889960)
+++ modules/ssl/ssl_private.h	(working copy)
@@ -137,13 +137,12 @@
         SSL_CTX_ctrl(ctx, SSL_CTRL_SET_MIN_PROTO_VERSION, version, NULL)
 #define SSL_CTX_set_max_proto_version(ctx, version) \
         SSL_CTX_ctrl(ctx, SSL_CTRL_SET_MAX_PROTO_VERSION, version, NULL)
-#elif LIBRESSL_VERSION_NUMBER < 0x2070000f
+#endif /* LIBRESSL_VERSION_NUMBER < 0x2060000f */
 /* LibreSSL before 2.7 declares OPENSSL_VERSION_NUMBER == 2.0 but does not
  * include most changes from OpenSSL >= 1.1 (new functions, macros, 
  * deprecations, ...), so we have to work around this...
  */
-#define MODSSL_USE_OPENSSL_PRE_1_1_API (1)
-#endif /* LIBRESSL_VERSION_NUMBER < 0x2060000f */
+#define MODSSL_USE_OPENSSL_PRE_1_1_API (LIBRESSL_VERSION_NUMBER < 0x2070000f)
 #else /* defined(LIBRESSL_VERSION_NUMBER) */
 #define MODSSL_USE_OPENSSL_PRE_1_1_API (OPENSSL_VERSION_NUMBER < 0x10100000L)
 #endif
```

rpluem: Looks sensible

notroj: Merged except I have not tried to incorporate the Libre changes since I can't test them, FYI @bigio

PR #191, opened 2021-06-02
Extended Remote IP functionality: Host, Port, "Protocol" (emulating HTTPS presence), etc. for running behind balancers/frontends

AlexAT:
This is related to #68 and is the reimplementation of it to 2.4.48, removing all the hacky parts.
Please tell if anything else needs to be adjusted. If someone is using my older variant (from #68), please test in production to make sure I did not introduce any new bugs (I did the initial testing myself though).

The feature description is:
mod_remoteip is extended by multiple options handy for frontend proxies moving traffic to Apache backends
- RemoteAllowOnlyInternalProxies - only allows 'internal' trusted proxies to connect to the Apache server, so no untrusted requests can be done (such are refused with Forbidden)
- RemoteHostHeader - accepted only from 'internal' trusted proxies, designates name of HTTP header to use as the supplementary Host header used to re-select virtual host inside request processing - keeping the original supplied Host header (if any) normally visible to i.e. user scripts so they don't know internal virtual host ID is different
- RemotePortHeader - accepted only from 'internal' trusted proxies, designates name of HTTP header to use as the port number originally used to connect which is used as 'default port' in the request and its subrequests
- RemoteProtoHeader - accepted only from 'internal' trusted proxies, designates name of HTTP header to supply for HTTPS 'simulation' (only sets the HTTPS used flag internally in the request and its subrequests, so i.e. rewrite conditions work like HTTPS is enabled when the header is present, even if the access to Apache from the proxy is done via HTTP, PHP shows HTTPS enabled, and so on, request HTTP schema is also adjusted)
- RemoteHTTPSEnableProto - accepted only from 'internal' trusted proxies, supplements RemoteProtoHeader, if set, additionally compares RemoteProtoHeader header contents to the value set and only sets HTTPS emulation if header contents match

---

Changes to 2.4.41-46 version in #68:
- No more obscure hacks of SSL module and internal metadata for HTTPS simulation handling and port alteration, rely fully on existing Apache hooks (including new 2.4.48 ssl_conn_is_ssl() hook)
- Use apr_table_nset() for connection annotation and header changes, we strdup() anyways
- RemoteHTTPSEnableProto option omitted now means setting HTTPS mode simulation on any RemoteProtoHeader header presence, without verifying header contents, for compatibility with the previous default "RemoteHTTPSEnableProto https" needs to be used


AlexAT:
Added 'static' declaration missing from the initial commit, build passes now.

AlexAT:
I think this also needs some justification about how it's usable. This one was made with shared hosting clusters and frontend-to-backend splits in mind.

- RemoteAllowOnlyInternalProxies is rather trivial, it allows only your trusted frontends to connect, denying all direct requests to backend servers, a safety measure.
- RemoteHostHeader and RemotePortHeader apply respective header contents after virtual host selection and allow to define a single virtual host for backend, performing real hostname and port mapping on frontends. All .htaccess and other configuration subrequests like rewrite are done to the original virtual host defined in the backend, while i.e. backend scripts see the original Host/Port that were accessed, thinking they are actually running with the specified hostname/port and generating proper URLs from this data. So you can i.e. call your virtual host 'cmr1234.mydomain.com' and map 'yourcustomerhost.com' to 'cmr1234.mydomain.com' on the proxying frontend, sending Host of 'cmr1234.mydomain.com' in the request to backend, and specific headers for the original host/port accessed.
- RemoteProtoHeader / RemoteHTTPSEnableProto is a widely used setup supplement. It allows to perform all SSL on frontend balancers, while running backend over HTTP only in a trusted network, removing SSL overhead from backends. This allows to create a header to inform backend scripts and other dynamic stuff the original request came via SSL to make it generate HTTPS URLs while the backend will always serve unencrypted. This also affects rewrite conditions so i.e. SSL=on/off dependent rules may be used. This feature removes the need of tricky SSL/non-SSL state handling on backends in such setups, faking SSL state and allowing third party stuff to run unmodified in SSL proxy setups.

tomsommer:
Should the options not be prefixed RemoteIP*? So `RemoteIPHostHeader` etc.

Would love to see this added to master 👍

AlexAT:
> Should the options not be prefixed RemoteIP*? So `RemoteIPHostHeader` etc.

Well, they are technically part of remoteip module, but as they do not exactly affect/specify remote IP but remote Host and Port, remote protocol flag or everything at once (in case of RemoteAllowOnlyInternalProxies), I've named them accordingly.

tomsommer:
I tried this patch. It works, but there is something with `RemoteAllowOnlyInternalProxies On`, it returns 403 for requests that run through .htaccess - i can't really figure out any other pattern of why it happens.

But excellent patch otherwise, thank you :)

AlexAT:
> I tried this patch. It works, but there is something with `RemoteAllowOnlyInternalProxies On`, it returns 403 for requests that run through .htaccess - i can't really figure out any other pattern of why it happens.

If you please can share a bit of configuration that allows to reproduce the issue, I'll definitely look into it as I want this to be as clean as possible.

If it's not exactly publicly sharable but possible to send in private, I can be reached by mail at alex [meow] alex-at.net

tomsommer:
It's really hard to debug, but after some time (many reloads) it's like things leak from one request to another. Suddenly %HTTPS in mod_rewrite doesn't return the correct value, where on the previous request it did. After a httpd restart it works again for a little while.

I'm using 
```
  RemoteIPInternalProxy 127.0.0.1/32
  RemoteIPHeader X-Forwarded-For
  RemoteHostHeader X-Forwarded-Host
  RemotePortHeader X-Forwarded-Port
  RemoteProtoHeader X-Forwarded-Proto
  RemoteHTTPSEnableProto https
  RemoteAllowOnlyInternalProxies Off
```

AlexAT:
Hello, @tomsommer 

I have been using this on my massive shared hostings for years, especially with mod_rewrite for customers, and am not seeing this issue. HTTPS state adjustment relies on the internal Apache hook, so I wonder.

Nevertheless, if there is any possibility of this happening, I want to investigate it as it potentially may shoot us in the leg as well. First things first, please state the MPM and list of loaded modules you are using, also the nature of the request (to static files, to dynamic code). Having full mod_rewrite rules (without sensitive data) you are able to reproduce this on also would be very helpful. That's because I suspect this could be mpm_event or other threaded MPM, then I would have to thoroughly doublecheck the local flags scope in the code and dig in that direction (we are using mpm_event at places for it as well, but not for shared hostings, still had never seen the issue, but yes, anyways I want to thoroughly check that).

AlexAT:
@tomsommer 

Okay, knowing it's mpm_event already points me to the test set creation. Not promising it would be fast to diagnose, but I'll surely post if I find (or do not) something.

tomsommer:
There is definitely some leakage between requests on mpm_event, the `HTTPS` variable is switching between on and off when hitting the same non-https page, even though `HTTP_X_FORWARDED_PROTO=http` and `HTTP_X_FORWARDED_PORT=80` remain.

Don't know if https://github.com/gnif/mod_rpaf/issues/42 is relevant

tomsommer:
I turned off keep-alive on the upstream in nginx and that solved the problem, but of course I would like to use keep-alive :)

AlexAT:
@tomsommer 

Yes, I see where it may happen now. It is always spurious 'https on' state that should not be there, am I correct?
Not related to MPM, but to keepalive connections indeed. This appeared due to RemoteHTTPSEnableProto handling changes.
Thanks!!!

AlexAT:
@tomsommer 

Please test with the latest commit added. Should fix the leakage of HTTPS flag under the conditions stated in the commit description.

modules/metadata/mod_remoteip.c
```diff
@@ -571,14 +597,99 @@ static int remoteip_modify_request(request_rec *r)
...
+                    /* Propagate host header value from the header set by proxy and update virtual host chosen (the primary purpose of it)
+                     */
+                    apr_array_header_t *hdrs = apr_array_make(r->pool, 0, sizeof(char*));
+                    while (*host && (val = ap_get_token(r->pool, &host, 1))) {
+                        *(char **)apr_array_push(hdrs) = apr_pstrdup(r->pool, val);
+                        if (*host != '\0')
+                            ++host;
+                    }
+                    apr_table_setn(r->headers_in, "Host", apr_pstrdup(r->pool, ((char **)hdrs->elts)[((hdrs->nelts)-1)]));
+                    r->hostname = apr_pstrdup(r->pool, ((char **)hdrs->elts)[((hdrs->nelts)-1)]);
```
notroj: Using APR_ARRAY_PUSH() and APR_ARRAY_IDX() might make this code easier to read

tomsommer:
@AlexAT Seems to be fixed now 👍

ylavic:
> Please test with the latest commit added. Should fix the leakage of HTTPS flag under the conditions stated in the commit description.

IIUC the scenario fixed here is httpd with `Keepalive on` (listening on an http or https port), a first request arrives at the frontend on an https port so it sets e.g. `X-Forwarded-Proto: https` which causes `HTTPS=on` in httpd, then a second request arrives at the frontend on an http port so it now sets `X-Forwarded-Proto: http` and forwards the request on the same httpd kept alive connection leading to `HTTPS=off` (or unset) in httpd.

So `RemoteHTTPSEnableProto` is supposed to force the `HTTPS` env var according to what the frontend says, regardless of whether httpd is listening on https or http, right?
I suppose that there are cases where the communications between the frontend and httpd can be considered secure/private even if it's not TLS (or the other way around), and that controlling `HTTPS` at the frontend can be useful for e.g. generating the right URL schemes in scripts/redirects, but that'll probably need a big warning in the docs stating that `HTTPS=on` can be set for non-TLS httpd connections (or the other way around, though maybe less critical).

Btw, it seems that `HTTPS=off` (or unsetting) is missing in remoteip_hook_fixups(), and that remoteip_hook_http_scheme() is registered APR_HOOK_LAST thus runs after mod_ssl's ssl_hook_http_scheme() at APR_HOOK_MIDDLE (so if SSLEngine is on, the scheme will always be "https" regardless of the frontend, which may not be consistent with the above?).

modules/metadata/mod_remoteip.c
```diff
@@ -571,14 +597,99 @@ static int remoteip_modify_request(request_rec *r)
...
+                        apr_table_unset(r->connection->notes, "remoteip_https");
+                    }
+                } else {
+                    /* Subsequent requests inside a single connection need to have the annotation reset if they do not have the header set
+                     */
+                    apr_table_unset(r->connection->notes, "remoteip_https");
+                }
+            }
+        }
+    }
```
ylavic, line 720: It seems that all the c->notes could go in remoteip_conn_config_t members instead, which would be more efficient.

modules/metadata/mod_remoteip.c
```diff
@@ -571,14 +597,99 @@ static int remoteip_modify_request(request_rec *r)
...
+            }
+
+            /* Remote HTTPS protocol option, due to transient nature of subrequests we have to annotate the connection itself for ap_conn_is_ssl() hook
+             */
+            if (config->proto_header_name) {
+                const char *proto;
+                if ( (proto = apr_table_get(r->headers_in, config->proto_header_name)) ) {
+                    /* Exact HTTPS protocol ID option not set means just the header presence is enough, otherwise check for a match
+                     */
+                    if (!config->proto_https_enable || !strcmp(proto, config->proto_https_enable)) {
```
ylavic: So if RemoteHTTPSEnableProto is not set we'll consider any RemoteProtoHeader value is https? Couldn't we check for default "https" instead?
Also using ap_cstr_casecmp() instead of strcmp() would be more flexible wrt "https"/config->proto_https_enable case sensitivity).

modules/metadata/mod_remoteip.c
```diff
@@ -1243,20 +1422,39 @@ static const command_rec remoteip_cmds[] =
...
+                  "Specifies a request header to trust as the original Host header, "
+                  "e.g. X-Forwarded-Host, valid only if direct client is internal proxy"),
+    AP_INIT_TAKE1("RemotePortHeader", port_header_name_set, NULL, RSRC_CONF,
+                  "Specifies a request header to trust as the original Port header, "
+                  "e.g. X-Forwarded-Port, valid only if direct client is internal proxy"),
+    AP_INIT_TAKE1("RemoteProtoHeader", proto_header_name_set, NULL, RSRC_CONF,
+                  "Specifies a request header to trust as the original protocol header, "
+                  "e.g. X-Forwarded-Proto, valid only if direct client is internal proxy"),
+    AP_INIT_TAKE1("RemoteHTTPSEnableProto", proto_https_enable_set, NULL, RSRC_CONF,
+                  "Specifies a value in the header specified by RemoteProtoHeaderrequest "
```
ylavic: typo: RemoteProtoHeaderrequest => RemoteProtoHeader

AlexAT:
Hello, @ylavic 

Thanks for the review, I'll do the changes accordingly today or tomorrow.

modules/metadata/mod_remoteip.c
AlexAT: This one is done so we may make virtual "HTTPS" mode dependent just on specific header being present or absent.

If we do default value of "https", then we need additional flag added like RemoteHTTPSEnableCheckProto (default to on, allowing setting it to off to enable "header presence" mode).

AlexAT: Ack for ap_cstr_casecmp()

modules/metadata/mod_remoteip.c
AlexAT: Thanks!

modules/metadata/mod_remoteip.c
AlexAT: Yes, definitely more efficient indeed, just have to check if it lives through mod_rewrite subrequests mechanics though.

AlexAT:
For the hook order and emulating off mode when it's SSL module handling requests, I assume I will have to use APR_HOOK_REALLY_FIRST given the nature of the function then if and then only if config->proto_header_name is set, return DONE instead of DECLINED so the process is complete, but the result is not OK (as it's compared with OK in ap_ssl_conn_is_ssl() @ ssl.c). 

Will first check what adverse effects it may have. [glancing through the code only shows mod_nw_ssl may be not compatible]
There's a fallback to ssl_is_https() inside ap_ssl_conn_is_ssl() and so it will probably not bode well with mod_nw_ssl.

modules/metadata/mod_remoteip.c
ylavic: > This one is done so we may make virtual "HTTPS" mode dependent just on specific header being present or absent.

The problem IMO is that it's the default, so configuring "RemoteProtoHeader X-Forwarded-Proto" alone does probably not work as expected (it would with default to "https").
So maybe default to "https" and have a special RemoteHTTPSEnableProto value like `"<any>"` or `"*"` that'd enable the "any value means HTTPS" mode?

AlexAT: Indeed, would simplify things for part of typical use cases. Probably can be done like `RemoteHTTPSEnableProto ""` then and checked once in the configuration part so to not call the string length function each request.

Ack.

ylavic: If prefer`"*"` if you ask me :)
It's more explicit as a value and still quite unlikely as a sensible remote proto header valeur..

ylavic:
> For the hook order and emulating off mode...

I was talking about the remoteip_hook_http_scheme() actually, I don't think mod_remote_ip should hook into ssl_conn_is_ssl, that one should stay local IMO (we need a way in httpd to know for sure whether the local connection is TLS or not).

I didn't look closely at all the ap_ssl_conn_is_ssl() usages, but the two local and remote indicators shouldn't be mixed..

ylavic:
This may mean adding a new ssl_conn_remote_is_ssl hook and using ap_ssl_conn_remote_is_ssl() where ap_ssl_conn_is_ssl() is currently misused if/when these changes to mod_remote_ip are in.

It could be an APR_OPTIONAL_FN too, provided by mod_remote_ip and uses by e.g. ap_conn_is_remote_ssl() if it exists.

tomsommer:
Just to add, that hooking into `ap_hook_ssl_conn_is_ssl` is the big feature of this module/patch, that you would otherwise need `mod_rpaf` for. There is no other way to let `mod_rewrite` think that the connection is secure and thus make rewrite rules act correctly as the connection to the visitor is indeed https but HTTPD doesn't know this otherwise.

Also the `HTTPS=off` environment variable would indeed be nice, but I solved that with a `SetEnvIf` for now.

I do agree that perhaps `RemoteHTTPSEnableProto` should default to empty/unset.

AlexAT:
> adding a new ssl_conn_remote_is_ssl hook

This would result in changes to multiple modules (rewrite, lua, etc. and importantly and possibly third party modules) that currently rely on ap_ssl_conn_is_ssl() to initialize variables. Currently I don't see how it's used besides this purpose, so hooking into it seems viable. Maybe ap_ssl_local_conn_is_ssl() is to be introduced later for the purpose of tracking local connection state instead, and then it can actually be just an alias of ap_ssl_conn_is_ssl hook inside all modules except remoteip.

ylavic:
> > adding a new ssl_conn_remote_is_ssl hook
> 
> This would result in changes to multiple modules (rewrite, lua, etc. and importantly and possibly third party modules) that currently rely on ap_ssl_conn_is_ssl() to initialize variables.

They rely on it because so far the HTTPS flag is based on local mod_ssl/mod_tls connections, and that's the only criteria. But with these changes to mod_remoteip it's not the case anymore.

For instance ap_ssl_conn_is_ssl() is used in mod_http2 too, and I don't think it cares about remote SSL for its use cases..

Local and remote SSL are to things that should be clearly distinguished in httpd core/modules, we can't pretend local connection is SSL when it's not (or the other way around) based on a frontend setting.

So I'm afraid merging this will come with changes to some other modules and/or core for the correct usage of ap_ssl_conn_is_ssl() or ap_remote_conn_is_ssl() where appropriate.

AlexAT:
Got it. This also asks for HTTPS_CONN variable though that will track the local state (or maybe named differently, but the point is that HTTPS=on/off is to track remote connection state if it is changed by mod_remoteip, and HTTPS_CONN still tracks the local one).

AlexAT:
I presume the new ap_ssl_remote_conn_is_ssl() should traverse the hooking modules, and if the result is still DECLINED, fall back to ap_ssl_conn_is_ssl() to get the answer. Does this feel ok? That way we won't need hooks in each and every module, just mod_remoteip will do the hook as APR_HOOK_MIDDLE then.

ylavic:
> I presume the new ap_ssl_remote_conn_is_ssl() should traverse the hooking modules, and if the result is still DECLINED, fall back to ap_ssl_conn_is_ssl() to get the answer. Does this feel ok?

Looks sensible. I'd call the function ap_remote_is_ssl() though since *remote* is the major word here, and since it would take a request_rec rather than a conn_rec (it's per request IIUC) the conn_ part looks misleading too?

Also in an (edited) comment above I proposed to use an APR_OPTIONAL_FN instead of a hook, maybe it's simpler to have ap_remote_is_ssl() use that function if it exists (mod_remoteip loaded) and fall back to ap_ssl_conn_is_ssl() otherwise. This also specializes the feature to mod_remoteip (which the hook would not), so the hook is probably better..

AlexAT:
> I'd call the function ap_remote_is_ssl() though

Ack.

modules/metadata/mod_remoteip.c
AlexAT: Yes, "*" looks more clear to me as well on a second thought.

modules/metadata/mod_remoteip.c
AlexAT: Thanks, will change.

AlexAT:
Pushed in the first part of changes after the reviews. 
Quick-tested for parameters leakage by a test set. mod_rewrite tests still pending.

The next step (part 2) would be adding a new hook agreed on for remote SSL state instead of using existing one for local SSL state and modifying modules to use it accordingly.

For anyone using it in production, please avoid applying latest commits yet: there are behavioral changes (https enable content check now defaults to 'https') and not all tests are yet performed on the changes.

AlexAT:
Please also consult on how to push proper numbers in APLOGNO(), I have assigned two next numbers in the sequence and updated next number, but I feel I'm doing it wrong.

AlexAT:
Part 2 of the changes pushed, introducing per-request remote_is_ssl hook and corresponding ap_remote_is_ssl() function, change expression evaluator and mod_rewrite to use it (also introducing HTTPS_LOCAL and HTTPS_REMOTE variables on the way).

This is totally untested yet (may even not work as I have not yet did a verification eye pass over the code and build/test) and I'm yet to scan other modules besides mod_rewrite (probably mod_lua, maybe something else) that need corresponding changes.

AlexAT:
Tested against base testset. Tested with mod_rewrite subrequests.

AlexAT:
Here I've introduced one important change that looks legit to me, but please doublecheck it as it may result in very specific behavioral changes. 

From where this comes: when we have remote IP header set, useragent address gets changed by the primary remoteip_modify_request() invocation if the source is 'internal proxy'. Let us assume the new real remote address set is not one of trusted proxies. Then, if our rewrites result in redirect or subrequest, remoteip_modify_request() will get invoked again for the redirect/subrequest, and if RemoteAllowOnlyInternalProxies is set, the request would be denied as our new useragent address set by the first invocation is not one of the trusted proxies anymore.

If I see it correctly, there is technically no need to do any request modification work on subrequests/redirects as all volatile parameters have already been mangled by the main request and thus are actual.

AlexAT:
Okay, that's that. I'm putting it into production test, will request re-review afterwards.

AlexAT:
Updated mod_lua to include new variables and alias is_https to the new hook.
French doc translation not updated, alas.
Also tested for resolution (removal) of interference with local SSL module state - ok.

tomsommer:
Are we sure "remote" is the correct term to use in the naming? I mean it's something that you are going to have a hard time changing later.

AlexAT:
> Are we sure "remote" is the correct term to use in the naming? I mean it's something that you are going to have a hard time changing later.

Well, yes, I'm concerned about it as well. While it's technically correct (with frontend it's remote frontend to remote client, without one it's local server to client, but as client still can be considered 'remote', doesn't feel too wrong) and the module is 'remoteip', but anyways a good thing to reconfirm. 

'Local' and 'Remote' are the same without frontend and differ with one in the end. Maybe renaming 'remote' to 'client' may indeed be viable, not sure.

AlexAT:
Just to add to it, 'HTTPS' and stuff variants without 'local' or 'remote' still remain, but they're technically just aliases to 'remote', whatever it is.

AlexAT:
Not in scope of the pull request itself but for those who use it, the variant for 2.4.56 (lacks APLOGNO(), lacks some docs changes).

[httpd-2.4-remoteip-2.4.56-20230327.patch](https://github.com/apache/httpd/files/11085917/httpd-2.4-remoteip-2.4.56-20230327.patch)

Take care this can still contain bugs I do not know about. I've put this on a single production cluster and it has no issues atm, but anyways, take care, test before using.
Functional change from the initial 2.4.48 variant: RemoteHTTPSEnableProto now defaults back to 'https', to check just for header presence, use "*" for the option. All other changes are based on the review and are described above & in the commit series.
Technically should now work properly with HTTPS connection from frontend to backend as well.

tomsommer:
Is this working in 2.4.57?

Maybe change the title of the pull-request :)

AlexAT:
Alas, maintaining this onwards for next versions has become harder as it contains more changes after reviewing :)

Here we go:

[httpd-2.4-remoteip-2.4.57-20220531.patch](https://github.com/apache/httpd/files/11614748/httpd-2.4-remoteip-2.4.57-20220531.patch)

Totally the same as 2.4.56 one above, adjusted for changed ap_mmn.h. Builds well, should work all the same as there are no related changes to the core it seems.

EDIT @ 31/05/2023: geez, uploaded wrong file 30/05/2023. Please use the one I placed here 31/05/2023 for builds.

AlexAT:
I think this can already be merged actually - but we need maintainers to somehow check and re-review it beforehand.

AlexAT:
I have by mistake uploaded adjusted older version of the patch in variant for 2.4.57. Terribly sorry for that.
The file above has been replaced and I'm duplicating new one here: [httpd-2.4-remoteip-2.4.57-20220531.patch](https://github.com/apache/httpd/files/11614748/httpd-2.4-remoteip-2.4.57-20220531.patch)

AlexAT:
This post is just to confirm latest patch version is running in my production environment @ version 2.4.57 without any new or noticeable issues for a month already.

shaneshort:
I too would love to see this merged, this would be exceptionally handy for our platform.

covener:
Maybe this has been debunked already, but is the new hook really needed vs existing scheme hook -- and trusting the scheme for things like %{HTTPS}e?

AlexAT:
@covener Yes, the point was remote scheme should not affect local modules behavior and be separate, as it indeed can be HTTP/HTTPS flag over HTTP or HTTPS or whatever.

shaneshort:
@AlexAT I don't suppose you've got some binaries built for this? I'm currently stuck between a rock and a hard place with RPAF not liking keepalive and another bug that goes away with keepalive enabled.

AlexAT:
@shaneshort 
Well, I can potentially provide "alternative module" RPM set for "CentOS 9" based platforms (RH9, OL9, etc.) with latest Apache, but as I won't do regular updates and check internal platform compatibility too deeply, its usage would as usual be on your own risk :)

Upd. It seems OL 9.3 would come with Apache 2.4.57 out of the box, for this I can build the replacement module rather seamlessly I think. I don't know if 'CentOS compatible' would be ok for you or you can wait for it, but once it's out, I'll post links to the 64-bit RPM with mod_remoteip_alexat (with a different name so it can be used along but instead of original module). 2.4.57/58 is what I actually run the modified module with, so should be more or less safe to use.

AlexAT:
@shaneshort Okay, here we go, there is a build for OL 9.3 release where native httpd is altered and rebuilt. Few points beforehand:

- Any use is at your own risk, totally no guarantees whatever
- OL 9.3 only, as it's the first version coming with native Apache 2.4.57, may be usable on CentOS or compatibles, but not tested
- This build replaces native httpd RPMs and stays over them (Epoch is increased) even if OS native RPMs update
- I can't guarantee I'll provide updated builds timely or at all when native source RPM changes
- Compared to the patch in this branch, I skipped increase in Apache internal module version number so we don't cause conflicts with existing modules
- I tested at least mod_perl to work with this build, so existing modules should work, but some may require rebuild (although I can't imagine why) - did not test too extensively

Two additional minor and totally non-intrusive / non-default-behavior-changing patches are included as they are in my 'hosting' patch subset which I based the build on (I'm actually going to use this build myself). Both patches are of course public and can obviously be found in the source RPM.
- One adds useful %a and %A specifiers to RequestHeader directive. %a returns local client IP of the request. %A returns local bound IP of the request. Both are intended to be used with 'early' modifier and are mostly to set connecting frontend IP and serving backend IP into headers. Sample use case: RequestHeader set X-Frontend-IP %a early , and now you can have your connecting frontend IP tracked along with X-Real-IP retrieved and substituted by mod_remoteip as request client IP
- Another adds 'host' option to UseCanonicalName directive. It allows to use contents of incoming Host header as 'canonical server name' for mod_rewrite and other processing. Intended to only be used in very specific scenarios where canonic server name in mod_rewrite and elsewhere needs to be different from the hostname passed by frontend for virtual host selection, is to be used only under Host-filtering frontend as it may have certain security implications when Host header contents are not checked and cannot be trusted. Sample use case: user accesses host.domainA. Frontend passes host.domainA as X-Real-Host for virtualhost selection, but the backend is domainB and so the accessed host is host.domainB. If we want canonical server name to still stay host.domainA for the request (so hosted code, mod_rewrite and whatever is to be made 100% unaware it runs as host.domainB virtualhost), we put UseCanonicalName host to the virtualhost and that's it.

Related repository for OL9.3 x86_64 can be browsed at: https://yum.alex-at.net/ol9/native/

If .repo file is desired, it is at https://yum.alex-at.net/alex-at-ol9-native.repo , but take care I don't gurantee uptime and overall availability of the repository

Source RPM is provided so you can check what is inside and build yourself if you (pretty much obviously) do not trust the binaries built, at https://yum.alex-at.net/ol9/native/httpd-2.4.57-5.0.1.alex.1.el9.x86_64.rpm

AlexAT:
To maintainers: please please check if there are any chances this pull request can be re-reviewed and possibly accepted.
It's really useful for i.e. split HTTPS(2/3) frontend and internal HTTP/HTTPS backend scenarios ultimately common nowadays.

tomsommer:
@AlexAT Thanks for all this. Could you make a patch for 2.4.58?

AlexAT:
Hello @tomsommer , sure, I'll look into that this weekend as I have to build it for myself anyways.

AlexAT:
Hello @tomsommer 
Here we go: [httpd-2.4-remoteip-2.4.58-20231211.patch](https://github.com/apache/httpd/files/13636271/httpd-2.4-remoteip-2.4.58-20231211.patch)
Please take care I have not tested in in production yet, but the changes are minor.

tomsommer:
@AlexAT and 2.4.59? :)

Thanks for all the work on this, I hope this gets merged ASAP. otherwise maybe you could create a new mod_remoteip so it's not just a patch?

AlexAT:
@tomsommer 

Here we go:
[httpd-2.4-remoteip-2.4.59-20240603.patch](https://github.com/user-attachments/files/15533094/httpd-2.4-remoteip-2.4.59-20240603.patch)

(only a minor 'magic number' change since the previous patch, it is running for me for a good while, no issues it seems)

AlexAT:
And here we go for 2.4.62. Again only a minor 'magic number' change.
[httpd-2.4-remoteip-2.4.62-20240801.patch](https://github.com/user-attachments/files/16454597/httpd-2.4-remoteip-2.4.62-20240801.patch)

AlexAT:
For 2.4.63, just needs adjusting MODULE_MAGIC_MINOR to 139.

[httpd-2.4-remoteip-2.4.63-20250413.patch](https://github.com/user-attachments/files/19725895/httpd-2.4-remoteip-2.4.63-20250413.patch)

AlexAT:
For 2.4.64, also just needs adjusting MODULE_MAGIC_MINOR.

[httpd-2.4-remoteip-2.4.64-20250711.patch](https://github.com/user-attachments/files/21181969/httpd-2.4-remoteip-2.4.64-20250711.patch)

---

Also an additional on-top (can be applied after the previous one) patch to test (take care, this one is working but was not extensively tested yet!) that adds possibility to change SERVER_ADDR to what is supplied by proxy/frontend. 

The associated header name parameter is **RemoteLocalIPHeader**. 

SERVER_ADDR is changed for CGI/modules, _mod_rewrite_ (local address can be obtained via LOCAL_ADDR substitution instead of SERVER_ADDR), _mod_setenvif_ (local_addr is also available). All other functions (virtual hosts selection whatever) are untouched, using the proper local server address - not sure, maybe I will implement another flag or two to make these utilize remotely supplied address as well before combining that with the main patch. 

[httpd-2.4-remoteip-localip-2.4.64-20250711.patch](https://github.com/user-attachments/files/21182016/httpd-2.4-remoteip-localip-2.4.64-20250711.patch)

AlexAT:
Just notifying that 2.4.66 compiles cleanly and has no changes affecting patch version 2.4.64.

AlexAT:
A little update for 2.4.67 and 2.4.68, nothing but just another MODULE_MAGIC_MINOR change

[httpd-2.4-hosting-remoteip-2.4.67.patch](https://github.com/user-attachments/files/30297632/httpd-2.4-hosting-remoteip-2.4.67.patch)

PR #192, opened 2021-06-04
For optional_no_ca allow invalid purpose

jdachman:
Currently with this setting
`SSLVerifyClient optional_no_ca`
If the client cert has an invalid purpose, SSL fails.

This is necessary for the case when we always want Httpd to pass the client cert with validation information to the backend--and for the backend to make decisions based on that.


notroj:
Per discussion in https://bz.apache.org/bugzilla/show_bug.cgi?id=70003 I think if we're going to extend this list the acceptable failures should be explicitly configurable.

PR #195, opened 2021-06-10
tighten ALPN for proxy connections

icing:
* mod_ssl: tighten the handling of ALPN for outgoing (proxy)
     connections. If ALPN protocols are provided and sent to the
     remote server, the received protocol selected is inspected
     and checked for a match. Without match, the peer handshake
     fails.
     An exception is the proposal of "http/1.1" where it is
     accepted if the remote server did not answer ALPN with
     a selected protocol. This accomodates for hosts that do
     not observe/support ALPN and speak http/1.x be default.

NOTE: the first 11 commits listed in this PR have already been merged. This stems from the fact that httpd github is a read-only mirror of Apache's subversion repository. In the future, I will start a new local branch for such a change, so it will be easier to follow the history of a PR. Thanks.


modules/ssl/ssl_engine_io.c
```diff
@@ -1265,9 +1267,16 @@ static apr_status_t ssl_io_filter_handshake(ssl_filter_ctx_t *filter_ctx)
 #ifdef HAVE_TLS_ALPN
         alpn_note = apr_table_get(c->notes, "proxy-request-alpn-protos");
         if (alpn_note) {
-            char *protos, *s, *p, *last;
+            char *protos, *s, *p, *last, *proto;
             apr_size_t len;
 
+            /* Transforma the note into a protocol formatted byte array:
```
bigio:
```suggestion
            /* Transform the note into a protocol formatted byte array:
```

bigio (approved):
Minor spell fix

modules/ssl/ssl_engine_io.c
icing: I thought it sounded more Italian!

icing: Fixed, thank you.

icing:
Merged in r1890693. Thanks!

PR #197, opened 2021-06-25
mod_ssl: Fix use of encrypted privkeys with OpenSSL 3.0

notroj:
```
* modules/ssl/ssl_engine_init.c (ssl_init_server_certs): Fix use of
  encrypted private keys with OpenSSL 3.0.

* test/travis_run_linux.sh: For TEST_SSL, test loading encrypted
  private keys.
```


notroj:
Fixed in 1fa837533c3850b19e4544b3eb931f37d64dcf85.  See also https://github.com/openssl/openssl/issues/15946

PR #198, opened 2021-06-28
mpm event/worker/prefork: late(r) children stop on restart.

ylavic:


ylavic:
Extended to all unix MPMs (event, worker and prefork).

ylavic:
Committed to trunk (https://github.com/apache/httpd/commit/243c5fad0a8da9a1008681ac60f5b981de6c18d6)

PR #199, opened 2021-06-29
Add child_stopping hook to core

icing:
* core/mpm: add hook 'child_stopping` that gets called when the MPM is
    stopping a child process. The additional `graceful` parameter allows
    registered hooks to free resources early during a graceful shutdown.


icing:
has been merged into 2.4.x

PR #203, opened 2021-07-06
new ap_ssl_bind_outgoing for multi ssl support in proxy connections

icing:
merge of r1890605 from apache svn:

  *) core/mod_proxy/mod_ssl:
     Adding `outgoing` flag to conn_rec, indicating a connection is
     initiated by the server to somewhere, in contrast to incoming
     connections from clients.
     Adding 'ap_ssl_bind_outgoing()` function that marks a connection
     as outgoing and is used by mod_proxy instead of the previous
     optional function `ssl_engine_set`. This enables other SSL
     module to secure proxy connections.
     The optional functions `ssl_engine_set`, `ssl_engine_disable` and
     `ssl_proxy_enable` are now provided by the core to have backward
     compatibility with non-httpd modules that might use them. mod_ssl
     itself no longer registers these functions, but keeps them in its
     header for backward compatibility.
     The core provided optional function wrap any registered function
     like it was done for `ssl_is_ssl`.

git-svn-id: https://svn.apache.org/repos/asf/httpd/httpd/trunk@1890605 13f79535-47bb-0310-9956-ffa450edef68


include/http_ssl.h
```diff
@@ -49,6 +51,40 @@ AP_DECLARE_HOOK(int,ssl_conn_is_ssl,(conn_rec *c))
...
+ * when it is re-used in different request contexts. It will at least be called after the
+ * connection was created and before the pre-connection hooks is invoked.
+ * All outgoing-connection hooks are run until one returns something other than ok or decline.
+ * if enable_ssl != 0, a hook that sets up SSL for the connection needs to return DONE.
+ *
+ * @param c The connection on which requests/data are to be sent.
+ * @param dir_conf The directory configuration in which this connection is being used.
+ * @param enable_ssl If != 0, the SSL protocol should be enabled for this connection.
+ * @return OK or DECLINED, DONE when ssl was enabled
+ */
```
rpluem, line 67: First of all sorry for missing this on #190.
This is a run first hook. Hence it stops when the hook participant returns anything else but DECLINED.

1. What should be the difference between `DONE` and `OK` then? `OK` says I did not enable SSL, but I wanted to stop the hook?
2. The hook implementation in mod_ssl (`ssl_hook_ssl_bind_outgoing`) returns `OK` and not `DONE` when it enabled SSL. And `ap_ssl_bind_outgoing` checks for `OK` and not `DONE`

I think the fix is to adjust the documentation (`OK` signals enabled, `DECLINED` signals nothing done) here and leave the code as is.

server/ssl.c
```diff
@@ -85,6 +96,77 @@ AP_DECLARE(int) ap_ssl_conn_is_ssl(conn_rec *c)
...
+         * any existing optional funcs need to not enable here */
+        if (module_ssl_engine_set) {
+            module_ssl_engine_set(c, dir_conf, 1, 0);
+        }
+        else if (module_ssl_engine_disable) {
+            module_ssl_engine_disable(c);
+        }
+    }
+    if (enable_ssl && !enabled) {
+        ap_log_cerror(APLOG_MARK, APLOG_ERR, 0,
```
rpluem, line 156: Maybe I miss the point, but this looks like the same condition as on line 136. Can't we do it at the end of the if block starting in line 136?

server/ssl.c
```diff
@@ -85,6 +96,77 @@ AP_DECLARE(int) ap_ssl_conn_is_ssl(conn_rec *c)
...
+                      c, APLOGNO(01961) " failed to enable ssl support "
+                      "[Hint: if using mod_ssl, see SSLProxyEngine]");
+        return DECLINED;
+    }
+    return OK;
+}
+
+AP_DECLARE(int) ap_ssl_has_outgoing_handlers(void)
+{
+    return (_hooks.link_ssl_bind_outgoing && _hooks.link_ssl_bind_outgoing->nelts > 0)
```
rpluem: Shouldn't we use `ap_hook_get_ssl_bind_outgoing()` instead of `_hooks.link_ssl_bind_outgoing`?

rpluem (changes requested):

include/http_ssl.h
icing: You are right. I will update the documentation.

server/ssl.c
icing: This checks if the old optional functions that might be present from another ssl module have enabled ssl or not. So, this could happen at the end of the if from line 136, but I preferred a catch at the end.

server/ssl.c
icing: There are checks in mod_proxy and friends on a NULL connection to see if ssl can be possibly enabled at all or if not ssl module is present. Is there a better way to check if any callbacks have registered at the hook?

include/http_ssl.h
icing: Resolved by update, I think.

rpluem: Thanks. It is resolved for the backport, but the old documentation is still in trunk

server/ssl.c
rpluem: `ap_hook_get_ssl_bind_outgoing()`  returns `_hooks.link_ssl_bind_outgoing`, but it avoids using the internal hook data structures directly. Instead we use a function that is part of the API. I admit  that for `_hooks.link_ssl_bind_outgoing->nelts`  `ap_hook_get_ssl_bind_outgoing()->nelts` would still use internal hook data structures, but we have no API for this.

icing: Thanks for the explanation. I changed the code in the recent update.

server/ssl.c
```diff
@@ -85,6 +96,77 @@ AP_DECLARE(int) ap_ssl_conn_is_ssl(conn_rec *c)
...
+                      "[Hint: if using mod_ssl, see SSLProxyEngine]");
+        return DECLINED;
+    }
+    return OK;
+}
+
+AP_DECLARE(int) ap_ssl_has_outgoing_handlers(void)
+{
+    return (ap_hook_get_ssl_bind_outgoing() && ap_hook_get_ssl_bind_outgoing()->nelts > 0)
+        || module_ssl_engine_set || module_ssl_proxy_enable;
```
ylavic, line 168:
```
    apr_array_header_t *hooks = ap_hook_get_ssl_bind_outgoing();
    return (hooks && hooks->nelts > 0)
        || module_ssl_engine_set || module_ssl_proxy_enable;
```
to avoid 2 function calls that the compiler won't optimize?

icing: Looks good.

icing: adding it.

icing: done.

icing:
merged.

PR #204, opened 2021-07-06
improve proxy alpn check

icing:
r1890696
r1890693

  *) mod_ssl: tighten the handling of ALPN for outgoing (proxy)
     connections. If ALPN protocols are provided and sent to the
     remote server, the received protocol selected is inspected
     and checked for a match. Without match, the peer handshake
     fails.
     An exception is the proposal of "http/1.1" where it is
     accepted if the remote server did not answer ALPN with
     a selected protocol. This accomodates for hosts that do
     not observe/support ALPN and speak http/1.x be default.

### DIFF
--- a/modules/ssl/ssl_engine_io.c
+++ b/modules/ssl/ssl_engine_io.c
@@ -1345,7 +1345,7 @@ static apr_status_t ssl_io_filter_handshake(ssl_filter_ctx_t *filter_ctx)
                  * it does not support ALPN (old server) or that it does not support
                  * any of our proposals (Apache itself up to 2.4.48 at least did that). */
                if (!alpn_empty_ok) {
-                    ap_log_cerror(APLOG_MARK, APLOG_INFO, 0, c, APLOGNO()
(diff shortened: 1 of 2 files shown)


modules/ssl/ssl_engine_io.c
```diff
@@ -1220,6 +1234,8 @@ static apr_status_t ssl_io_filter_handshake(ssl_filter_ctx_t *filter_ctx)
                 ap_log_cerror(APLOG_MARK, APLOG_WARNING, 0, c, APLOGNO(03310)
                               "error setting alpn protos from '%s'", alpn_note);
                 ssl_log_ssl_error(SSLLOG_MARK, APLOG_WARNING, server);
+                /* If ALPN was requested and we cannot do it, we must fail */
+                return MODSSL_ERROR_BAD_GATEWAY;
```
rpluem, line 1238: Don't we need to increase `APLOG_WARNING` to `APLOG_ERROR` in `ap_log_cerror` and `ssl_log_ssl_error` above then?

PR #205, opened 2021-07-08
Try always building with -g in CFLAGS at minimum?

notroj:


rpluem:
Why not doing this on trunk as well and only on 2.4.x?

notroj:
> Why not doing this on trunk as well and only on 2.4.x?

Only because the previous 2.4.x build failed so I got to test whether it still failed here.  (I'm sure we should get -g in CFLAGS automatically from autoconf, not sure why we are not)

rpluem:
> 
> 
> > Why not doing this on trunk as well and only on 2.4.x?
> 
> Only because the previous 2.4.x build failed so I got to test whether it still failed here. (I'm sure we should get -g in CFLAGS automatically from autoconf, not sure why we are not)

Thanks for explaining. Makes sense. Hence you do not really want to merge this to 2.4.x yet, but only test how the tests go with `-g` set, correct?

PR #207, opened 2021-07-11
Add application/vnd.geogebra.slides mime type

zbynek:
The type is registered by IANA: https://www.iana.org/assignments/media-types/application/vnd.geogebra.slides


notroj:
Thanks for the PR.

PR #260, opened 2021-08-20
http2 test suite from mod_h2 github repository

icing:
How to use:

 * run `configure` again after you checked this out
 * the following components need to be installed on your system:
   * python3, pytest
   * curl, nghttp, h2load
 * run the tests:   

```
> make install
> cd test
> pytest
```

This starts the installed httpd on local ports 40001 + 40002, runs the test suite and tears it down again. To run
individual test cases, use

```
> cd test/modules/http2
> pytest -k test_004               # run all tests in test_004_post.py
> pytest -k test_004_07         # run test 07 in test_004_post.py

```


icing:
Merged in r1892476 in svn trunk.

PR #263, opened 2021-08-28
Fix a small error due to the unreleased lock before program exit.

jenny-cheung:
Around the fixed codes, when `pid == 0`, there is the  accept_mutex_off(). However, other places are not. We would like to fix a small error due to the unreleased lock before the program exits in the left place.

It is a small, harmless error since the program exits and all of the program resources will be cleaned up by some common OS. However, other OS systems (e.g., embedded systems) do not automatically free some resources at the exit. It is a good manner for resource management. Adding the unlock statement adds symmetry, so the code looks better. Also, the debugger would not warn this case : )

### DIFF
--- a/test/time-sem.c
+++ b/test/time-sem.c
@@ -548,6 +548,7 @@ void main (int argc, char **argv)
             exit (0);
         } else if (pid == -1) {
             perror ("fork");
+            accept_mutex_off ();
             exit (1);
         }
     }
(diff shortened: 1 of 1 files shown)


jenny-cheung:
@ylavic Can you take a look? Thank you very much!

ylavic:
Thanks, fixed in 9e02e2000cc05754a9c1c9a9aead7ce2cc8ea6ee

jenny-cheung:
@ylavic Thank you very much.

PR #270, opened 2021-09-23
  * mod_http2: synchronizes sources with trunk

icing:
Complete synchronisation of sources with trunk version.

  *) mod_http2: when a server is restarted gracefully, any idle h2 worker
     threads are shut down immediately.
     Also, change OpenSSL API use for deprecations in OpenSLL 3.0.

to verify, run

```
> make install
> pytest
```

### DIFF
--- a/modules/http2/h2_version.h
+++ b/modules/http2/h2_version.h
@@ -27,7 +27,7 @@
  * @macro
  * Version number of the http2 module as c string
  */
-#define MOD_HTTP2_VERSION "1.15.18"
+#define MOD_HTTP2_VERSION "1.15.24-git"
 
 /**
  * @macro
@@ -35,7 +35,7 @@
  * release. This is a 24 bit number with 8 bits for major number, 8 bits
  * for minor and 8 bits for patch. Version 1.2.3 becomes 0x010203.
  */
-#define MOD_HTTP2_VERSION_NUM 0x010f12
+#define MOD_HTTP2_VERSION_NUM 0x010f18
 
 
 #endif /* mod_h2_h2_version_h */
--- a/modules/http2/h2_version.h
+++ b/modules/http2/h2_version.h
@@ -27,7 +27,7 @@
  * @macro
  * Version number of the http2 module as c string
  */
-#define MOD_HTTP2_VERSION "1.15.24-git"
+#define MOD_HTTP2_VERSION "1.15.24"
 
 /**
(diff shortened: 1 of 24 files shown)


modules/http2/h2_version.h
```diff
@@ -27,15 +27,15 @@
  * @macro
  * Version number of the http2 module as c string
  */
-#define MOD_HTTP2_VERSION "1.15.18"
+#define MOD_HTTP2_VERSION "1.15.24-git"
```
rpluem: Is this the correct version number?

modules/http2/h2_version.h
```diff
@@ -27,15 +27,15 @@
...
 
 /**
  * @macro
  * Numerical representation of the version number of the http2 module
  * release. This is a 24 bit number with 8 bits for major number, 8 bits
  * for minor and 8 bits for patch. Version 1.2.3 becomes 0x010203.
  */
-#define MOD_HTTP2_VERSION_NUM 0x010f12
+#define MOD_HTTP2_VERSION_NUM 0x010f18
```
rpluem, line 38: Is this the correct version number?

modules/http2/h2_workers.c
```diff
@@ -271,14 +314,50 @@ static apr_status_t workers_pool_cleanup(void *data)
...
+                         "still seeing %d workers living",
+                         apr_atomic_read32(&workers->worker_count));
+            continue;
+        }
+    }
+    if (i >= n) {
+        ap_log_error(APLOG_MARK, APLOG_WARNING, 0, workers->s,
+                     APLOGNO(10291) "h2_workers: cleanup, %d idle workers "
+                     "did not exit after %d seconds.",
+                     apr_atomic_read32(&workers->worker_count), i);
```
rpluem: Isn't this i*timeout seconds (provided timeout is in seconds)?

modules/http2/h2_version.h
icing: Fixed by removing suffix. Thanks for noticing.

modules/http2/h2_version.h
icing: yes, on github updated by configure.

modules/http2/h2_workers.c
icing: Changed the implementation to have an end time and reporting the waited seconds.

PR #280, opened 2021-12-03
Tweak FLUSH handling in mod_deflate input filter

notroj:
```
* modules/filters/mod_deflate.c (deflate_in_filter): Handle FLUSH in
  the input brigade even if done inflating (ctx->done is true), but
  don't try to flush the inflate stream in that case.  (Caught by
  Coverity)
```

### DIFF
--- a/modules/filters/mod_deflate.c
+++ b/modules/filters/mod_deflate.c
@@ -1316,38 +1316,40 @@ static apr_status_t deflate_in_filter(ap_filter_t *f,
             if (APR_BUCKET_IS_FLUSH(bkt)) {
                 apr_bucket *tmp_b;
 
-                ctx->inflate_total += ctx->stream.avail_out;
-                zRC = inflate(&(ctx->stream), Z_SYNC_FLUSH);
-                ctx->inflate_total -= ctx->stream.avail_out;
-                if (zRC != Z_OK) {
-                    inflateEnd(&ctx->stream);
-                    ap_log_rerror(APLOG_MARK, APLOG_WARNING, 0, r, APLOGNO(01391)
-                                  "Zlib error %d inflating data (%s)", zRC,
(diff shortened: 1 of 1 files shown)


modules/filters/mod_deflate.c
```diff
@@ -1298,6 +1298,7 @@ static apr_status_t deflate_in_filter(ap_filter_t *f,
         {
             const char *data;
             apr_size_t len;
+            apr_bucket *tmp_b;
```
rpluem: Can't we move this to line 1354? The block starting there seems to be the only one that needs it.

modules/filters/mod_deflate.c
```diff
@@ -1313,9 +1314,7 @@ static apr_status_t deflate_in_filter(ap_filter_t *f,
                 break;
             }
 
-            if (APR_BUCKET_IS_FLUSH(bkt)) {
-                apr_bucket *tmp_b;
-
+            if (APR_BUCKET_IS_FLUSH(bkt) && !ctx->done) {
```
rpluem: Wouldn't it be more efficient and saving one `APR_BUCKET_IS_FLUSH(bkt)` if we do the following:

```suggestion
            if (APR_BUCKET_IS_FLUSH(bkt)) {
                if (!ctx->done) {
```
And put line 1319 till 1350 in that block?
Of course this makes my first comment moot and `apr_bucket *tmp_b` could stay where it is in the current code. Hence:

```suggestion
            if (APR_BUCKET_IS_FLUSH(bkt)) {
                apr_bucket *tmp_b;
                
                if (!ctx->done) {
```

notroj: Thanks!  Yes, I was playing around with some other changes and ended up with this... makes much more sense that way.

PR #281, opened 2021-12-09
  *) mod_http2: fixes 2 regressions in server limit handling.

icing:
1. When reaching server limits, such as MaxRequestsPerChild, the
        HTTP/2 connection send a GOAWAY frame much too early on new
        connections, leading to invalid protocol state and a client
        failing the request. See PR65731.
        The module now initializes the HTTP/2 protocol correctly and
        allows the client to submit one request before the shutdown
        via a GOAWAY frame is being announced.
     2. A regression in v1.15.24 was fixed that could lead to httpd
        child processes not being terminated on a graceful reload or
        when reaching MaxConnectionsPerChild. When unprocessed h2
        requests were queued at the time, these could stall.
        See <https://github.com/icing/mod_h2/issues/212>.

### DIFF
--- a/change-entries/h2_pr65731_issue212.txt
+++ b/change-entries/h2_pr65731_issue212.txt
@@ -1,0 +1,14 @@
+  *) mod_http2: fixes 2 regressions in server limit handling.
+     1. When reaching server limits, such as MaxRequestsPerChild, the
+        HTTP/2 connection send a GOAWAY frame much too early on new
+        connections, leading to invalid protocol state and a client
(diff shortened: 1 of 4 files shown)


minfrin:
Backported to v2.4.52:

  *) mod_http2: fixes PR65731 and https://github.com/icing/mod_h2/issues/212
     trunk patch: na, fixed on 2.4.x source base
     backport PR: https://github.com/apache/httpd/pull/281
     +1: icing, minfrin, ylavic

PR #289, opened 2022-01-31
PCRE2 with Thread Local Storage

ylavic:
~~First commit is the current backport proposal.~~ (now in 2.4.x as r1898399 / c602ba14811ede722017c4e59e4e30d9990227b4)

Next is TLS work:
r1897240
r1897241
r1897242
r1897244
r1897248
r1897250
r1897260
r1897261
r1897263
r1897386
r1897459
r1897460
r1897461
r1897462
r1897472
r1897543
r1897651
r1897680
r1897689
r1897691
r1897692

Finally (but not chronologically), PCRE version printing:
r1612934
r1612940
r1613189

### DIFF
--- a/server/main.c
+++ b/server/main.c
@@ -99,13 +99,17 @@ static void show_compile_settings(void)
     printf("Server's Module Magic Number: %u:%u\n",
            MODULE_MAGIC_NUMBER_MAJOR, MODULE_MAGIC_NUMBER_MINOR);
 #if APR_MAJOR_VERSION >= 2
-    printf("Server loaded:  APR %s\n", apr_version_string());
-    printf("Compiled using: APR %s\n", APR_VERSION_STRING);
+    printf("Server loaded:  APR %s, PCRE %s\n",
+           apr_version_string(), ap_pcre_version_string(AP_REG_PCRE_LOADED));
+    printf("Compiled using: APR %s, PCRE %s\n",
+           APR_VERSION_STRING, ap_pcre_version_string(AP_REG_PCRE_COMPILED));
 #else
-    printf("Server loaded:  APR %s, APR-UTIL %s\n",
-           apr_version_string(), apu_version_string());
-    printf("Compiled using: APR %s, APR-UTIL %s\n",
-           APR_VERSION_STRING, APU_VERSION_STRING);
+    printf("Server loaded:  APR %s, APR-UTIL %s, PCRE %s\n",
+           apr_version_string(), apu_version_string(),
+           ap_pcre_version_string(AP_REG_PCRE_LOADED));
+    printf("Compiled using: APR %s, APR-UTIL %s, PCRE %s\n",
+           APR_VERSION_STRING, APU_VERSION_STRING,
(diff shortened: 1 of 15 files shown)


server/main.c
```diff
@@ -348,6 +353,23 @@ static process_rec *init_process(int *argc, const char * const * *argv)
...
+#if AP_HAS_THREAD_LOCAL
+    {
+        apr_status_t rv;
+        apr_thread_t *thd = NULL;
+        if ((rv = ap_thread_main_create(&thd, process->pool))) {
+            char ctimebuff[APR_CTIME_LEN];
+            apr_ctime(ctimebuff, apr_time_now());
+            fprintf(stderr, "[%s] [crit] (%d) %s: %s failed "
+                            "to initialize thread context (%i), exiting\n",
+                            ctimebuff, stat, (*argv)[0], failed, rv);
```
rpluem:
```suggestion
            fprintf(stderr, "[%s] [crit] (%d) %s: Failed "
                            "to initialize thread context, exiting\n",
                            ctimebuff, rv, (*argv)[0]);
```
As otherwise we take over the stuff `apr_app_initialize`which seems wrong.

server/util_pcre.c
```diff
@@ -217,7 +264,134 @@ AP_DECLARE(int) ap_regcomp(ap_regex_t * preg, const char *pattern, int cflags)
...
+
+#ifdef HAVE_PCRE2
+        pcre2_match_data_free(tls->data); /* NULL safe */
+        tls->data = pcre2_match_data_create(tls->size, NULL);
+        if (!tls->data) {
+            tls->size = 0;
+            return NULL;
+        }
+#else
+        tls->data = apr_palloc(tp, tls->size * sizeof(int) * 3);
```
rpluem, line 376: If we are using PCRE and not PCRE2 we accumulate memory here. Shouldn't we create a subpool of `tp`, store it in `tls` as well and clean it before the `apr_palloc`? Or is this not worth the effort as we do not expect frequent reallocations? If it is not worth, we may should document this in a comment.

server/main.c
ylavic: Bad copypasta, will fix thanks.

server/util_pcre.c
ylavic: The growth is x 2 so I don't expect to enter here many times. We'll hardly reach 8K per thread for the PCRE1 ovector since it's already `8K / 3*sizeof(int) ~= 682` captures. Filling the current node of `tp` (or unkileky reserve one more node at some point) like we do now sounds better to me than the reserving 8K from the begining with a subpool.

server/main.c
ylavic: Fixed in r1897680 (a31b463532e83d557a4898cfb6411e144bc39266) and cheery picked here.

server/util_pcre.c
rpluem: Good points.

ylavic:
close/reopen to restart ci

ylavic:
Merged in r1898467 / 285b6285e799d3480150375cb853d40973d3ab4c

PR #291, opened 2022-02-03
Icing/http1 separation

icing:
This is a rather large PR that requires some explanation:

## Motivation

Historically, httpd grew on HTTP/1.0 and added later HTTP/1.1 specifics like `Expect` headers
and `Transfer-Encoding` in its request/response handling. It was assumed that HTTP/1.2, if
it ever came, would add more special cases but basically stay very similar on the wire.

With the standardization of HTTP/2 as a binary format, this held no longer true. The HTTP
IETF specifications made later a switch to differentiate between the *serialization* of the
different protocol versions and the inherent HTTP *semantics*. What a request header
*means*, irregardless of how it arrived in a server. This PR is a start in reflecting this separation inside the Apache httpd as well.

The httpd filter stack already defines the different layers necessary:

 * RESOURCE:  the resource served
 * CONTENT: content-encoding, e.g. gzip
 * PROTOCOL:  generic HTTP
 * TRANSCODE:  HTTP/1.x or HTTP/2
 * CONNECTION: TLS security
 * NETWORK:  sockets

The addition of HTTP/2 to the server made it necessary to replace the existing PROTOCOL/TRANSCODE filters for its requests. As the existing ones *knew* about HTTP/1.1 specifics that do not apply to h2. This led to code duplication of generic HTTP semantics. Improved error handling in one part did not necessarily made it into the other.

## The Change

The changes can be divided in three parts and are easier understood that way.

### Core HTTP Handling

Three new meta bucket type have been added: 

 1. `request`: holds `method`, `uri`, `protocol` and `headers` of a HTTP request.
 2. `response`: holds `status`, `reason`, `headers` and `notes` of a HTTP response.
 3.  `headers`: holds `headers` generated before the end of a HTTP request or response. Used for trailers.

With these new meta buckets, input/output filters on protocol level will see buckets in the following patterns:

```
input: (REQUEST|ERROR) (DATA|FLUSH|ERROR|META|HEADERS)* EOS
output: RESPONSE+ (DATA|FLUSH|ERROR|META|HEADERS)* EOS
```

This means that `HTTP` protocol processing is no longer concerned with parsing incoming bytes or formatting output. All `DATA` buckets on this filter level are part of request/response bodies and have no special protocol interpretation. Interim responses, such as the common 101, are sent as `response` buckets and will not confuse any content filters.

The core request function `ap_read_request()` was changed to expect either a `request` or `error` bucket from the transcode/network input filters. Transcoding input filters use the `error` bucket to trigger a HTTP error response by `ap_read_request()`.

All HTTP/1.1 parsing and formatting has been moved from core/mod_http into the new `mod_http1`.

### HTTP/1.1 Handling

The code to parse and format HTTP/1.1 messages has been moved to the new `mod_http1` (which is a mandatory module, just like mod_http, no loading necessary).

For example, in the `pre_read_request` hook, the module installs the filter that parses HTTP/1.x requests *if* the connection protocol is `http/1.1`. It also has the filters for chunking and response formatting. Moved here from `mod_http` and slightly adapted.

### HTTP/2 Handling

`mod_http2` had major code removals, due to workarounds no longer being required. It now can use the standard `ap_process_connection()` to handle requests. Interim responses need no longer parsing. Chunked encoding needs no longer be emulated for requests without a content-length header.

## Trailers

The core HTTP filters now add incoming `headers` buckets to `r->trailers_in`. They add an outgoing `headers` bucket on EOS when `r->trailers_out` is not empty.

These are always sent on a HTTP/2 connection, but on HTTP/1.1 this is only possible when output is already being chunked. How/if/when this should be triggered is to be determined should this PR be merged. 

## Proxying

`mod_http1` adds functions to write a table of headers in HTTP/1.1 format to the server API. Those are now also used in `mod_proxy` in some places. There is more potential in having common functionality, for example parsing of a HTTP/1.1 response is quite similar to requests. And yet, the current code is duplicate.

Support for HTTP trailers has been added to `mod_proxy_http`. When a `Trailers` header is present in the client request, the backend request will forward it and use `Transfer-Encoding: chunked`, even though a Content-Length might be known.

Trailers from the backend are added to the client response and, given that either chunked HTTP/1.1 or HTTP/2 are in play on the front connection, will be sent.

## Tests

A small range of tests has been added in `test/modules/http1`, mainly to verify the  trailers implementation. The trailers checks in `test/modules/http2` have also been extended.

The Apache Perl Test framework passes on my machine. Let's see what travis says.


jfclere:
Would it be possible to group the commits in less commits? It is huge to review ;-)

icing:
I could make a PR with just the core changes, but that one would not pass our CI. If anyone thinks it'd help, let me know.

server/protocol.c
```diff
@@ -1477,9 +1306,62 @@ request_rec *ap_read_request(conn_rec *conn)
 
     ap_run_pre_read_request(r, conn);
 
-    /* Get the request... */
-    if (!read_request_line(r, tmp_bb) || !ap_parse_request_line(r)) {
+    r->request_time = apr_time_now();
+    rv = ap_get_brigade(r->proto_input_filters, tmp_bb, AP_MODE_READBYTES, APR_BLOCK_READ, 0);
```
ylavic, line 1132: IIUC, before we had proto_input_filters that were called by read_request_line() and ap_get_mime_headers_core() for each line with AP_MODE_GETLINE, while now those filters will be called once with AP_MODE_READBYTES (of zero bytes) and will get a REQUEST meta bucket from downstream/TRANSCODE filters.

I wonder how much this can break existing protocol input filters, those that still want to do something on a per header basis will now have to work/iterate on ((ap_bucket_request *)b->data)->headers_in at AP_FTYPE_PROTOCOL level still or register at AP_FTYPE_TRANSCODE + something.

Can't we leave http1_request_in_filter() as AP_FTYPE_PROTOCOL (like h2_c2_filter_request_in() is, it seems)?

changes-entries/http1_separation.txt
```diff
@@ -0,0 +1,31 @@
...
+     - Response header check and completion done in mod_http
+       filter for all HTTP procotol versions.
+     - ap_send_interim_response() now generates a RESPONSE bucket
+       and sends it through the protocol output filters of a request.
+     - there are still several code locations in the server where
+     - several "Transfer-Encoding" checks have been moved from
+       generic HTTP handling into mod_http1.
+       A new note AP_NOTE_REQUEST_BODY_INDETERMINATE has been
+       introduced that, when present in r->notes, indicates that
+       the length of a request body is not known.
```
ylavic: Maybe a new request_rec `body_indeterminate` field/flag rather than a `notes`?
(notes are not very effectively accessed, so it depends on how often it needs to be known by filters/handlers)

include/http_protocol.h
```diff
@@ -67,6 +67,31 @@ AP_DECLARE(request_rec *) ap_create_request(conn_rec *c);
  */
 request_rec *ap_read_request(conn_rec *c);
 
+/**
+ * Tokenize a HTTP/x.x request line.
+ * @param r the request for which to parse the line
+ * @param line the line to tokenize, will be written to.
```
ylavic: The returned `line` is normalized? garbage?
Maybe a const char *line here could be copied then tokenized?
(for instance it seems that ap_parse_request_line() will overwrite r->the_request)

modules/http/http_filters.c
```diff
@@ -420,224 +94,92 @@ apr_status_t ap_http_filter(ap_filter_t *f, apr_bucket_brigade *b,
...
+
+    rv = ap_get_brigade(f->next, b, mode, block, readbytes);
+    if (APR_SUCCESS == rv) {
+        for (e = APR_BRIGADE_FIRST(b);
+             e != APR_BRIGADE_SENTINEL(b);
+             e = next)
+        {
+            next = APR_BUCKET_NEXT(e);
+            if (!APR_BUCKET_IS_METADATA(e)) {
+                ctx->seen_data = (e->length != 0);
```
ylavic: This may overwrite seen_data, maybe:
```
if (e->length != 0) {
    ctx->seen_data = 1;
}
```
?

modules/http1/chunk_filter.c
```diff
@@ -0,0 +1,762 @@
+/* Licensed to the Apache Software Foundation (ASF) under one or more
```
ylavic, line 1: Could this http1/chunk_filter.c be moved from the old one in http/ (using git mv for now and probably svn move later..)?
It'd help reviewing only the changes here, and would preserve history too.

modules/http1/chunk_filter.c
```diff
@@ -0,0 +1,762 @@
...
+         */
+        if (ctx->state == BODY_NONE && f->r->proxyreq != PROXYREQ_RESPONSE) {
+            ctx->at_eos = 1; /* send EOS below */
+        }
+    }
+
+    /* sanity check in case we're read twice */
+    if (ctx->at_eos) {
+        ap_log_rerror(APLOG_MARK, APLOG_TRACE3, 0, f->r,
+                      "ap_http1_transcode_in_filter: return eos");
```
ylavic: s/ap_http1_transcode_in_filter/http1_body_in_filter/ ?

(there are other references to ap_http1_transcode_in_filter elsewhere which probably need to be changed too)

modules/http1/http1_core.c
```diff
@@ -0,0 +1,142 @@
...
+        }
+        ap_add_output_filter_handle(ap_http1_response_out_filter_handle,
+                                    NULL, r, r->connection);
+    }
+}
+
+static int http1_post_read_request_early(request_rec *r)
+{
+    if (!r->main && !r->prev
+        && !strcmp(AP_PROTOCOL_HTTP1, ap_get_protocol(r->connection))) {
```
ylavic: r->proto_num >= HTTP_VERSION(1,1)?

modules/http1/http1_core.c
```diff
@@ -0,0 +1,142 @@
...
+    return OK;
+}
+
+static int http1_post_read_request(request_rec *r)
+{
+    const char *tenc;
+
+    if (!r->main && !r->prev
+        && !strcmp(AP_PROTOCOL_HTTP1, ap_get_protocol(r->connection))) {
+        if (r->proto_num >= HTTP_VERSION(1,0)) {
```
ylavic, line 73: r->proto_num >= HTTP_VERSION(1,1) (instead of strcmp + proto_num checks)?

modules/http1/http1_filters.c
```diff
@@ -0,0 +1,373 @@
...
+                    /* should not happen. do we generate a 500 here? */
+                }
+                ap_remove_output_filter(f);
+                goto pass;
+            }
+            else if (AP_BUCKET_IS_RESPONSE(e)) {
+                ap_bucket_response *resp = e->data;
+
+                ap_log_rerror(APLOG_MARK, APLOG_TRACE2, 0, r,
+                              "ap_http1_transcode_out_filter seeing response bucket status=%d",
```
ylavic: ap_http1_response_out_filter (found elsewhere in the PR too).

server/protocol.c
```diff
@@ -681,245 +681,102 @@ static int field_name_len(const char *field)
...
 
-static const char m_invalid_str[] = "-";
-
-static int read_request_line(request_rec *r, apr_bucket_brigade *bb)
+typedef enum {
+    ra_none,
+    ra_badprotocol,
+    ra_reject09,
+    ra_badmethod09,
+} ra_error;
```
ylavic: `ra_`?? Maybe rl_error_ (for request_line) or http_error_?

server/protocol.c
```diff
@@ -681,245 +681,102 @@ static int field_name_len(const char *field)
...
-    /* Set r->request_time before any logging, mod_unique_id needs it. */
-    r->request_time = apr_time_now();
-
-    if (APLOGrtrace5(r)) {
-        ap_log_rerror(APLOG_MARK, APLOG_TRACE5, 0, r,
-                      "Request received from client: %s",
-                      ap_escape_logitem(r->pool, r->the_request));
-    }
-
-    return 1;
```
ylavic, line 757: `switch (error)` below? This will better catch potential unhandled cases with modern compilers.

ylavic:
First pass, I will come back later to it (no more free time for now, and a need to digest all the changes a bit ;)

modules/http/http_filters.c
ylavic: or `|=`..

modules/http1/chunk_filter.c
ylavic: Likewise for http1/http_filters.c, it seems that all the code could be moved from http/http_filters.c or the few things remaining there don't belng in http1/ ?

server/protocol.c
icing: We can make it a PROTOCOL filter, yes. Any other filter that wants to mess with the raw HTTP/1.1 bytes would need to install itself before. But maybe we add http1_request_in_filter() as REALLY_LAST in the pre_read hook?

changes-entries/http1_separation.txt
icing: Agreed. I started the note early when I did not want to mess with request_rec.

include/http_protocol.h
icing: This method was, I think, static before and is now exposed. So, it would make sense to give it a cleaner calling convention at the cost of an additional pstrdup().

modules/http1/chunk_filter.c
icing: Any suggestion how to do that now in this PR?

include/http_protocol.h
icing: Actually, it has become purely internal to mod_http1 now. So, I'll remove it from the API and use it internally only.

icing: Updated PR with change.

modules/http/http_filters.c
icing: Updated PR with change.

modules/http1/chunk_filter.c
icing: Updated PR with change.

changes-entries/http1_separation.txt
icing: Added a field to `request_rec` named `body_indeterminate` and removed all occurences of `AP_NOTE_REQUEST_BODY_INDETERMINATE`.

modules/http1/http1_core.c
icing: Implemented.

modules/http1/http1_core.c
icing: Implemented.

icing:
I just made a PR with only the API changes from #291.

https://github.com/icing/httpd/pull/1

Since this will never be merged to apache/httpd, I made
it against my own fork.

I started a review and commented on the new things added.
Should you find this useful, I'll make other partial PRs
for core/http1/http2/proxy changes.

Kind Regards,
Stefan

modules/http1/chunk_filter.c
icing: Giving up after several attempts. Changes to a git mv file show up as delete/new here. :(

server/protocol.c
icing: I moved the new functions to `protocol_new.c` so that they do not show up as changes to removed code here.

server/protocol.c
icing: I made http1_request_in_filter a PROTOCOL one.

modules/http1/http1_filters.c
icing: Changed.

server/protocol.c
icing: Changed to `http_error` prefix.

icing:
Replaced by staged PRs #306, #307 and #311.

PR #293, opened 2022-02-04
Trunk ssl handshake nonblocking

icing:
A PR with the changes by minfrin to make SSL handshakes non-blocking. These have been reverted in trunk  due to CI failures and are added here for easier analysis and discussion.


ylavic:
Now in #294

PR #294, opened 2022-02-04
Event wip

ylavic:
mpm_event changes/improvements for async, debugging and pass pytests (see each commit's message).

First commits are cherry-picked from @minfrin's work (also in https://github.com/apache/httpd/tree/trunk-ssl-handshake-nonblocking)

### DIFF
--- a/changes-entries/mod_ssl_async_handshakes.txt
+++ b/changes-entries/mod_ssl_async_handshakes.txt
@@ -1,0 +1,1 @@
+  *) mod_ssl: Perform non blocking and async TLS handshakes. [Graham Leggett]
--- a/configure.in
+++ b/configure.in
@@ -471,6 +471,8 @@ AC_CHECK_HEADERS( \
 string.h \
 limits.h \
 unistd.h \
+time.h \
+mach/mach_time.h \
 sys/socket.h \
 pwd.h \
 grp.h \
@@ -534,6 +536,9 @@ getpwnam \
 getgrnam \
 initgroups \
 bindprocessor \
+clock_getres \
+clock_gettime \
+clock_gettime_nsec_np \
 prctl \
 procctl \
 pthread_getthreadid_np \
--- a/include/ap_mmn.h
+++ b/include/ap_mmn.h
@@ -735,14 +735,18 @@
  *                         ap_check_output_pending()
  * 20211221.27 (2.5.1-dev) Add min_connection_timeout hook and
  *                         ap_get_connection_timeout()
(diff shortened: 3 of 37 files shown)


covener:
Hi @ylavic I have a question about some code that you removed here :)

I am trying to grok current 2.4.x connections_above_limit(). Specifically, why it cares about both threadsperchild and the number of idlers at the same time.  I am hoping you recall what was misguided about it and that will make the rest click for me.

IIUC, the left side is the # of connections that might spring to life and need a thread. We don't want too many to collect into this state, because once they're established it is a pain to shed them.

Other than the flip-flop issue, why did we need to worry about more than threads_per_child*scale on the right side?
Doesn't it double-count the idlers capacity? Or is it some intentional optimism that despite being > tpc*scale we are kind of second-guessing the scaling factor?

I am thinking you came to the same conclusion, I was quite pleased with myself to spot that I arrived (on the surface) at a similar spot as e.g.
https://github.com/apache/httpd/blob/8f8b5ddd607027985792776c81024ba36b3beb61/server/mpm/event/event.c#L4746

But then I took two steps back: in the same WIP PR, i couldn't make much sense of how it works.

Aren't the two sides of pending_above_limit() check not really comparable?  The left side IIUC is the number of idle workers, or the depth of the queue of work for them.  But the limit we compare against is threads_per_child * async_factor?  It seems like both sides need to be scaled (or not) and both sides need to be overall capacity/state or current capacity/state but that kind of gets you back to connections_above_limit but maybe with a different name.

```
    pending_limit = threads_per_child * async_factor;
...
static APR_INLINE int pending_above_limit(int limit)
{
    return ap_queue_info_count(worker_queue_info) < -limit;                                                 
}
...

static APR_INLINE int should_disable_listensocks(void)
{
    if (listensocks_disabled())
        return 0;

    return pending_above_limit(pending_limit);
}
```



```
static APR_INLINE int connections_above_limit(int *busy)
{
    apr_uint32_t i_count = ap_queue_info_num_idlers(worker_queue_info);
    if (i_count > 0) {
        apr_uint32_t c_count = apr_atomic_read32(&connection_count);
        apr_uint32_t l_count = apr_atomic_read32(&lingering_count);
        if (c_count <= l_count
                /* Off by 'listeners_disabled()' to avoid flip flop */
                || c_count - l_count < (apr_uint32_t)threads_per_child +
                                       (i_count - listeners_disabled()) *
                                       (worker_factor / WORKER_FACTOR_SCALE)) {
            return 0;
        }
    }
    else if (busy) {
        *busy = 1;
    }
    return 1;
}
```

PR #295, opened 2022-02-23
Merge r1898127 from trunk

rpluem:
Merge r1898127 from trunk


rpluem:
Backported as 06daa68add52385e3e6d96431a12617a01a5883b (r1898561)

PR #296, opened 2022-02-28
mod_ext_filter: check exit status of filter processes

dimitry-unified-streaming:
Whenever a filter process returns a non-zero exit status, or is killed by a signal, return a HTTP 500 error, and log the reason.

This was a feature we were missing in mod_ext_filter: there was no way to determine whether the filter process had failed, and if so, to fail the entire HTTP request. We think the most reasonable error to return is 500 Internal Server Error.

While here, also unify the way errors from calling `unified_filter()` are reported. I.e. in `ef_input_filter()`, any failures from calling `unified_filter()` would not be logged with `APLOGNO(01468)`.


dimitry-unified-streaming:
It would be nice if this got merged into 2.4 at some point.

PR #298, opened 2022-03-17
  *) core: make ap_escape_quotes() work correctly on strings

icing:
with more than MAX_INT/2 characters, counting quotes double.
     Credit to <generalbugs@zippenhop.com> for finding this.


icing:
Added in trunk as r1899609.

server/util.c
```diff
@@ -2615,7 +2615,7 @@ AP_DECLARE(void) ap_content_type_tolower(char *str)
  */
 AP_DECLARE(char *) ap_escape_quotes(apr_pool_t *p, const char *instring)
 {
-    int newlen = 0;
+    apr_ssize_t extra = 0;
     const char *inchr = instring;
```
tititiou36, line 2619: Nitpick: Why apr_ssize_t and not apr_size_t?

server/util.c
```diff
@@ -2624,21 +2624,24 @@ AP_DECLARE(char *) ap_escape_quotes(apr_pool_t *p, const char *instring)
...
-            newlen++;
         }
         inchr++;
     }
-    outstring = apr_palloc(p, newlen + 1);
+
+    if (!extra) {
+        return apr_pstrdup(p, instring);
+    }
+
```
tititiou36, line 2643: Maybe apr_pmemdup(), as we can easily compute the length, as done below.

icing: Good suggestion.

server/util.c
icing: `extra` is used in a `ptrdiff_t` expression below, which is signed. `apr_ssize_t` is `ptrdiff_t` on all but the most ancients of platforms. 

But is it correct to use? The question is, does it deal correctly with strings that are longer than ptrdiff_t can hold (lets call this PTRDIFF_MAX)? 

 * an `instring` that is longer than PTRDIFF_MAX: this would mean trouble as `(inchr - instring)`in the palloc expression would overflow. Which is undefined behaviour in C. C defines that you can substract pointers into 
 *the same array* to get a number. So, maybe we can ignore that one.
 * an `instring` that is almost as long as PTRDIFF_MAX with enough quote characters that the resulting length is larger than PTRDIFF_MAX. This is currently not checked.

I think we need a check that the expression `(inchr - instring) + extra + 1` did not overflow, e.g. became < 0, to be totally safe. And for this, we need to calculdate in `apr_ssize_t`, I guess.

What do you think?

PR #299, opened 2022-03-18
  *) mod_http2: use pollset only for main connection and wakeups

icing:
for events on streams. Provide streams in INPUT pipe when
     needed and supported on the platform.

This is based on feedback by @ylavic on PR #297.


icing:
Merged to trunk in r1899102.

PR #301, opened 2022-03-25
[WIP] allow rotatelogs sending data to UDS

cztk:
My goal is to improve lifetime for ssd/nvme based systems with lots of data beeing written to.
Sure there may be some 3rd party software to work around this, however I found this a viable solution for me.

Not tested for windows nor am I working on this. However with this small change one can easily use a socket from filebeat e.g.
to send logs to and have it watch the access.log which is used as fallback.

works for me, thought I'd share as I am too lazy to look up how to feature request. Glad to hear how to improve or better ways
I missed but its written now, keep it open, use or trash it :)


notroj:
Writing logs to a Unix domain socket seems like a useful feature but this seems quite orthogonal to the purpose of rotatelogs, since if you are using a socket then you don't the file rotation management which is the purpose of that tool - or am I missing something?  If not, I'd suggest it would make sense as a RFE to add a self-contained piped logger tool for it.

cztk:
First, I am fully aware services exists which can forward log to network or other services but I don't want to depend on them plus the increased maintenance needed.
Also I don't want to miss out log entries if the UDS fails ( upgrade of filebeat for example ) in this case I want file logging taking place as usual.
As filebeat can read logfiles and save its state aswell, I did not want to reinvent the wheel as rotatelogs does the job perfectly well.
So basically: As long the UDS is working, stuff entries there, if it fails write logs as usual and check if UDS is working again.

I am very happy the way it works, using it in production for the time beeing and it just works the way we need.
Having a piped log to UDS only logger would not allow me to catch data when the socket is down unless I implement some sort of cache which brings up the problem: how much data should it be able to cache and where ( disk/memory ) ... so relying on well known implemented logic like rotatelogs is just perfect.

Or did I miss a point?

cztk:
I think I learned my lesson.
Filebeat stopped accepting data and for some reason the socket wasn't handled as expected leading to filling up apaches scoreboard with reading request requests.
Situation resolved after filebeat accepted data again.

I though I did test that, and indeed it did start to write data to files, though the logic of retrying to use the socket should be handled in a thread or something like that.

Even though I dislike beeing dependent and it's pretty dumb I send data to rsyslog via logger now which itself spams data to filebeat using udp.
System resource usage and loopback traffic increased by a lot though, however it works.

Therefore and I don't have enough resources to investigate this further, going to close this.

PR #307, opened 2022-04-04
Use RESPONSE buckets to send HTTP responses out

icing:
## Abstract

  * core: use the new RESPONSE buckets to send out response     status, reason, headers and notes through the output filter from general HTTP protocol handling.
 *  mod_http: install a new HTTP1_RESPONSE_OUT output filter that transforms RESPONSE buckets to HTTP/1.1 format.

## Changes

In the core server, there are 2 places where response are written: 

1. `ap_http_header_filter()`: this is the standard output filter. As soon as it sees some data coming its way, it uses the status of the current `request_rec` to write the corresponding HTTP/1.1 response bytes. Irregardless of the actual protocol version in play.
1. `ap_send_interim_response()`: when a response status like 101 needs to be send, this function writes the HTTP/1.1 bytes onto the output stream.

No more. Instead, both places now create a RESPONSE meta bucket and write that to the output filters. How that gets converted into the correct bytes on the connection is the concern of protocol version specific filters. And indeed, for HTTP/1.x requests, there is a new output filter added that does exactly that. HTTP/2 will convert these meta buckets to the binary frames defined by the h2 protocol (not included in this PR).

## Motivation

There are many checks done in HTTP core that are not specific to version 1.1. The current design makes it difficult or impossible to reuse these checks in other protocol versions like h2. This leads to code duplication and a server more complicated than it needs to be.

While the HTTP/2 implementation could switch off the 1. case listed above, the intermediate response always arrived in HTTP/1.1 format. This was problematic, as h2 code need to *parse* output bytes to check if they looked like an intermediate response (it did not look at response bodies - if everything else works correctly - but it was a weakness and highly inefficient at best).

## Further Work

I have the HTTP/2 changes locally and tested but did not want them in this PR for an easier review by everyone interested. Also, the HTTP/1.x focussed parts should go into a new `mod_http1?` so our code structure reflects that we differentiate between pure HTTP and bytes on the wire.

Also outstanding is the use of the new REQUEST buckets for input.

### DIFF
--- a/include/mod_core.h
+++ b/include/mod_core.h
@@ -31,6 +31,7 @@
 #include "apr_buckets.h"
 
 #include "httpd.h"
+#include "http_protocol.h"
 #include "util_filter.h"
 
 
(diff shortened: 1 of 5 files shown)


covener (approved):

jfclere (approved):

icing:
Thanks for the reviews! Add to trunk in r1899648.

PR #309, opened 2022-04-07
Fix an undefined behavior of dav_fs_dir_file_name()

zhou1615:
dav_fs_dir_file_name() will not set *fname_p to NULL on failure,
and all callers of dav_fs_dir_file_name() does not check the
return value of dav_fs_dir_file_name(), which could lead to an
undefined behavior against fname_p.

Fix this by adding return value check of dav_fs_dir_file_name()


notroj:
Reviewing some old PRs, sorry. This looks good, thanks for sending it in.

PR #310, opened 2022-04-07
Add err return about set_allow_opts() in set_override()

zhou1615:
In set_override(), the return value of set_allow_opts() is not considered.

Fix this by adding an error check about set_allow_opts().


notroj:
Thanks, good catch.

PR #311, opened 2022-04-08
Using REQUEST buckets in HTTP processing

icing:
## Abstract

 * core: expect the new REQUEST buckets from input filters when reading the next request on a connection. No longer parses itself or is tied to a specific HTTP version format.
 * mod_http: install a new HTTP1_REQUEST_IN input filter that reads the HTTP/1.x lines from the network and transforms it into a REQUEST bucket for handling in the core.

## Changes

The HTTP/1.x parsing code in `server/protocol.c` has moved to `modules/http/http_filter.c` and `modules/http/http_protocol.c`. The checking on HTTP conformance (seen in the old `rrl_error` enum) has been split into the generic `http_error` enum in core and the rest has been moved to mod_http and the parsing code.

This means that `ap_read_request()`, the work horse in connection processing, now is agnostic to exactly *how* a request came into the server. The task of transforming network bytes to HTTP requests is now the task of the input filters. Core only expects either a REQUEST or an ERROR bucket. Everything else is a general error for aborting the connection. ERROR buckets can be used to specify the status code for the HTTP response.

On connections, where `http/1.1` is the protocol, mod_http installs the new input filter HTTP1_REQUEST_IN that basically does the parts formerly in `ap_read_request()`. 

## Motivation

As with the PRs regarding HEADERS and RESPONSE buckets, making the httpd core agnostic to specific protocol versions avoids code duplication and workarounds in non HTTP/1.x handling. This comes at basically no cost for HTTP/1.x processing  as the same code is run, only in a slightly different sequence.

## TBD

Again, the changes to mod_http2 are not part of this PR to make review easier. There are several parts that can be removed after this PR. Also, h2 worker threads will no longer really be h2 specific. Which means we could think about a "secondary connection workers" pool in mpm. And monitoring/scoreboards for that etc. etc.

### DIFF
--- a/include/mod_core.h
+++ b/include/mod_core.h
@@ -41,6 +41,7 @@ extern "C" {
 
 /* Handles for core filters */
 AP_DECLARE_DATA extern ap_filter_rec_t *ap_http_input_filter_handle;
+AP_DECLARE_DATA extern ap_filter_rec_t *ap_h1_request_in_filter_handle;
 AP_DECLARE_DATA extern ap_filter_rec_t *ap_h1_body_in_filter_handle;
 AP_DECLARE_DATA extern ap_filter_rec_t *ap_http_header_filter_handle;
 AP_DECLARE_DATA extern ap_filter_rec_t *ap_chunk_filter_handle;
@@ -53,6 +54,10 @@ AP_DECLARE_DATA extern ap_filter_rec_t *ap_byterange_filter_handle;
 apr_status_t ap_http_filter(ap_filter_t *f, apr_bucket_brigade *b,
                             ap_input_mode_t mode, apr_read_type_e block,
                             apr_off_t readbytes);
+
+apr_status_t ap_h1_request_in_filter(ap_filter_t *f, apr_bucket_brigade *bb,
+                                     ap_input_mode_t mode, apr_read_type_e block,
+                                     apr_off_t readbytes);
 
 apr_status_t ap_h1_body_in_filter(ap_filter_t *f, apr_bucket_brigade *b,
                                      ap_input_mode_t mode, apr_read_type_e block,
--- a/modules/http/http_core.c
+++ b/modules/http/http_core.c
@@ -24,6 +24,7 @@
 #include "http_config.h"
 #include "http_connection.h"
 #include "http_core.h"
+#include "http_log.h"
 #include "http_protocol.h"   /* For index_of_response().  Grump. */
 #include "http_request.h"
 
@@ -36,6 +37,7 @@
 
 /* Handles for core filters */
 AP_DECLARE_DATA ap_filter_rec_t *ap_http_input_filter_handle;
+AP_DECLARE_DATA ap_filter_rec_t *ap_h1_request_in_filter_handle;
 AP_DECLARE_DATA ap_filter_rec_t *ap_h1_body_in_filter_handle;
 AP_DECLARE_DATA ap_filter_rec_t *ap_http_header_filter_handle;
 AP_DECLARE_DATA ap_filter_rec_t *ap_h1_response_out_filter_handle;
@@ -269,13 +271,71 @@ static int http_create_request(request_rec *r)
     return OK;
 }
 
-static void http_pre_read_request(request_rec *r, conn_rec *c)
+static void h1_pre_read_request(request_rec *r, conn_rec *c)
 {
     if (!r->main && !r->prev
         && !strcmp(AP_PROTOCOL_HTTP1, ap_get_protocol(c))) {
+        if (r->proxyreq == PROXYREQ_NONE) {
+            ap_add_input_filter_handle(ap_h1_request_in_filter_handle,
+                                       NULL, r, r->connection);
+        }
         ap_add_output_filter_handle(ap_h1_response_out_filter_handle,
                                     NULL, r, r->connection);
     }
+}
+
+static int h1_post_read_request_early(request_rec *r)
+{
+    if (!r->main && !r->prev && r->proto_num <= HTTP_VERSION(1,1)) {
+        /* HTTP1_BODY_IN takes care of chunked encoding and content-length.
+         */
+        ap_add_input_filter_handle(ap_h1_body_in_filter_handle,
+                                   NULL, r, r->connection);
+    }
+    return OK;
+}
+
+static int h1_post_read_request(request_rec *r)
+{
+    const char *tenc;
+
+    if (!r->main && !r->prev && r->proto_num <= HTTP_VERSION(1,1)) {
+        if (r->proto_num >= HTTP_VERSION(1,0)) {
+            tenc = apr_table_get(r->headers_in, "Transfer-Encoding");
+            if (tenc) {
+                r->body_indeterminate = 1;
+
+                /* https://tools.ietf.org/html/rfc7230
+                 * Section 3.3.3.3: "If a Transfer-Encoding header field is
+                 * present in a request and the chunked transfer coding is not
+                 * the final encoding ...; the server MUST respond with the 400
+                 * (Bad Request) status code and then close the connection".
+                 */
+                if (!ap_is_chunked(r->pool, tenc)) {
+                    ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(02539)
+                                  "client sent unknown Transfer-Encoding "
+                                  "(%s): %s", tenc, r->uri);
+                    return HTTP_BAD_REQUEST;
+                }
+
+                /* https://tools.ietf.org/html/rfc7230
+                 * Section 3.3.3.3: "If a message is received with both a
+                 * Transfer-Encoding and a Content-Length header field, the
+                 * Transfer-Encoding overrides the Content-Length. ... A sender
+                 * MUST remove the received Content-Length field".
+                 */
+                if (apr_table_get(r->headers_in, "Content-Length")) {
+                    apr_table_unset(r->headers_in, "Content-Length");
+
+                    /* Don't reuse this connection anyway to avoid confusion with
+                     * intermediaries and request/reponse spltting.
+                     */
+                    r->connection->keepalive = AP_CONN_CLOSE;
+                }
+            }
+        }
+    }
+    return OK;
 }
 
 static int http_send_options(request_rec *r)
@@ -308,10 +368,17 @@ static void register_hooks(apr_pool_t *p)
     ap_hook_map_to_storage(http_send_options,NULL,NULL,APR_HOOK_MIDDLE);
     ap_hook_http_scheme(http_scheme,NULL,NULL,APR_HOOK_REALLY_LAST);
     ap_hook_default_port(http_port,NULL,NULL,APR_HOOK_REALLY_LAST);
+
     ap_hook_create_request(http_create_request, NULL, NULL, APR_HOOK_REALLY_LAST);
-    ap_hook_pre_read_request(http_pre_read_request, NULL, NULL, APR_HOOK_REALLY_LAST);
+    ap_hook_pre_read_request(h1_pre_read_request, NULL, NULL, APR_HOOK_REALLY_LAST);
+    ap_hook_post_read_request(h1_post_read_request_early, NULL, NULL, APR_HOOK_REALLY_FIRST);
+    ap_hook_post_read_request(h1_post_read_request, NULL, NULL, APR_HOOK_REALLY_LAST);
+
     ap_http_input_filter_handle =
         ap_register_input_filter("HTTP_IN", ap_http_filter,
+                                 NULL, AP_FTYPE_PROTOCOL);
+    ap_h1_request_in_filter_handle =
+        ap_register_input_filter("HTTP1_REQUEST_IN", ap_h1_request_in_filter,
                                  NULL, AP_FTYPE_PROTOCOL);
     ap_h1_body_in_filter_handle =
         ap_register_input_filter("HTTP1_BODY_IN", ap_h1_body_in_filter,
--- a/modules/http/http_core.c
+++ b/modules/http/http_core.c
@@ -284,17 +284,6 @@ static void h1_pre_read_request(request_rec *r, conn_rec *c)
     }
 }
 
-static int h1_post_read_request_early(request_rec *r)
-{
-    if (!r->main && !r->prev && r->proto_num <= HTTP_VERSION(1,1)) {
-        /* HTTP1_BODY_IN takes care of chunked encoding and content-length.
-         */
-        ap_add_input_filter_handle(ap_h1_body_in_filter_handle,
-                                   NULL, r, r->connection);
-    }
-    return OK;
-}
-
 static int h1_post_read_request(request_rec *r)
 {
     const char *tenc;
@@ -334,6 +323,10 @@ static int h1_post_read_request(request_rec *r)
                 }
             }
         }
+        /* HTTP1_BODY_IN takes care of chunked encoding and content-length.
+         */
+        ap_add_input_filter_handle(ap_h1_body_in_filter_handle,
+                                   NULL, r, r->connection);
     }
     return OK;
 }
@@ -371,8 +364,7 @@ static void register_hooks(apr_pool_t *p)
 
     ap_hook_create_request(http_create_request, NULL, NULL, APR_HOOK_REALLY_LAST);
     ap_hook_pre_read_request(h1_pre_read_request, NULL, NULL, APR_HOOK_REALLY_LAST);
-    ap_hook_post_read_request(h1_post_read_request_early, NULL, NULL, APR_HOOK_REALLY_FIRST);
-    ap_hook_post_read_request(h1_post_read_request, NULL, NULL, APR_HOOK_REALLY_LAST);
+    ap_hook_post_read_request(h1_post_read_request, NULL, NULL, APR_HOOK_REALLY_FIRST);
 
     ap_http_input_filter_handle =
         ap_register_input_filter("HTTP_IN", ap_http_filter,
--- a/modules/http/http_filters.c
+++ b/modules/http/http_filters.c
@@ -1856,12 +1856,12 @@ AP_CORE_DECLARE_NONSTD(apr_status_t) ap_h1_response_out_filter(ap_filter_t *f,
                 ap_bucket_response *resp = e->data;
 
                 ap_log_rerror(APLOG_MARK, APLOG_TRACE2, 0, r,
-                              "ap_http1_response_out_filter seeing response bucket status=%d",
+                              "ap_h1_response_out_filter seeing response bucket status=%d",
                               resp->status);
                 if (strict && resp->status < 100) {
                     /* error, not a valid http status */
                     ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, r, APLOGNO(10386)
-                                  "ap_http1_response_out_filter seeing headers "
+                                  "ap_h1_response_out_filter seeing headers "
                                   "status=%d in strict mode",
                                   resp->status);
                     rv = AP_FILTER_ERROR;
@@ -1871,7 +1871,7 @@ AP_CORE_DECLARE_NONSTD(apr_status_t) ap_h1_response_out_filter(ap_filter_t *f,
                     /* already sent the final response for the request.
                      */
                     ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, r, APLOGNO(10387)
-                                  "ap_http1_response_out_filter seeing headers "
+                                  "ap_h1_response_out_filter seeing headers "
                                   "status=%d after final response already sent",
                                   resp->status);
                     rv = AP_FILTER_ERROR;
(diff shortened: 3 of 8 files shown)


include/mod_core.h
```diff
@@ -54,6 +55,10 @@ apr_status_t ap_http_filter(ap_filter_t *f, apr_bucket_brigade *b,
                             ap_input_mode_t mode, apr_read_type_e block,
                             apr_off_t readbytes);
 
+apr_status_t ap_h1_request_in_filter(ap_filter_t *f, apr_bucket_brigade *bb,
+                                     ap_input_mode_t mode, apr_read_type_e block,
+                                     apr_off_t readbytes);
+
```
icing, line 61: The new input filter that does the parsing of network bytes.

modules/http/http_core.c
```diff
@@ -269,15 +271,73 @@ static int http_create_request(request_rec *r)
...
 
-static void http_pre_read_request(request_rec *r, conn_rec *c)
+static void h1_pre_read_request(request_rec *r, conn_rec *c)
 {
     if (!r->main && !r->prev
         && !strcmp(AP_PROTOCOL_HTTP1, ap_get_protocol(c))) {
+        if (r->proxyreq == PROXYREQ_NONE) {
+            ap_add_input_filter_handle(ap_h1_request_in_filter_handle,
+                                       NULL, r, r->connection);
+        }
```
icing, line 281: The filter installs itself on non-proxy http/1.1 connections.

modules/http/http_core.c
```diff
@@ -269,15 +271,73 @@ static int http_create_request(request_rec *r)
...
+static int h1_post_read_request_early(request_rec *r)
+{
+    if (!r->main && !r->prev && r->proto_num <= HTTP_VERSION(1,1)) {
+        /* HTTP1_BODY_IN takes care of chunked encoding and content-length.
+         */
+        ap_add_input_filter_handle(ap_h1_body_in_filter_handle,
+                                   NULL, r, r->connection);
+    }
+    return OK;
+}
```
icing: Install the body filter very early, after a request has been read so that anyone added later can rely on its handling.

modules/http/http_core.c
```diff
@@ -269,15 +271,73 @@ static int http_create_request(request_rec *r)
...
+                     * intermediaries and request/reponse spltting.
+                     */
+                    r->connection->keepalive = AP_CONN_CLOSE;
+                }
+            }
+        }
+    }
+    return OK;
+}
+
```
icing, line 333: Very late, after all post hook have run, we check HTTP/1.x specific request properties, e.g. transfer-encoding and the need to close the connection after this request.

modules/http/http_filters.c
```diff
@@ -1856,12 +1856,12 @@ AP_CORE_DECLARE_NONSTD(apr_status_t) ap_h1_response_out_filter(ap_filter_t *f,
                 ap_bucket_response *resp = e->data;
 
                 ap_log_rerror(APLOG_MARK, APLOG_TRACE2, 0, r,
-                              "ap_http1_response_out_filter seeing response bucket status=%d",
+                              "ap_h1_response_out_filter seeing response bucket status=%d",
```
icing, line 1859: Just a rename forgotten in the previous PR.

modules/http/http_filters.c
```diff
@@ -2080,3 +2080,184 @@ static apr_bucket *create_trailers_bucket(request_rec *r, apr_bucket_alloc_t *bu
...
+        REQ_ERROR, /* failed, terminal */
+    } state;
+
+    request_rec *r;
+    char *request_line;
+    char *method;
+    char *uri;
+    char *protocol;
+} h1_request_ctx;
+
```
icing, line 2099: The input filter context that collects all request information to generate the REQUEST bucket.

Code is basically moved here from `server/protocol.c`.

modules/http/http_protocol.c
```diff
@@ -1636,3 +1636,195 @@ AP_DECLARE(void) ap_h1_add_end_chunk(apr_bucket_brigade *b,
         if (tmp) APR_BRIGADE_CONCAT(b, tmp);
     }
 }
+
+typedef enum {
+    rrl_none, rrl_badprotocol, rrl_badmethod, rrl_badwhitespace, rrl_excesswhitespace,
+    rrl_missinguri, rrl_baduri, rrl_trailingtext,
+} rrl_error;
+
```
icing, line 1644: Error checks formerly in `server/protocol.c` that are HTTP/1.x specific and happen during request parsing.

server/protocol.c
```diff
@@ -682,359 +682,13 @@ static int field_name_len(const char *field)
     return end - field;
 }
 
-static const char m_invalid_str[] = "-";
```
icing, line 685: All this has moved to `modules/http/`.

server/protocol.c
```diff
@@ -682,359 +682,13 @@ static int field_name_len(const char *field)
...
-    if (APLOGrtrace5(r)) {
-        ap_log_rerror(APLOG_MARK, APLOG_TRACE5, 0, r,
-                      "Request received from client: %s",
-                      ap_escape_logitem(r->pool, r->the_request));
-    }
-
-    return 1;
-}
-
 AP_DECLARE(int) ap_parse_request_line(request_rec *r)
```
icing, line 685: This function was in the API before and I left it here, but it is now using code from `modules/http`.

server/protocol.c
```diff
@@ -1466,22 +1120,196 @@ static void apply_server_config(request_rec *r)
     r->per_dir_config = r->server->lookup_defaults;
 }
 
+typedef enum {
+    http_error_none,
+    http_error_badprotocol,
+    http_error_reject09,
+    http_error_badmethod09,
+} http_error;
```
icing, line 1128: The parts from the former `rrl_error` that apply to HTTP in general. Used when assigning a REQUEST bucket to a `request_rec`.

server/protocol.c
```diff
@@ -1466,22 +1120,196 @@ static void apply_server_config(request_rec *r)
...
+    else {
+        r->assbackwards = 1;
+        protocol = "HTTP/0.9";
+        proto_num = HTTP_VERSION(0, 9);
+    }
+    r->protocol = protocol;
+    r->proto_num = proto_num;
+    return error;
+}
+
```
icing, line 1174: Code formerly part of HTTP/1.x parsing that now checks the values delivered in a REQUEST bucket.

server/protocol.c
```diff
@@ -1466,22 +1120,196 @@ static void apply_server_config(request_rec *r)
...
 AP_DECLARE(int) ap_assign_request(request_rec *r,
                                   const char *method, const char *uri,
                                   const char *protocol)
 {
-    /* dummy, for now */
-    (void)r;
-    (void)method;
-    (void)uri;
-    (void)protocol;
+    core_server_config *conf = ap_get_core_module_config(r->server->module_config);
```
icing, line 1179: Assign `method`, `uri` and `protocol` to a `request_rec`. These are coming from a received REQUEST bucket and need to be checked for validity. And if valid, all concerned fields need to be updated.

Again, code moved here from former http/1.x parsing.

server/protocol.c
```diff
@@ -1490,8 +1318,59 @@ request_rec *ap_read_request(conn_rec *conn)
 
     ap_run_pre_read_request(r, conn);
 
-    /* Get the request... */
-    if (!read_request_line(r, tmp_bb) || !ap_parse_request_line(r)) {
+    r->request_time = apr_time_now();
+    rv = ap_get_brigade(r->proto_input_filters, tmp_bb, AP_MODE_READBYTES, APR_BLOCK_READ, 0);
```
icing, line 1322: This is a bit quirky,  to read 0 bytes from the input filter to get only the meta buckets. But, if we read any more, we need to push it back somehow so that the request handlers can read it all during processing.

server/protocol.c
```diff
@@ -1490,8 +1318,59 @@ request_rec *ap_read_request(conn_rec *conn)
...
+         e != APR_BRIGADE_SENTINEL(tmp_bb);
+         e = APR_BUCKET_NEXT(e))
+    {
+        if (AP_BUCKET_IS_REQUEST(e)) {
+            if (!breq) breq = e->data;
+        }
+        else if (AP_BUCKET_IS_ERROR(e)) {
+            access_status = ((ap_bucket_error *)(e->data))->status;
+            goto die_unusable_input;
+        }
```
icing, line 1340: The filter detected an error that should result in a HTTP response with the error status code.

server/protocol.c
```diff
@@ -1490,8 +1318,59 @@ request_rec *ap_read_request(conn_rec *conn)
...
+        else if (!APR_BUCKET_IS_METADATA(e) && e->length != 0) {
+            break;
+        }
```
icing, line 1344: We should not really get those on our 0 length read, but better be safe.

server/protocol.c
```diff
@@ -1490,8 +1318,59 @@ request_rec *ap_read_request(conn_rec *conn)
...
+        uri = breq->uri;
+        protocol = breq->protocol;
+        headers = breq->headers;
+    }
+    else {
+        method = apr_pstrdup(r->pool, breq->method);
+        uri = apr_pstrdup(r->pool, breq->uri);
+        protocol = apr_pstrdup(r->pool, breq->protocol);
+        headers = breq->headers? apr_table_clone(r->pool, breq->headers) : NULL;
+    }
```
icing, line 1377: Some optimisation on assigning the values. If the input filter used `r->pool` for the values in the REQUEST bucket, we can just work with these. Otherwise copy.

modules/http/http_core.c
ylavic: Why this late and not in h1_post_read_request_early() too?
A post_read_request hook or input filter run from there might need to know about TE vs CL no?

modules/http/http_filters.c
```diff
@@ -2080,3 +2080,184 @@ static apr_bucket *create_trailers_bucket(request_rec *r, apr_bucket_alloc_t *bu
...
+    }
+
+    if (!ctx) {
+        f->ctx = ctx = apr_pcalloc(r->pool, sizeof(*ctx));
+        ctx->r = r;
+        ctx->state = REQ_LINE;
+    }
+
+    /* This filter needs to get out of the way of read_request_line() */
+    ap_remove_input_filter(f);
```
ylavic, line 2182: Hm, enough for now I suppose but it does not look nonblocking compatible. We probably should think of nonblocking in the new design..

modules/http/http_filters.c
```diff
@@ -2080,3 +2080,184 @@ static apr_bucket *create_trailers_bucket(request_rec *r, apr_bucket_alloc_t *bu
...
+            }
+
+            if (!ap_h1_tokenize_request_line(r, ctx->request_line,
+                                             &ctx->method, &ctx->uri, &ctx->protocol)) {
+                http_status = HTTP_BAD_REQUEST;
+                goto cleanup;
+            }
+            /* got the request line and it looked to contain what we need */
+            ctx->state = REQ_HEADERS;
+            break;
```
ylavic, line 2214: fall through?

modules/http/http_filters.c
```diff
@@ -2080,3 +2080,184 @@ static apr_bucket *create_trailers_bucket(request_rec *r, apr_bucket_alloc_t *bu
...
+            APR_BRIGADE_INSERT_HEAD(bb, e);
+            ctx->state = REQ_BODY;
+            ap_log_rerror(APLOG_MARK, APLOG_TRACE1, rv, r,
+                          "http1 request and headers parsed: %s %s %s",
+                          ctx->method, ctx->uri, ctx->protocol);
+            goto cleanup;
+
+        case REQ_BODY:
+            /* we should not come here */
+            AP_DEBUG_ASSERT(0);
```
ylavic, line 2242: It seems to me that REQ_BODY and REQ_ERROR are hardly expected here, let's not code them and have a single `default:` case which assert()s + EINVAL?

modules/http/http_filters.c
```diff
@@ -2080,3 +2080,184 @@ static apr_bucket *create_trailers_bucket(request_rec *r, apr_bucket_alloc_t *bu
...
+        default:
+            rv = APR_EINVAL;
+            goto cleanup;
+        }
+    } /* while(APR_SUCCESS == rv) */
+
+cleanup:
+    if (http_status != HTTP_OK) {
+        ap_log_rerror(APLOG_MARK, APLOG_TRACE3, 0, r,
+                      "failed reading request line, returning error bucket %d", http_status);
```
ylavic, line 2256: Returning an error bucket from input filters in not usual (and probably not expected by upstream filters and handlers, why not return rv and let the caller decide with ap_map_http_request_error()?

[edit: I saw it was caught in ap_read_request() actually? so early enough for later handling to never see an error bucket?]

modules/http/http_protocol.c
```diff
@@ -1636,3 +1636,195 @@ AP_DECLARE(void) ap_h1_add_end_chunk(apr_bucket_brigade *b,
...
+/* get the length of a name for logging, but no more than 80 bytes */
+#define LOG_NAME_MAX_LEN 80
+static int log_name_len(const char *name)
+{
+    apr_size_t len = strlen(name);
+    return (len > LOG_NAME_MAX_LEN)? LOG_NAME_MAX_LEN : (int)len;
+}
+
+static void rrl_log_error(request_rec *r, rrl_error error, char *etoken)
```
ylavic: **const** char *etoken

modules/http/http_protocol.c
```diff
@@ -1636,3 +1636,195 @@ AP_DECLARE(void) ap_h1_add_end_chunk(apr_bucket_brigade *b,
...
+{
+    if (error == rrl_badprotocol)
```
ylavic: switch (error)?

modules/http/http_protocol.c
```diff
@@ -1636,3 +1636,195 @@ AP_DECLARE(void) ap_h1_add_end_chunk(apr_bucket_brigade *b,
...
+    *pmethod = method;
+    *puri = uri;
+    *pprotocol = protocol;
+    *perror_token = etoken;
+    return e;
+}
+
+AP_DECLARE(int) ap_h1_tokenize_request_line(
+        request_rec *r, const char *line,
+        char **pmethod, char **puri, char **pprotocol)
```
ylavic: The returned `char**`s seem to be passed to ap_assign_request() and probably don't need to be modified inplace by the caller, so `const char**`? This would apply to tokenize_request_line() then too.

server/protocol.c
ylavic: I'm not sure it needed to, it makes the changes harder to follow and protocol.c seems to be calling ap_h1_ specific functions still.
Maybe move them after the functional changes if needed?

server/protocol.c
ylavic: Yeah all the story of the protocol.c, hard to move things to a module..

server/protocol.c
ylavic: Maybe call it ap_assign_request_line()?

server/protocol.c
ylavic: Ah OK, so the input error bucket is caught here, why not.
So up to ap_read_request() some special input filtering is in place and after that only body filtering right?
Could you please show the output of `dump_filters r->[proto_]input_filters` (from our gdbinit) at different ap_read_request() steps? This'd possibly help have a picture here, or however you'd like to describe it :)

server/protocol.c
ylavic: So shouldn't we error out in this case (i.e. even if we saw REQUEST bucket)?

server/protocol.c
```diff
@@ -1531,18 +1410,7 @@ request_rec *ap_read_request(conn_rec *conn)
...
-        apr_brigade_cleanup(tmp_bb);
-        if (r->status != HTTP_OK) {
-            ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(00567)
-                          "request failed: error reading the headers");
-            access_status = r->status;
-            goto die_unusable_input;
-        }
-
-        clen = apr_table_get(r->headers_in, "Content-Length");
+        const char *clen = apr_table_get(r->headers_in, "Content-Length");
```
ylavic, line 1426: IIUC, we should have accurate r->body_indeterminate at this point, so check !r->body_indeterminate here and fail/internal error if no C-L in this case?

modules/http/http_core.c
icing: Valid point. I was thinking about hooks that might tinker with the request regarding this. But you are right, hooks might want to access that information.

I'll move it.

server/protocol.c
icing: 🚴‍♀️Makes it easier to understand for all the h1 boomers, yes.

modules/http/http_filters.c
icing: Agreed. I wanted to limit the scope of this PR in just moving pieces around. And the current `ap_read_request` is blocking. But we could make it non-blocking with run_connection returning on a APR_EAGAIN.
Is that what you have in mind?

modules/http/http_filters.c
icing: I wanted to modify/improve the existing code as little as possible, so the review can be more clear. Maybe we'll have a look at it in another PR afterwards for code improvements?

modules/http/http_filters.c
icing: Then a filter that knows the HTTP status it wants to trigger would have to reverse engineer ap_map_http_request_error()'s mapping? It seemed rather straightforward to use ERROR buckets for this, but maybe I fail to see the complications.

modules/http/http_protocol.c
icing: Will change.

modules/http/http_protocol.c
icing: Will change.

server/protocol.c
icing: I think we should indeed. It will not work if we get request body buckets before we have read and analysed the request. Will add the check and fail.

modules/http/http_core.c
icing: done.

server/protocol.c
icing: done.

server/protocol.c
icing: done.

modules/http/http_protocol.c
icing: done

modules/http/http_protocol.c
icing: done.

modules/http/http_core.c
icing: done

modules/http/http_filters.c
ylavic: That's what I meant indeed, this filter will have to be reworked to not remove itself unconditionally on the first call then.

modules/http/http_filters.c
ylavic: +1

modules/http/http_filters.c
ylavic: I was wondering if any handler calling ap_get_brigade(), like mod_proxy forwarding or even the default_handler(), could now receive a bucket as error signal (while they usually don't handle input errors like this) thus simply ignoring the error.
Then I saw ap_read_request() which handles the error bucket and [edit]ed my comment, IIUC it shouldn't be an issue because ap_h1_request_in_filter() is supposed to remove itself before ap_read_request() ends.

modules/http/http_filters.c
icing: Yes, I'd like `ap_read_request()` to get the brigade NONBLOCK.

server/protocol.c
icing: I believe this is too early. Currently, we set this in the post_read hook. Also, requests may have neither C-L or body_indeterminate, e.g. if they are known to have no request body.

Maybe converting body_indeterminate into a no_body flag would make things more clear?

server/protocol.c
icing: Added a macro in util_filter.h to dump a filter chain and now see in the logs for a http/1.1 request:

```
filter-dump before pre_read[0]: reqtimeout [58 CONNECTION]
filter-dump before pre_read[1]: log_input_output [59 CONNECTION]
filter-dump before pre_read[2]: core_in [60 NETWORK]
filter-dump after pre_read[0]: http1_request_in [30 PROTOCOL]
filter-dump after pre_read[1]: reqtimeout [58 CONNECTION]
filter-dump after pre_read[2]: log_input_output [59 CONNECTION]
filter-dump after pre_read[3]: core_in [60 NETWORK]
ap_tokenize_request_line: 'GET / HTTP/1.1'
ap_tokenize_request: error=0, method=GET, uri=/, protocol=HTTP/1.1
http1 request and headers parsed: GET / HTTP/1.1
filter-dump after get_brigade[0]: reqtimeout [58 CONNECTION]
filter-dump after get_brigade[1]: log_input_output [59 CONNECTION]
filter-dump after get_brigade[2]: core_in [60 NETWORK]
checking request: GET / HTTP/1.1
assigned protocol: HTTP/1.1, error=0
assigned request_line: GET / HTTP/1.1, error=0
parsed uri: r->status=200, error=0
filter-dump after post_read[0]: http_in [30 PROTOCOL]
filter-dump after post_read[1]: http1_body_in [40 TRANSCODE]
filter-dump after post_read[2]: reqtimeout [58 CONNECTION]
filter-dump after post_read[3]: log_input_output [59 CONNECTION]
filter-dump after post_read[4]: core_in [60 NETWORK]
```

ylavic (approved):
+1

icing:
Thanks for reviewing! Merged as r1899799 into trunk.

PR #313, opened 2022-04-12
[2.4.x] mpm_event: Fix accounting of active/total processes on ungraceful restart

ylavic:
Children processes terminated by ap_{reclaim,relieve}_child_processes() were
were not un-accounted for total_daemons and active_daemons, which was done in
server_main_loop() only. This led to perform_idle_server_maintenance() thinking
it was over the limit of children processes and never create new ones.

Have this accounting right in event_note_child_{started,stopped}() which is
called both at runtime and reload time.

* server/mpm/event/event.c(struct event_retained_data):
  Rename field max_daemons_limit to max_daemon_used to better describe what
  it's about and to align with AP_MPMQ_MAX_DAEMON_USED.

* server/mpm/event/event.c(event_note_child_stopped):
  Renamed from event_note_child_killed() to clarify that it's not only called
  when a child is killed (i.e. on restart) but whenever a child has stopped.

* server/mpm/event/event.c(event_note_child_stopped):
  Move decrementing {active,total}_daemons and marking child's threads as
  SERVER_DEAD from server_main_loop() so that it's done both at runtime and
  reload time. Log the current number/state of daemons at APLOG_DEBUG level
  for each child stopped.

* server/mpm/event/event.c(event_note_child_started):
  Move incrementing {active,total}_daemons from make_child() for symmetry,
  given that make_child() calls event_note_child_started(). Log the current
  number/state of daemons at APLOG_DEBUG level for each child started.

* server/mpm/event/event.c(perform_idle_server_maintenance):
  Fix possible miscounting of retained->max_daemon_used accross the multiple
  calls to perform_idle_server_maintenance() if ListenCoresBucketsRatio > 0.
  Pass an int *max_daemon_used which starts at zero and is bumped consistently
  for all the buckets, while retained->max_daemon_used is updated only after
  all the buckets have been maintained.

* server/mpm/event/event.c(perform_idle_server_maintenance):
  Use event_note_child_stopped() to handle exited children processes.

Fixes: BZ 66004

git-svn-id: https://svn.apache.org/repos/asf/httpd/httpd/trunk@1899777 13f79535-47bb-0310-9956-ffa450edef68

### DIFF
--- a/changes-entries/event_note_child_stopped.txt
+++ b/changes-entries/event_note_child_stopped.txt
@@ -1,0 +1,2 @@
+  *) MPM event: Fix accounting of active/total processes on ungraceful restart,
+     PR 66004 (follow up to PR 65626 from 2.4.52).  [Yann Ylavic]
--- a/server/mpm/event/event.c
+++ b/server/mpm/event/event.c
@@ -379,7 +379,7 @@ typedef struct event_retained_data {
      * We use this value to optimize routines that have to scan the entire
      * scoreboard.
      */
-    int max_daemons_limit;
+    int max_daemon_used;
 
     /*
      * All running workers, active and shutting down, including those that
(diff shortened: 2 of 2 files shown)


ylavic:
Backported to 2.4.x (r1901199 / 138691a51ced29d2717b98784080bc50e9a0610a)

PR #319, opened 2022-05-30
mod_proxy_http: Avoid 417 responses for non forwardable 100-continue.…

ylavic:
… PR 65666.

Stop returning 417 when mod_proxy has to forward an HTTP/1.1 request with both
"Expect: 100-continue" and "force-proxy-request-1.0" set, mod_proxy can instead
handle the 100-continue by itself before forwarding the request, like in the
"Proxy100Continue Off" case.

Note that this does not change the behaviour of httpd receiving an HTTP/1.0
request with an Expect header, ap_check_request_header() will still correctly
return 417 in this case.

git-svn-id: https://svn.apache.org/repos/asf/httpd/httpd/trunk@1901420 13f79535-47bb-0310-9956-ffa450edef68

### DIFF
--- a/changes-entries/proxy_no_417.txt
+++ b/changes-entries/proxy_no_417.txt
@@ -1,0 +1,2 @@
+  *) mod_proxy_http: Avoid 417 responses for non forwardable 100-continue.
+     PR 65666.  [Yann Ylavic]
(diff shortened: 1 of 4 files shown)


ylavic:
Merged in [r1901584](https://svn.apache.org/repos/asf/httpd/httpd/branches/2.4.x@1901584)

PR #323, opened 2022-06-17
H2 worker pool

icing:
A new implementation of h2 workers, inspired by apr_thread_pool implementation.

 * O(1) cost at registration of connection processing producers
 * no limit on registered producers
 * `join` of ongoing work on unregister
 * callbacks to unlink dependencies into other h2 code
 * memory cleanup on workers deactivation (on idle timeouts)
 * idle_limit as `apr_time_t` instead of seconds


icing:
Merged as r1902005.

PR #324, opened 2022-07-25
Update README with typical markdown

jmcscript:
I believe this would be a definite improvement on the current version. However, I'm aware you may not want a single file modified. Perhaps you'd want them all migrated to markdown at once. If so, I may be able to help.


rbowen:
This has been done in another PR already.

PR #325, opened 2022-07-26
Make DirectorySlash always output a relative value.

FGasper:
Issue #61355: This causes DirectorySlash to create a relative Location
response header rather than an absolute one. Thus, httpd won’t
inadvertently downgrade TLS-secured connections to non-TLS connections
when it operates behind a (TLS-terminating) proxy.

### DIFF
--- a/modules/mappers/mod_dir.c
+++ b/modules/mappers/mod_dir.c
@@ -292,8 +292,8 @@ static int fixup_dir(request_rec *r)
                                 "/", NULL);
         }
 
-        apr_table_setn(r->headers_out, "Location",
-                       ap_construct_url(r->pool, ifile, r));
+        apr_table_setn(r->headers_out, "Location", ifile);
+
         return HTTP_MOVED_PERMANENTLY;
     }
 


FGasper:
Bugzilla: https://bz.apache.org/bugzilla/show_bug.cgi?id=61355

tlandschoff-scale:
We just ran into this, inadvertantly doing a https => http => https redirect chain. This fix seems quite obvious to me, any reason this can't get merged?

covener:
obsoleted RFC 2616 required an absoluteURI in Location: headers. Even without that, there could be rewrite-like rules out there expecting it to be maintained. So I don't think it's a no-brainer to switch.

 "inadvertent" part aside, if Apache is behind an HTTPS offload it should have a https://-prefixed `ServerName`

mobeigi:
This bug is extremely annoying, its causing extra 301 hops from https -> http -> https on every request that invokes the directory "fixing" (i.e. /subdir -> /subdir/).

The proposed solution with relative paths seems good to me.
Another could be to respect `X-Forwarded-Proto` and set the scheme accordingly.

Setting the `ServerName` to be `https` prefixed in the VirtualHost does fix this but shouldn't be explicitly required in my opinion.

FGasper:
FWIW, designating a URI as a "server name" seems unintuitive.

PR #326, opened 2022-08-02
  *) mod_ssl: when a proxy connection had handled a request using SSL…

icing:
…, an

     error was logged when "SSLProxyEngine" was only configured in the
     location/proxy section and not the overall server. The connection
     continued to work, the error log was in error. Fixed PR66190.


notroj:
Merged in c2e15a4

PR #328, opened 2022-08-30
Added PHP script support inside html files [Optional]

dhrubasaha08:
Added option to enable PHP scripts support inside html files by processing the html files with PHP engine.


covener:
PHP project doesn't recommend AddType for some time: https://www.php.net/manual/en/install.unix.apache2.php
(addtype is for actual content types)

However, I still don't think we should have example AddHandler/SetHandler for a third-party module in the default configuration.

dhrubasaha08:
okay thanks !

PR #329, opened 2022-09-15
Bring in mod_http2 v2.0.4 from github

icing:
The 2.4.x/trunk and standalone mod_h2 sources have diverged while
the pollset changes have been verified in several installations from the mod_h2
github repository.

The current state of trunk with its new request/response meta bucket is not suitable
for backport into 2.4.x. But with reports of people using the mod_h2 in production,
I feel comfortable to bring these changes into the 2.4.x branch.

Changes from the github mod_h2 repository:

- new directive `H2HeaderStrictness (highest|rfc7540|rfc9113)` to
       configure the strictness of HTTP header checks. rfc9113 forbids leading
       and trailing whitespace in headers and is supported when building
       against nghttp2 v1.50.0 or newer. The default is the backward
       compatible behavior, e.g. rfc7540.
- new directive 'H2StreamTimeout' to configure a separate value for
       HTTP/2 streams, overriding server's 'Timeout' configuration. [rpluem]
- HTTP/2 connections now use pollsets to monitor the status of the
       ongoing streams and their main connection, if the platform supports
       them and APR >= 1.6.0 is in play.
- H2SerializeHeaders no longer has an effect. A warning is logged when
       it is set to "on". The switch enabled the internal writing of requests
       to be parsed by the internal HTTP/1.1 protocol handler and was
       introduced to avoid potential incompatibilities during the introduction
       of HTTP/2.
- Removed the abort/redo of tasks when mood swings lower the active
       limit. This probably hurts the server more than it helps. Some
       setup, like proxied connections, can be slow to respond to the
       abort of a connection.
- a bug that prevented trailers (e.g. HEADER frame at the end) to be
       generated in certain cases was fixed. See #233 where it prevented
       gRPC responses to be properly generated.


notroj:
Since 2.4.x already has v2.0.22 I'm assuming this is safe to close.

PR #331, opened 2022-10-17
mod_http2 v2.0.10

icing:
Version 2.0.10 of `mod_http2` plus all test case changes.


notroj:
Fixed in 9767274b884a110e9244f59f50bd31ff1cae2933

PR #332, opened 2022-10-20
Add WebAssembly MIME type and file extension

dhodder:
Please add the WASM media type to httpd, along with its associated file extension. Many thanks.

* IANA registration: https://www.iana.org/assignments/media-types/application/wasm

See also:

* WASM in CPython: https://github.com/python/cpython/blob/main/Lib/mimetypes.py
* WASM in nginx: https://github.com/nginx/nginx/blob/master/conf/mime.types


notroj:
Thanks for the PR!

PR #333, opened 2022-10-31
Fix seperator in mod_session.c

caoyebo:
In the previous commit fixing CVE-2021-26690 (67bd9bfe6c38831e14fe7122f1d84391472498f8), it introduced a seperator mistake.

https://github.com/apache/httpd/commit/67bd9bfe6c38831e14fe7122f1d84391472498f8

Though there is no exploit on this and it does not impact the functioning, the seperator in line 419 should be psep instead of sep because "=" is the seperator for pair. To prevent further issues, it should be fixed.


bigio:
Code has been committed, could you please close the PR ?

caoyebo:
Definitely! Thank you : )

PR #334, opened 2022-11-07
mod_remoteip: clear incoming RemoteIPProxiesHeader by default

erkia:
[RemoteIPProxiesHeader](https://httpd.apache.org/docs/2.4/mod/mod_remoteip.html#remoteipproxiesheader) can be used to collect addresses of intermediate trusted proxies. Unfortunately, it is impossible to distinguish, if this header was set by mod_remoteip (can be trusted) or was set in a direct request, which did not pass any trusted proxies.

This change makes sure that RemoteIPProxiesHeader, if configured, is only set by mod_remoteip and is cleared for any requests, that do not come from trusted proxy.

### DIFF
--- a/modules/metadata/mod_remoteip.c
+++ b/modules/metadata/mod_remoteip.c
@@ -563,6 +563,13 @@ static int remoteip_modify_request(request_rec *r)
         return OK;
     }
 
+    /* Clear incoming RemoteIPProxiesHeader by default,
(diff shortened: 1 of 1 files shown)


AlexAT:
Did you try **RequestHeader unset _headername_ early** to unset the header before processing takes place?

PR #341, opened 2023-02-03
Allow wildcards in Filters

blar:
Allow `AddOutputFilterByType DEFLATE text/*`

### DIFF
--- a/modules/filters/mod_filter.c
+++ b/modules/filters/mod_filter.c
@@ -180,8 +180,7 @@ static int filter_lookup(ap_filter_t *f, ap_filter_rec_t *filter)
                 }
                 else if (sep                                     
                     && strncmp(*type, r->content_type, sep) == 0     
-                    && strncmp(*type + sep, wildcard, strlen(wildcard)) == 0
-                    && (*type)[sep + strlen(wildcard)] == '\0') {                
+                    && strcmp(*type + sep, wildcard) == 0) {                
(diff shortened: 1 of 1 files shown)


modules/filters/mod_filter.c
```diff
@@ -176,6 +177,14 @@ static int filter_lookup(ap_filter_t *f, ap_filter_rec_t *filter)
                     match = 1;
                     break;
                 }
+                else if (strncmp(*type, r->content_type, sep) == 0
```
covener: maybe sep && strncmp ...to be safe here? Even though the source is from the config

modules/filters/mod_filter.c
```diff
@@ -176,6 +178,15 @@ static int filter_lookup(ap_filter_t *f, ap_filter_rec_t *filter)
                     match = 1;
                     break;
                 }
+                else if (sep                                     
+                    && strncmp(*type, r->content_type, sep) == 0     
+                    && strncmp(*type + sep, wildcard, strlen(wildcard)) == 0
+                    && (*type)[sep + strlen(wildcard)] == '\0') {
```
covener: IDK if it compiles away, but I think the 2nd strncmp could just be strcmp allowing you to drop the last condition (and both strlens)

covener (changes requested):

notroj (approved):
Approving for CI run.

notroj (approved):

PR #343, opened 2023-02-16
[cmake] Fix check for debug static build of pcre2

tobil4sk:
The correct name is `pcre2-8-staticd.lib`, not `pcre2-8d-static.lib`.

If the static debug build is the only one available, the value of `default_pcre_libraries` would previously default to `pcre.lib`, and so there was a linking error if this library did not exist.

Now the library will be correctly detected, which avoids errors if the static debug build is the only one available.


tobil4sk:
Any chance this could be merged? This can be verified in the pcre2 cmake file:

The library name is `pcre2-8-static`:
https://github.com/PCRE2Project/pcre2/blob/44d95814f6414d40de6300817ae9d2bec0babc39/CMakeLists.txt#L703

And the `d` suffix comes after the library name, from `CMAKE_DEBUG_POSTFIX`:
https://github.com/PCRE2Project/pcre2/blob/44d95814f6414d40de6300817ae9d2bec0babc39/CMakeLists.txt#L584

PR #349, opened 2023-03-19
Rerun failing

covener:
re-run failing tests with -v so they can be debugged by the job output

currently not saving and restoring t/logs/error_log. They are ~600MB and don't seem to upload as artifacts reliably.
I rarely find the error_log too useful for test failures anyway.  AORN, it would be the t/logs/error_log from the re-run.


covener:
@ylavic @notroj wdyt? Could we just run the basic perl framework invocation with -v?
We already have an enormous amount of output for casually browsing.

ylavic:
+1, the error_log for the re-run of the failing tests is great as artefacts

notroj:
Very nice idea. Thanks @covener

PR #353, opened 2023-03-26
Json log for v2.4

thomasmey:


notroj (approved):
Interesting idea, thanks for the submission! Some comments:

- some code style issues and C++-style comments, our style guide is at https://httpd.apache.org/dev/styleguide.html
- it looks like you need to now make the log_format_item struct visible outside mod_log_config so it's usable in the new module, but I don't see where that is happening? it shouldn't be exported as void * anyway

Approving so we get a CI run.

notroj (approved):

notroj (approved):

notroj (approved):

modules/loggers/mod_log_config.c
```diff
@@ -179,19 +179,32 @@ module AP_MODULE_DECLARE_DATA log_config_module;
...
                            int *strl,
                            int nelts,
+                           void *items,
                            apr_size_t len);
+static apr_status_t ap_json_log_writer(request_rec *r,
+                           void *handle,
+                           const char **strs,
+                           int *strl,
+                           int nelts,
+                           void *items,
```
ylavic: `items` shouldn't be changed by the writers so `const void *items` for the 3 above?

modules/loggers/mod_log_config.c
```diff
@@ -285,6 +298,7 @@ typedef struct {
  */
 
 typedef struct {
+    char *tag; /* tag that did create this lfi */
```
ylavic: `const char *tag`, ditto for the existing `char *arg` below though it does not originate from this commit.

modules/loggers/mod_log_config.c
```diff
@@ -1602,13 +1634,204 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+    } else if((c & 0b11000000) == 0b11000000) { // 2 byte utf-8
+        code_point = c & 0b00011111;
+        code_point <<= 5;
+        expect_more_bytes = 1;
+        len++;
+    } else if((c & 0b10000000) == 0b10000000) { // 1 byte utf-8 continuation
+        // this is an actual error, probably wrong idx
+        // idx did point to a byte which looks like an utf-8 1 byte continuation
+        return -1;
+    } else if(c < 0x80) { // 1 byte utf-8
```
ylavic: Possibly this most common case should be tested first?

modules/loggers/mod_log_config.c
```diff
@@ -1602,13 +1634,204 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+    apr_uint32_t code_point;
+    int idx = 0;
+    int len = utf8_next_code_point(string, idx, &code_point);
+
+    while(len > 0) {
+        if(code_point  < 0x20 ||
+           code_point == 0x22 ||
+           code_point == 0x5c) {
+            return 1; // true, this json utf-8 string, needs encoding
+        }
```
ylavic: There could be a json_code_point_needs_encoding() helper since this same check is in json_encode() below too.

modules/loggers/mod_log_config.c
```diff
@@ -1602,13 +1634,204 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
 
+{
+    log_format_item *items = (log_format_item *) itms;
+    apr_size_t len_file_write;
+
+    const char* attribute_name;
+    const char* attribute_value;
+    const char* json_str;
+
+    struct iovec strcats[1 + (nelts * 7) ];
```
ylavic: VLA (Variable Length Array) should be avoided in httpd, you could use an `apr_array_header_t` of `struct iovec` and modify the add_iovec_str() to push there instead.

modules/loggers/mod_log_config.c
```diff
@@ -1602,13 +1634,204 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+        if(json_needs_encoding(attribute_value)) {
+            attribute_value = json_encode(r->pool, attribute_value);
+        }
+        add_iovec_str(&strcats[isc++], attribute_value);
+        add_iovec_str(&strcats[isc++], "\",");
+    }
+    // replace last '",' with '"}'
+    add_iovec_str(&strcats[isc - 1], "\"}" APR_EOL_STR);
+    json_str = apr_pstrcatv(r->pool, strcats, isc, &len_file_write);
+    return apr_file_write((apr_file_t*)handle, json_str, &len_file_write);
```
ylavic: Using `apr_file_writev_full()` directly with the iovec array would avoid allocating `json_str` here.

modules/loggers/mod_log_config.c
ylavic: Maybe `0x7f` (another control character) is missing?

modules/loggers/mod_log_config.c
```diff
@@ -1602,13 +1634,204 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+    apr_size_t isc = 0;
+
+    // build json object
+    add_iovec_str(&strcats[isc++], "{");
+    for (int i = 0; i < nelts; ++i) {
+        if(items[i].tag == NULL) {
+                continue;
+        }
+
+        attribute_name = apr_hash_get(json_hash, items[i].tag, APR_HASH_KEY_STRING );
```
yakatz: Since JSON is already pretty verbose, it might be nice to have an option to make the keys be the single chars from the format definition and not be converted back to the attribute name.

modules/loggers/mod_log_config.c
```diff
@@ -1398,6 +1428,8 @@ AP_INIT_TAKE12("LogFormat", log_format, NULL, RSRC_CONF,
      "a log format string (see docs) and an optional format name"),
 AP_INIT_FLAG("BufferedLogs", set_buffered_logs_on, NULL, RSRC_CONF,
                  "Enable Buffered Logging (experimental)"),
+AP_INIT_FLAG("JsonLogs", set_json_logs_on, NULL, RSRC_CONF,
+                 "Enable JSON Logging (experimental)"),
```
yakatz: What about having both kinds of logs at once - for example, maybe beginning the `LogFormat` line with `json:` or an additional flag to the `LogFormat` directive that could specify an (arbitrary) format. This would be useful to allow migration and further output options.

thomasmey:
hi, @yakatz, @notroj and @ylavic do you prefer rebase, i.e. incorporate the fixes in the existing commits or should I add more commits to better follow the change history?

And one general question regarding the UTF-8 handling in JSON: After digging a bit more into the code I'm not sure if this is necessary at all in this form...

What charset is are the strings given to mod_log_config are in? All "extractor" function do call ap_escape_logitem() in the end which do seem to use an 8-bit index to deceide if to escape or not, so the assumption is to have an 8-bit fixed charater set at this stage?
is the charset documented somewhere?
so maybe I need to convert to UTF-8 instead of doing escaping?
and another question: why does ap_escape_logitem() return a copy of the input string even if no escpaing is done, bascially duplicates all strings in mod_log_config?

ylavic:
I (personnaly) think rebasing is fine in PRs/feature branches.

Regarding the encoding, you are right that ap_escape_logitem() should already \\-escape controls, double quote and non-ascii characters, so there is nothing left to json_needs_encoding() currently (besides `\` itself because this \\-escaping is not json compatible..).
Whether that's the right encoding to use for json I'm not sure either, there is necessarily an encoding to apply (in addition to json) if we want a preserving/reversible encoding because HTTP is not necessarily UTF-8 (could be ISO-8859-1, sometimes CP-1252..). But ap_escape_logitem() will escape valid UTF-8 (in headers for instance) which may be unfortunate.
Maybe that's the right thing to do because SIEMs / log managers may already know how to unescape ap_escape_logitem(), though probably not in json strings, but possibly we want to preserve UTF-8 if it's there already.
So possibly we want a all-in-one ap_escape_logjson() to replace ap_escape_logitem() in json logging mode, which would let valid UTF-8 through and for non-UTF-8 characters we could either:
1. use the unicode replacement character `U+FFFD` (non preserving obviously)
2. assume latin1/ISO-8859-1 and do a trivial code point to UTF-8 conversion (that's non preserving though if the source is mixed with native UTF-8, and latin1 extended control characters get no special encoding)
3. assume latin1/ISO-8859-1 and print them as `\u00xx` (visually different though semantically the same as 2.)
4. \\x-escape them, and `\` itself too then (preserving, but a lot of `\` since json requires it to be escaped already)
5. ...

I think I like 3. better (assuming ISO-8859-1 is not that insane IMO), but it's non strictly preserving. Or 4. What do you/others think?

ylavic:
> another question: why does ap_escape_logitem() return a copy of the input string even if no escpaing is done, bascially duplicates all strings in mod_log_config?

It returns a `char *`, so it can't return the `const char *str` source directly.
There could be a `const char *ap_escape_logitem_c(...)` or something that returns the original string if nothing needs escaping though, but we can't change the API for the existing function.
Maybe it's something to think about for an eventual new ap_escape_logjson() function..

ylavic:
I was thinking something like this for ap_escape_logjson():
```
/* See http://bjoern.hoehrmann.de/utf-8/decoder/dfa/ for details. */
#define DECODE_UTF8_ACCEPT 0
#define DECODE_UTF8_REJECT 1
static const unsigned char decode_utf8d[] = {
    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, /* 00..1f */
    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, /* 20..3f */
    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, /* 40..5f */
    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, /* 60..7f */
    1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,9,9,9,9,9,9,9,9,9,9,9,9,9,9,9,9, /* 80..9f */
    7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7,7, /* a0..bf */
    8,8,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2, /* c0..df */
    0xa,0x3,0x3,0x3,0x3,0x3,0x3,0x3,0x3,0x3,0x3,0x3,0x3,0x4,0x3,0x3, /* e0..ef */
    0xb,0x6,0x6,0x6,0x5,0x8,0x8,0x8,0x8,0x8,0x8,0x8,0x8,0x8,0x8,0x8, /* f0..ff */
    0x0,0x1,0x2,0x3,0x5,0x8,0x7,0x1,0x1,0x1,0x4,0x6,0x1,0x1,0x1,0x1, /* s0..s0 */
    1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,0,1,0,1,1,1,1,1,1, /* s1..s2 */
    1,2,1,1,1,1,1,2,1,2,1,1,1,1,1,1,1,1,1,1,1,1,1,2,1,1,1,1,1,1,1,1, /* s3..s4 */
    1,2,1,1,1,1,1,1,1,2,1,1,1,1,1,1,1,1,1,1,1,1,1,3,1,3,1,1,1,1,1,1, /* s5..s6 */
    1,3,1,1,1,1,1,3,1,3,1,1,1,1,1,1,1,3,1,1,1,1,1,1,1,1,1,1,1,1,1,1, /* s7..s8 */
};
static APR_INLINE apr_uint32_t decode_utf8_byte(apr_uint32_t state,
                                                apr_uint32_t *codep,
                                                unsigned char c)
{
    apr_uint32_t byte = c;
    apr_uint32_t type = decode_utf8d[byte];

    if (codep) {
        *codep = (state != DECODE_UTF8_ACCEPT) ?
            (byte & 0x3fu) | (*codep << 6) :
            (0xff >> type) & (byte);
    }

    return decode_utf8d[256 + state * 16 + type];
}

static APR_INLINE unsigned char *escape_json_byte(unsigned char d[6],
                                                  unsigned char c)
{
    apr_uint32_t byte = c;
    d[0] = '\\';
    d[1] = 'u';
    d[2] = '0';
    d[3] = '0';
    d[4] = c2x_table[byte >> 4];
    d[5] = c2x_table[byte & 15];
    return d + 6;
}

AP_DECLARE(const char *) ap_escape_logjson(apr_pool_t *p, const char *str,
                                           apr_size_t *len, int quoted)
{
    const unsigned char *s = (const unsigned char *)str;
    apr_size_t seq, pos, length, escapes = 0;
    apr_uint32_t state, prev;
    unsigned char *d;
    char *ret;

    /* Compute how many characters need to be escaped */
    state = prev = DECODE_UTF8_ACCEPT;
    length = (len && *len != APR_SIZE_MAX) ? *len : (str) ? strlen(str) : 0;
    for (seq = pos = 0; pos < length; ++pos, prev = state) {
        unsigned char c = s[pos];
        state = decode_utf8_byte(state, NULL, c);
        switch (state) {
        case DECODE_UTF8_ACCEPT:
            switch (c) {
            case '\b': case '\t': case '\n':
            case '\f': case '\r': case '\\':
            case '"':
                /* \-escape these */
                escapes++;
                break;
            default:
                /* \u-escape controls (U+0000 through U+001F) */
                if (c < 0x20) {
                    escapes += 5;
                }
            }
            seq = pos + 1;
            break;

        case DECODE_UTF8_REJECT:
            if (prev == DECODE_UTF8_ACCEPT) {
                /* \u-escape invalid [pos] (only) */
                escapes += 5;
                seq++;
            }
            else {
                /* \u-escape each byte of invalid [seq,pos[ sequence and
                 * let above decide for [pos]
                 */
                if (seq < pos) {
                    escapes += (pos - seq) * 5;
                    seq = pos;
                }
                pos--;
            }
            state = DECODE_UTF8_ACCEPT;
            break;

        default:
            /* valid continuation byte */
            break;
        }
    }
    if (seq < length) {
        /* \u-escape each byte of unterminated [seq,length[ sequence */
        escapes += (length - seq) * 5;
    }

    /* Fast path: nothing to escape */
    if (escapes == 0) {
        if (quoted) {
            if (str) {
                str = apr_pstrcat(p, "\"", str, "\"", NULL);
                length += 2;
            }
            else {
                str = "null";
                length = 4;
            }
        }
        if (len) {
            *len = length;
        }
        return str;
    }

    /* Do the actual escaping in a suitable buffer */
    state = prev = DECODE_UTF8_ACCEPT;
    ret = apr_palloc(p, length + escapes + (quoted ? 2 : 0) + 1);
    d = (unsigned char *)ret;
    if (quoted) {
        *d++ = '"';
    }
    for (seq = pos = 0; pos < length; ++pos, prev = state) {
        unsigned char c = s[pos];
        state = decode_utf8_byte(state, NULL, c);
        switch (state) {
        case DECODE_UTF8_ACCEPT:
            switch (c) {
            case '\b':
                *d++ = '\\';
                *d++ = 'b';
                break;
            case '\t':
                *d++ = '\\';
                *d++ = 't';
                break;
            case '\n':
                *d++ = '\\';
                *d++ = 'n';
                break;
            case '\f':
                *d++ = '\\';
                *d++ = 'f';
                break;
            case '\r':
                *d++ = '\\';
                *d++ = 'r';
                break;
            case '\\':
            case '"':
                *d++ = '\\';
                *d++ = c;
                break;
            default:
                if (c < 0x20) {
                    /* control to \u-escape */
                    d = escape_json_byte(d, c);
                }
                else {
                    /* valid (multi-)byte as is */
                    while (seq < pos) {
                        *d++ = s[seq++];
                    }
                    *d++ = c;
                }
            }
            seq++;
            break;

        case DECODE_UTF8_REJECT:
            if (prev == DECODE_UTF8_ACCEPT) {
                d = escape_json_byte(d, c);
                seq++;
            }
            else {
                while (seq < pos) {
                    d = escape_json_byte(d, s[seq++]);
                }
                pos--;
            }
            state = DECODE_UTF8_ACCEPT;
            break;
        }
    }
    while (seq < length) {
        d = escape_json_byte(d, s[seq++]);
    }
    if (quoted) {
        *d++ = '"';
    }
    if (len) {
        *len = d - (const unsigned char *)ret;
    }
    *d = '\0';

    return ret;
}
```
Feel free to take/modify/ignore (only compile-tested..).

modules/loggers/mod_log_config.c
thomasmey: Strangley 0x7f is not part of the codepoints that needs escaping in JSON strings, see https://www.rfc-editor.org/rfc/rfc8259#page-9
Searching for JSON and 0x7f seems to at least reveal on bug report https://github.com/dominictarr/JSON.sh/issues/46

thomasmey:
> I (personnaly) think rebasing is fine in PRs/feature branches.
> 
> Regarding the encoding, you are right that ap_escape_logitem() should already \-escape controls, double quote and non-ascii characters, so there is nothing left to json_needs_encoding() currently (besides `\` itself because this \-escaping is not json compatible..). Whether that's the right encoding to use for json I'm not sure either, there is necessarily an encoding to apply (in addition to json) if we want a preserving/reversible encoding because HTTP is not necessarily UTF-8 (could be ISO-8859-1, sometimes CP-1252..). But ap_escape_logitem() will escape valid UTF-8 (in headers for instance) which may be unfortunate. Maybe that's the right thing to do because SIEMs / log managers may already know how to unescape ap_escape_logitem(), though probably not in json strings, but possibly we want to preserve UTF-8 if it's there already. So possibly we want a all-in-one ap_escape_logjson() to replace ap_escape_logitem() in json logging mode, which would let valid UTF-8 through and for non-UTF-8 characters we could either:
> 
>     1. use the unicode replacement character `U+FFFD` (non preserving obviously)
> 
>     2. assume latin1/ISO-8859-1 and do a trivial code point to UTF-8 conversion (that's non preserving though if the source is mixed with native UTF-8, and latin1 extended control characters get no special encoding)
> 
>     3. assume latin1/ISO-8859-1 and print them as `\u00xx` (visually different though semantically the same as 2.)
> 
>     4. \x-escape them, and `\` itself too then (preserving, but a lot of `\` since json requires it to be escaped already)
> 
>     5. ...
> 
> 
> I think I like 3. better (assuming ISO-8859-1 is not that insane IMO), but it's non strictly preserving. Or 4. What do you/others think?

Hi,

cool, I had similar ideas to solve this problem, very nice!

I will try to work on all the comments here the next days and thanks for your valuable feedback very appreciated!

But be a bit patient with me as this is purely a hobby project I do in my spare time and depending how stressful my day ways and when the little one falls asleep it may take some time 😄

modules/loggers/mod_log_config.c
ylavic: I think `0x7f` is covered in https://www.rfc-editor.org/rfc/rfc8259#section-7
```
   All Unicode characters may be placed within the
   quotation marks, except for the characters that MUST be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).
```

ylavic: Hm, scratch that, it stops at U+001F (not U+007F). TIL..

thomasmey:
>     * some code style issues and C++-style comments, our style guide is at https://httpd.apache.org/dev/styleguide.html

running GNU indent on mod_log_config.[ch] (as described in the styleguide) gives interessting results... looks like nobody seems to run indent before each commit; I assume that probably no git hooks, no automation are in place.

I can add one last "GNU ident"-commit in this PR to comply with styleguide rules, but the fallout will be big and will not cover only my changes, if you like.

thomasmey:
some more ideas inspired by @yakatz. please have a look, it can be activated with:

CustomLog "logs/access_log" combined formatter=json
CustomLog "logs/access_log" combined formatter=json,short

@ylavic I moved json escape function to where is fits better, but the actual escape function you proposed is still a todo, I'll work on this next.

yakatz:
> `CustomLog "logs/access_log" combined formatter=json`
> `CustomLog "logs/access_log" combined formatter=json,short`

Looks good to me

thomasmey:
Hi, any update ls on this?
Something else I could do for this PR?

modules/loggers/mod_log_config.c
```diff
@@ -350,9 +367,6 @@ static const char *log_remote_user(request_rec *r, char *a)
     else if (strlen(rvalue) == 0) {
         rvalue = "\"\"";
```
ylavic, line 368: This needs to go in the caller probably, we don't need it in json for instance (where `""` is better than `"\"\""` as a value). See my comment in config_log_transaction().

modules/loggers/mod_log_config.c
```diff
@@ -1165,16 +1180,28 @@ static int config_log_transaction(request_rec *r, config_log_state *cls,
...
 
     for (i = 0; i < format->nelts; ++i) {
-        strs[i] = process_item(r, orig, &items[i]);
-        len += strl[i] = strlen(strs[i]);
+        lfd->portions[i] = process_item(r, orig, &items[i]);
+
+        /* if we don't have a custom log formatter escape all values unconditionally,
+         * else let the custom log formatter handle the escaping
+         */
+        if (!cls->log_formatter) {
```
ylavic, line 1188: We don't want empty parts here, so possibly:
```
    if (*lfd->portions[i]) {
        lfd->portions[i] = ap_escape_logitem(r->pool, lfd->portions[i])
    }
    else {
        lfd->portions[i] = "\"\"";
    }
```
?

modules/loggers/mod_log_config.c
```diff
@@ -263,7 +277,9 @@ typedef struct {
     const char *fname;
     const char *format_string;
     apr_array_header_t *format;
-    void *log_writer;
+    void *log_writer; /* log_writer_data */
```
ylavic, line 280: Maybe do the renaming then? `log_writer` the function vs `log_writer` the data is quite confusing..

ylavic:
It looks quite good to me, but it should be based on `trunk` rather than `2.4.x` (nothing goes to 2.4 without making it to trunk first). Not sure if the target branch can be changed without creating a new PR, sorry for noticing this only now :/

server/util.c
```diff
@@ -2190,6 +2190,219 @@ AP_DECLARE(char *) ap_escape_html2(apr_pool_t *p, const char *s, int toasc)
     x[j] = '\0';
     return x;
 }
+
+/* TODO: not sure about copyright&license of below code */
```
ylavic, line 2194: This code is provided under the MIT license IIUC, so a note in the /LICENSE file should be fine I think. We can probably do that when merging.

modules/loggers/mod_log_config.c
ylavic: Maybe even NULL should be handled here, like:
```
    if (!lfd->portions[i]) {
        lfd->portions[i] = "-";
    }
    else if (!*lfd->portions[i]) {
        lfd->portions[i] = "\"\"";
    }
    else {
        lfd->portions[i] = ap_escape_logitem(r->pool, lfd->portions[i])
    }
```
This would let the formatter handle it (using the `null` value in json for instance).

thomasmey:
> It looks quite good to me, but it should be based on `trunk` rather than `2.4.x` (nothing goes to 2.4 without making it to trunk first). Not sure if the target branch can be changed without creating a new PR, sorry for noticing this only now :/

updated trunk version: https://github.com/apache/httpd/pull/373

OkayAndUTry:
Only 3 escape youngest child.

OkayAndUTry:
Attribution-NonCommercial-ShareAlike 4.0 International ======================================================================= Creative Commons Corporation ("Creative Commons") is not a law firm and does not provide legal services or legal advice. Distribution of Creative Commons public licenses does not create a lawyer-client or other relationship. Creative Commons makes its licenses and related information available on an "as-is" basis. Creative Commons gives no warranties regarding its licenses, any material licensed under their terms and conditions, or any related information. Creative Commons disclaims all liability for damages resulting from their use to the fullest extent possible. Using Creative Commons Public Licenses Creative Commons public licenses provide a standard set of terms and conditions that creators and other rights holders may use to share original works of authorship and other material subject to copyright and certain other rights specified in the public license below. The following considerations are for informational purposes only, are not exhaustive, and do not form part of our licenses. Considerations for licensors: Our public licenses are intended for use by those authorized to give the public permission to use material in ways otherwise restricted by copyright and certain other rights. Our licenses are irrevocable. Licensors should read and understand the terms and conditions of the license they choose before applying it. Licensors should also secure all rights necessary before applying our licenses so that the public can reuse the material as expected. Licensors should clearly mark any material not subject to the license. This includes other CC- licensed material, or material used under an exception or limitation to copyright. More considerations for licensors: wiki.creativecommons.org/Considerations_for_licensors Considerations for the public: By using one of our public licenses, a licensor grants the public permission to use the licensed material under specified terms and conditions. If the licensor's permission is not necessary for any reason--for example, because of any applicable exception or limitation to copyright--then that use is not regulated by the license. Our licenses grant only permissions under copyright and certain other rights that a licensor has authority to grant. Use of the licensed material may still be restricted for other reasons, including because others have copyright or other rights in the material. A licensor may make special requests, such as asking that all changes be marked or described. Although not required by our licenses, you are encouraged to respect those requests where reasonable. More considerations for the public: wiki.creativecommons.org/Considerations_for_licensees ======================================================================= Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International Public License By exercising the Licensed Rights (defined below), You accept and agree to be bound by the terms and conditions of this Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International Public License ("Public License"). To the extent this Public License may be interpreted as a contract, You are granted the Licensed Rights in consideration of Your acceptance of these terms and conditions, and the Licensor grants You such rights in consideration of benefits the Licensor receives from making the Licensed Material available under these terms and conditions. Section 1 -- Definitions. a. Adapted Material means material subject to Copyright and Similar Rights that is derived from or based upon the Licensed Material and in which the Licensed Material is translated, altered, arranged, transformed, or otherwise modified in a manner requiring permission under the Copyright and Similar Rights held by the Licensor. For purposes of this Public License, where the Licensed Material is a musical work, performance, or sound recording, Adapted Material is always produced where the Licensed Material is synched in timed relation with a moving image. b. Adapter's License means the license You apply to Your Copyright and Similar Rights in Your contributions to Adapted Material in accordance with the terms and conditions of this Public License. c. BY-NC-SA Compatible License means a license listed at creativecommons.org/compatiblelicenses, approved by Creative Commons as essentially the equivalent of this Public License. d. Copyright and Similar Rights means copyright and/or similar rights closely related to copyright including, without limitation, performance, broadcast, sound recording, and Sui Generis Database Rights, without regard to how the rights are labeled or categorized. For purposes of this Public License, the rights specified in Section 2(b)(1)-(2) are not Copyright and Similar Rights. e. Effective Technological Measures means those measures that, in the absence of proper authority, may not be circumvented under laws fulfilling obligations under Article 11 of the WIPO Copyright Treaty adopted on December 20, 1996, and/or similar international agreements. f. Exceptions and Limitations means fair use, fair dealing, and/or any other exception or limitation to Copyright and Similar Rights that applies to Your use of the Licensed Material. g. License Elements means the license attributes listed in the name of a Creative Commons Public License. The License Elements of this Public License are Attribution, NonCommercial, and ShareAlike. h. Licensed Material means the artistic or literary work, database, or other material to which the Licensor applied this Public License. i. Licensed Rights means the rights granted to You subject to the terms and conditions of this Public License, which are limited to all Copyright and Similar Rights that apply to Your use of the Licensed Material and that the Licensor has authority to license. j. Licensor means the individual(s) or entity(ies) granting rights under this Public License. k. NonCommercial means not primarily intended for or directed towards commercial advantage or monetary compensation. For purposes of this Public License, the exchange of the Licensed Material for other material subject to Copyright and Similar Rights by digital file-sharing or similar means is NonCommercial provided there is no payment of monetary compensation in connection with the exchange. l. Share means to provide material to the public by any means or process that requires permission under the Licensed Rights, such as reproduction, public display, public performance, distribution, dissemination, communication, or importation, and to make material available to the public including in ways that members of the public may access the material from a place and at a time individually chosen by them. m. Sui Generis Database Rights means rights other than copyright resulting from Directive 96/9/EC of the European Parliament and of the Council of 11 March 1996 on the legal protection of databases, as amended and/or succeeded, as well as other essentially equivalent rights anywhere in the world. n. You means the individual or entity exercising the Licensed Rights under this Public License. Your has a corresponding meaning. Section 2 -- Scope. a. License grant. 1. Subject to the terms and conditions of this Public License, the Licensor hereby grants You a worldwide, royalty-free, non-sublicensable, non-exclusive, irrevocable license to exercise the Licensed Rights in the Licensed Material to: a. reproduce and Share the Licensed Material, in whole or in part, for NonCommercial purposes only; and b. produce, reproduce, and Share Adapted Material for NonCommercial purposes only. 2. Exceptions and Limitations. For the avoidance of doubt, where Exceptions and Limitations apply to Your use, this Public License does not apply, and You do not need to comply with its terms and conditions. 3. Term. The term of this Public License is specified in Section 6(a). 4. Media and formats; technical modifications allowed. The Licensor authorizes You to exercise the Licensed Rights in all media and formats whether now known or hereafter created, and to make technical modifications necessary to do so. The Licensor waives and/or agrees not to assert any right or authority to forbid You from making technical modifications necessary to exercise the Licensed Rights, including technical modifications necessary to circumvent Effective Technological Measures. For purposes of this Public License, simply making modifications authorized by this Section 2(a) (4) never produces Adapted Material. 5. Downstream recipients. a. Offer from the Licensor -- Licensed Material. Every recipient of the Licensed Material automatically receives an offer from the Licensor to exercise the Licensed Rights under the terms and conditions of this Public License. b. Additional offer from the Licensor -- Adapted Material. Every recipient of Adapted Material from You automatically receives an offer from the Licensor to exercise the Licensed Rights in the Adapted Material under the conditions of the Adapter's License You apply. c. No downstream restrictions. You may not offer or impose any additional or different terms or conditions on, or apply any Effective Technological Measures to, the Licensed Material if doing so restricts exercise of the Licensed Rights by any recipient of the Licensed Material. 6. No endorsement. Nothing in this Public License constitutes or may be construed as permission to assert or imply that You are, or that Your use of the Licensed Material is, connected with, or sponsored, endorsed, or granted official status by, the Licensor or others designated to receive attribution as provided in Section 3(a)(1)(A)(i). b. Other rights. 1. Moral rights, such as the right of integrity, are not licensed under this Public License, nor are publicity, privacy, and/or other similar personality rights; however, to the extent possible, the Licensor waives and/or agrees not to assert any such rights held by the Licensor to the limited extent necessary to allow You to exercise the Licensed Rights, but not otherwise. 2. Patent and trademark rights are not licensed under this Public License. 3. To the extent possible, the Licensor waives any right to collect royalties from You for the exercise of the Licensed Rights, whether directly or through a collecting society under any voluntary or waivable statutory or compulsory licensing scheme. In all other cases the Licensor expressly reserves any right to collect such royalties, including when the Licensed Material is used other than for NonCommercial purposes. Section 3 -- License Conditions. Your exercise of the Licensed Rights is expressly made subject to the following conditions. a. Attribution. 1. If You Share the Licensed Material (including in modified form), You must: a. retain the following if it is supplied by the Licensor with the Licensed Material: i. identification of the creator(s) of the Licensed Material and any others designated to receive attribution, in any reasonable manner requested by the Licensor (including by pseudonym if designated); ii. a copyright notice; iii. a notice that refers to this Public License; iv. a notice that refers to the disclaimer of warranties; v. a URI or hyperlink to the Licensed Material to the extent reasonably practicable; b. indicate if You modified the Licensed Material and retain an indication of any previous modifications; and c. indicate the Licensed Material is licensed under this Public License, and include the text of, or the URI or hyperlink to, this Public License. 2. You may satisfy the conditions in Section 3(a)(1) in any reasonable manner based on the medium, means, and context in which You Share the Licensed Material. For example, it may be reasonable to satisfy the conditions by providing a URI or hyperlink to a resource that includes the required information. 3. If requested by the Licensor, You must remove any of the information required by Section 3(a)(1)(A) to the extent reasonably practicable. b. ShareAlike. In addition to the conditions in Section 3(a), if You Share Adapted Material You produce, the following conditions also apply. 1. The Adapter's License You apply must be a Creative Commons license with the same License Elements, this version or later, or a BY-NC-SA Compatible License. 2. You must include the text of, or the URI or hyperlink to, the Adapter's License You apply. You may satisfy this condition in any reasonable manner based on the medium, means, and context in which You Share Adapted Material. 3. You may not offer or impose any additional or different terms or conditions on, or apply any Effective Technological Measures to, Adapted Material that restrict exercise of the rights granted under the Adapter's License You apply. Section 4 -- Sui Generis Database Rights. Where the Licensed Rights include Sui Generis Database Rights that apply to Your use of the Licensed Material: a. for the avoidance of doubt, Section 2(a)(1) grants You the right to extract, reuse, reproduce, and Share all or a substantial portion of the contents of the database for NonCommercial purposes only; b. if You include all or a substantial portion of the database contents in a database in which You have Sui Generis Database Rights, then the database in which You have Sui Generis Database Rights (but not its individual contents) is Adapted Material, including for purposes of Section 3(b); and c. You must comply with the conditions in Section 3(a) if You Share all or a substantial portion of the contents of the database. For the avoidance of doubt, this Section 4 supplements and does not replace Your obligations under this Public License where the Licensed Rights include other Copyright and Similar Rights. Section 5 -- Disclaimer of Warranties and Limitation of Liability. a. UNLESS OTHERWISE SEPARATELY UNDERTAKEN BY THE LICENSOR, TO THE EXTENT POSSIBLE, THE LICENSOR OFFERS THE LICENSED MATERIAL AS-IS AND AS-AVAILABLE, AND MAKES NO REPRESENTATIONS OR WARRANTIES OF ANY KIND CONCERNING THE LICENSED MATERIAL, WHETHER EXPRESS, IMPLIED, STATUTORY, OR OTHER. THIS INCLUDES, WITHOUT LIMITATION, WARRANTIES OF TITLE, MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, NON-INFRINGEMENT, ABSENCE OF LATENT OR OTHER DEFECTS, ACCURACY, OR THE PRESENCE OR ABSENCE OF ERRORS, WHETHER OR NOT KNOWN OR DISCOVERABLE. WHERE DISCLAIMERS OF WARRANTIES ARE NOT ALLOWED IN FULL OR IN PART, THIS DISCLAIMER MAY NOT APPLY TO YOU. b. TO THE EXTENT POSSIBLE, IN NO EVENT WILL THE LICENSOR BE LIABLE TO YOU ON ANY LEGAL THEORY (INCLUDING, WITHOUT LIMITATION, NEGLIGENCE) OR OTHERWISE FOR ANY DIRECT, SPECIAL, INDIRECT, INCIDENTAL, CONSEQUENTIAL, PUNITIVE, EXEMPLARY, OR OTHER LOSSES, COSTS, EXPENSES, OR DAMAGES ARISING OUT OF THIS PUBLIC LICENSE OR USE OF THE LICENSED MATERIAL, EVEN IF THE LICENSOR HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH LOSSES, COSTS, EXPENSES, OR DAMAGES. WHERE A LIMITATION OF LIABILITY IS NOT ALLOWED IN FULL OR IN PART, THIS LIMITATION MAY NOT APPLY TO YOU. c. The disclaimer of warranties and limitation of liability provided above shall be interpreted in a manner that, to the extent possible, most closely approximates an absolute disclaimer and waiver of all liability. Section 6 -- Term and Termination. a. This Public License applies for the term of the Copyright and Similar Rights licensed here. However, if You fail to comply with this Public License, then Your rights under this Public License terminate automatically. b. Where Your right to use the Licensed Material has terminated under Section 6(a), it reinstates: 1. automatically as of the date the violation is cured, provided it is cured within 30 days of Your discovery of the violation; or 2. upon express reinstatement by the Licensor. For the avoidance of doubt, this Section 6(b) does not affect any right the Licensor may have to seek remedies for Your violations of this Public License. c. For the avoidance of doubt, the Licensor may also offer the Licensed Material under separate terms or conditions or stop distributing the Licensed Material at any time; however, doing so will not terminate this Public License. d. Sections 1, 5, 6, 7, and 8 survive termination of this Public License. Section 7 -- Other Terms and Conditions. a. The Licensor shall not be bound by any additional or different terms or conditions communicated by You unless expressly agreed. b. Any arrangements, understandings, or agreements regarding the Licensed Material not stated herein are separate from and independent of the terms and conditions of this Public License. Section 8 -- Interpretation. a. For the avoidance of doubt, this Public License does not, and shall not be interpreted to, reduce, limit, restrict, or impose conditions on any use of the Licensed Material that could lawfully be made without permission under this Public License. b. To the extent possible, if any provision of this Public License is deemed unenforceable, it shall be automatically reformed to the minimum extent necessary to make it enforceable. If the provision cannot be reformed, it shall be severed from this Public License without affecting the enforceability of the remaining terms and conditions. c. No term or condition of this Public License will be waived and no failure to comply consented to unless expressly agreed to by the Licensor. d. Nothing in this Public License constitutes or may be interpreted as a limitation upon, or waiver of, any privileges and immunities that apply to the Licensor or You, including from the legal processes of any jurisdiction or authority. ======================================================================= Creative Commons is not a party to its public licenses. Notwithstanding, Creative Commons may elect to apply one of its public licenses to material it publishes and in those instances will be considered the “Licensor.” The text of the Creative Commons public licenses is dedicated to the public domain under the CC0 Public Domain Dedication. Except for the limited purpose of indicating that material is shared under a Creative Commons public license or as otherwise permitted by the Creative Commons policies published at creativecommons.org/policies, Creative Commons does not authorize the use of the trademark "Creative Commons" or any other trademark or logo of Creative Commons without its prior written consent including, without limitation, in connection with any unauthorized modifications to any of its public licenses or any other arrangements, understandings, or agreements concerning use of licensed material. For the avoidance of doubt, this paragraph does not form part of the public licenses. Creative Commons may be contacted at creativecommons.org.

PR #355, opened 2023-04-11
Fix a possible NULL pointer dereference in hook_uri2file

uhliarik:
is_absolute_uri() dereferences r->filename in the function body, but there is no check before if r->filename == NULL.

### DIFF
--- a/modules/mappers/mod_rewrite.c
+++ b/modules/mappers/mod_rewrite.c
@@ -4823,8 +4823,8 @@ static int hook_uri2file(request_rec *r)
     }
 
     if (rulestatus) {
-        unsigned skip_absolute = is_absolute_uri(r->filename, NULL);
         apr_size_t flen =  r->filename ? strlen(r->filename) : 0;
+        unsigned skip_absolute = flen ? is_absolute_uri(r->filename, NULL) : 0;
         int to_proxyreq = (flen > 6 && strncmp(r->filename, "proxy:", 6) == 0);
         int will_escape = skip_absolute && (rulestatus != ACTION_NOESCAPE);
 


uhliarik:
I'm not sure if r->filename can ever be NULL in this `if (rulestatus) {` block, but I think we should be cautious here since we check r->filename for NULL on the next line ...

covener (approved):

notroj (approved):

ylavic (approved):
Three approvals to get ci started?

ngie-eign:
Hi maintainers!
Is this a potential security vulnerability (denial of service -> crash) depending on input provided by an untrusted user in r->uri, or is the input validation good enough to ensure that an attacker couldn't exploit this potential weakness? https://github.com/apache/httpd/commit/0de661765b63037c34f4aaea4bc6adc7b2524923#diff-c79653a5ec3ccb8725b180419680b328e34ab9a0a7f453208d79570448860a6aL4809 ?

ylavic:
I don't think `r->uri` (thus `r->filename` here) can be NULL when a request is being processed (see `ap_parse_request_line()::ap_parse_uri()` which always set it).
This PR is more about not dereferencing `r->filename` before checking for NULL, which a static analyser might warn about even if it cannot actually happen.

notroj:
Yes, exactly what @ylavic suggests, we saw a new static analyser warning for this change.

ngie-eign:
Yeah -- I see if now. `flen` computation after this change could result in a SIGSEGV/SIGBUS issue.
Is there any way that a malicious user could exploit this problem either before or after the change where it could result in a denial of service problem? I'm really trying to make sure there isn't a security issue, because functional issues and security issues have different priority in terms of how they are fixed at my $work.

ylavic:
> Yeah -- I see if now. `flen` computation after this change could result in a SIGSEGV/SIGBUS issue.

No, r->filename can't be possibly NULL here, per this snippet above the changed code:
```
    if (!(saved_rulestatus = apr_table_get(r->notes,"mod_rewrite_rewritten"))) {
        /* If r->filename was not initially set or if it's a pre_trans reverse
         * "proxy:" scheme, we start with the requested URI.
         */
        if (r->filename == NULL || (r->proxyreq == PROXYREQ_REVERSE &&
                                    strncmp(r->filename, "proxy:", 6) == 0)) {
            r->filename = apr_pstrdup(r->pool, r->uri);
            rewritelog(r, 2, NULL, "init rewrite engine with requested uri "
                       "%s. Original filename = %s", r->filename,
                       (ofilename) ? ofilename : "n/a");
        }
        else {
            rewritelog(r, 2, NULL, "init rewrite engine with passed filename "
                        "%s. Original uri = %s", r->filename, r->uri);
        }
        /*
         *  now apply the rules ...
         */
        rulestatus = apply_rewrite_list(r, conf->rewriterules, NULL);
        apr_table_setn(r->notes, "mod_rewrite_rewritten",
                       apr_psprintf(r->pool,"%d",rulestatus));
    }
    else {
        rewritelog(r, 2, NULL, "uri already rewritten. Status %s, Uri %s, "
                   "r->filename %s", saved_rulestatus, r->uri, r->filename);
        rulestatus = atoi(saved_rulestatus);
    }
```
I.e. if r->filename is NULL when hook_uri2file() is called it is set to r->uri, which is never NULL.

> Is there any way that a malicious user could exploit this problem either before or after the change where it could result in a denial of service problem?

So no, it wasn't a security issue before the change and it's not one either now (we just don't hear about static code analysers anymore..).

ngie-eign:
@ylavic : got it -- thanks :).
I'll maybe get around to submitting a patch to eliminate the Coverity issues and guard the code against undesirable pathological code/logic paths.

PR #356, opened 2023-04-13
Fix "Failed linking C shared library mod_md.so"

tumatanquang:
The cURL variable was declared but not assigned, resulting in a "Failed linking C shared library mod_md.so" error because cURL was not found. This change will automatically find and assign values to cURL variables, fixing the error that cURL cannot be found during HTTPD compilation without having to recompile cURL again as the solution in [this error](https://bz.apache.org/bugzilla/show_bug.cgi?id=65602).
And trim trailing spaces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Licensed to the Apache Software Foundation (ASF) under one or more
-# contributor license agreements.  See the NOTICE file distributed with
+# contributor license agreements. See the NOTICE file distributed with
 # this work for additional information regarding copyright ownership.
 # The ASF licenses this file to You under the Apache License, Version 2.0
 # (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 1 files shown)


icing:
Thanks for the PR. Admittedly, I am not using cmake and am unable to verify. I am willing to merge that. If you would then care to check the results?

tumatanquang:
> @icing Thanks for the PR. Admittedly, I am not using cmake and am unable to verify. I am willing to merge that. If you would then care to check the results?

Hi, of course! I have been using the modified `CMakeLists.txt` file as in the commit, I can provide some screenshots that HTTPd has successfully detected the cURL library file and successfully compiled without error below:
cURL detected
HTTPd compile successfully
P/s: I also tried cURL as static library and of course it didn't have any error.
I have tested and verified the modifications in the `CMakeLists.txt` file on all recent versions (include latest version), and the tools (`cmake`, `nmake`,...) and source code (HTTPd, APR, cURL,...) that I use are all the latest, all compiled successfully without errors.
It works very well, much better than the previous solution I provided which was to have to compile cURL again during HTTPd compilation (as in [bugzilla I said](https://bz.apache.org/bugzilla/show_bug.cgi?id=65602#c6)).

icing:
Submitted as r1909215 in apache subversion (trunk).

I removed the whitespace changes to focus the commit on the relevant lines and making merging this into the 2.4.x branch easier. I will propose this for backport.

tumatanquang:
> Submitted as r1909215 in apache subversion (trunk).
> 
> I removed the whitespace changes to focus the commit on the relevant lines and making merging this into the 2.4.x branch easier. I will propose this for backport.

It looks good to me. I am looking forward to these changes in the next release version. As for spaces, I can remove it manually.

tumatanquang:
I see lines 497 and 798 in the `CMakeLists.txt` file which still haven't been trim trailing spaces in the release source file.

notroj:
Looks like this was merged in 6b1f1209393a2783c5eac719c3a7a53cf521f70d probably best to file a new PR if there are outstanding changes.

PR #357, opened 2023-04-20
Make using errorlog provider and file at the same time possible

TomasKorbar:
This commit allows user to use errorlog provider and file or pipe to utility at the same time. Previous behaviour of ErrorLog directive is kept intact.

### DIFF
--- a/include/httpd.h
+++ b/include/httpd.h
@@ -1385,12 +1385,16 @@ struct server_rec {
 
     /** The name of the error log */
     char *error_fname;
+    /** true if error log file was set */
+    char errorlog_altered;
     /** A file descriptor that references the error log */
     apr_file_t *error_log;
     /** The log level configuration */
     struct ap_logconf log;
     /** External error log writer provider */
     struct ap_errorlog_provider *errorlog_provider;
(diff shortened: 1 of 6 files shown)


TomasKorbar:
Hi guys,
I tested this feature on my setup with virtual hosts and several log files and everything seems to be working.
ErrorLog directive can now be used twice with 1 file or pipe to executable and 1 use of errorlog provider. Old behaviour
should stay intact. I am open to implement whatever changes you suggest.

**Example use:**
```
ErrorLog "| /usr/bin/tee logs/error_log1 logs/error_log2"
ErrorLog syslog:local6
```
Will make httpd log into both error_log1 and error_log2 while still logging into the syslog.
This feature is usefull when you want to for example log to syslog while still keeping backup copy.

Could you guide me on how to execute the test suite?
Can i provide a test for this?

Thanks for any help.

notroj:
Tomas, if you do:

```
svn co https://svn.apache.org/repos/asf/httpd/test/framework/ test/framework
./configure --with-test-suite=test/framework
```

you can then run the test suite just by doing "make check". However there are not many tests around logging atm

PR #358, opened 2023-05-03
mod_proxy_html filter's ProxyHTMLURLMap environment variables interpolation caused startup warnings

sgreiner:
mod_proxy_html filter's ProxyHTMLURLMap environment variables interpolation causes startup warnings.

It used v2.2 ${var} notation style to reference environment variables instead of v2.4 notation style %{ENV:var} (like mod_rewrite syntax)

So it correctly caused startup warnings like "AH00111: Config variable ${var} is not defined" because, under v2.4,
the ${var} syntax is reserved for variables defined by a DEFINE directive or for environment variables existing before the server was started.

See:
- https://httpd.apache.org/docs/2.4/en/configuring.html
- https://bz.apache.org/bugzilla/show_bug.cgi?id=58467

Before:
```
RewriteRule ^/path/(foo|bar)/(.*)$ - [E=mypath:$1]
<LocationMatch "/path/(foo|bar)/.*">
	SetOutputFilter proxy-html
	ProxyHTMLInterp On
	ProxyHTMLURLMap ^/a/path/to/change/(.*)$ /newpath/${mypath}/images/$1 ecRiLV
</LocationMatch>
```

After:
```
RewriteRule ^/path/(foo|bar)/(.*)$ - [E=mypath:$1]
<LocationMatch "/path/(foo|bar)/.*">
	SetOutputFilter proxy-html
	ProxyHTMLInterp On
	ProxyHTMLURLMap ^/a/path/to/change/(.*)$ /newpath/%{ENV:mypath}/images/$1 ecRiLV
</LocationMatch>
```

This might be a breaking change for people who used to live with the startup warning.

Thank you for considering the change.


covener:
It would be nice to tolerate the old syntax so it can be backported to 2.4.x

An alternate way to do this would be to allow the second arg to be an an actual expression. Other modules do this cautiously by looking for an `expr=` prefix on the argument.

sgreiner:
I totally agree with you regarding the backport to 2.4, but I guess that requires changes to core.c to not complain if this directive uses it wrongly.
https://github.com/apache/httpd/blob/382aba75811fc71c1cf91afe992383413de3b7eb/server/core.c#LL1417C1-L1425C18
Unfortunately I am by far not experienced enough to implement this 😔

Regarding the expr: There is a forth argument to [ProxyHTMLURLMap](https://httpd.apache.org/docs/2.4/mod/mod_proxy_html.html#proxyhtmlurlmap): cond

> A cond is evaluated by the [Expression Parser](https://httpd.apache.org/docs/2.4/expr.html). In addition, the simpler syntax of conditions in mod_proxy_html 3.x for HTTPD 2.0 and 2.2 is also supported.

With this expr the map can be enabled or disabled. But inside the from-pattern or to-pattern currently no expr can be used to return a value.

sgreiner:
With my additional commit the old syntax should still work. (and will continue to emit a startup warning, just as before)

sgreiner:
Is there anything else I can do to get this merged? Thx.

sgreiner:
@covener What do I need to do to get this merged? Thank you!

PR #360, opened 2023-06-09
VS2022 doesn't like ERROR.

jfclere:
The ERROR in the code won't allow to compile with MS Visual Studio 2022


notroj:
This was fixed in ea46c590c46be521670744398cc9edfd8c51cd43

PR #362, opened 2023-06-15
Add HTTP/2 WebSockets support

icing:
Adds support for bootstrapping WebSockets via HTTP/2, as  described in RFC 8441.

 A new directive 'H2WebSockets on|off' has been  added. The feature is by default not enabled by default.

As also discussed in the manual, this feature should work for setups  using `ProxyPass backend-url upgrade=websocket` without further changes.  Special server modules for WebSockets will have to be adapted,  most likely, as the handling if IO events is different with HTTP/2.

## Sample Setup

Assume you have a HTTP/1.1 WebSockets service sitting on `localhost` port `5001`, and want to expose that underneath `/ws` for HTTP/1.1 and HTTP/2 clients on your `my-domain.org` site:

```apache

<VirtualHost *:443>
  ServerName your-domain.org
  SSLEngine on
  -.. other SSL things ...

    H2WebSockets on
    ProxyPass /ws/ http://127.0.0.1:5100/   upgrade=websocket
   ...
</VirtualHost>
```

Depending on what you do over those WebSockets, you might need to configure larger timeouts than for standard HTTP requests. There is a `timeout=seconds` parameter on the `ProxyPass` directive. And there is `H2StreamTimeout` and the general `Timeout` for all connections to tweak.

Update:
CI: due to github's ancient OS versions, we get a websockets python module installed that is "too old". I have no platform to debug this and no real will to. I added a check for a minimum version 10.4 of the module that is running on my debian sid. So, in github CI, the websockets tests in mod_http2 will skip for now.


modules/proxy/proxy_util.c
```diff
@@ -4938,14 +4941,43 @@ PROXY_DECLARE(apr_status_t) ap_proxy_tunnel_create(proxy_tunnel_rec **ptunnel,
...
+         * - use socket strategy described above for client only
+         * otherwise the same
+         */
+        tunnel->client->pfd->reqevents = 0;
+        tunnel->origin->pfd->reqevents = APR_POLLIN | APR_POLLHUP |
+                                         APR_POLLOUT | APR_POLLERR;
+        if ((rv = apr_pollset_add(tunnel->pollset, tunnel->origin->pfd))) {
+            return rv;
+        }
+    }
```
ylavic, line 4972: What's the problem when the same/original strategy applies to both sockets and pipes?
It seems to me that a working POLLOUT on the client/pipe side is essential here, or what happens if the backend has a lot of data to send and the h2_c1 connection can't cope?
Doesn't h2 stop reading from the pipe when the h2_c1 would block, so that the pipe fills in and proxy_tunnel_transfer() can then stop reading from the origin socket to calm things down until POLLOUT is back on the pipe?

include/http_core.h
```diff
@@ -1109,6 +1109,28 @@ AP_DECLARE(int) ap_state_query(int query_code);
  */
 AP_CORE_DECLARE(conn_rec *) ap_create_slave_connection(conn_rec *c);
 
+struct apr_pollfd_t;
```
ylavic: #include <apr_poll.h> ?

include/http_core.h
```diff
@@ -1109,6 +1109,28 @@ AP_DECLARE(int) ap_state_query(int query_code);
...
+ * Note that APR_NO_DESC may be set to indicate that the connection
+ * input is already closed.
+ *
+ * @param pfd  the pollfd to set the descriptor in
+ * @param ptimeout  != NULL to retrieve the timeout in effect
+ * @return ARP_SUCCESS when the information was assigned.
+ */
+AP_CORE_DECLARE(apr_status_t) ap_get_conn_in_pollfd(conn_rec *c,
+                                      struct apr_pollfd_t *pfd,
+                                      apr_interval_time_t *ptimeout);
```
ylavic, line 1136: Sorry, name bikeshedding :/
ap_get_conn_in_pollfd() sounds like it's filling a conn_rec from an apr_pollfd_t but it's the other way around IIUC.
Maybe ap_get_pollfd_from_conn() or ap_set_conn_in_pollfd()?

modules/http2/h2_request.c
```diff
@@ -324,36 +329,40 @@ static void assign_headers(request_rec *r, const h2_request *req,
...
+        }
 #endif /* else AP_HAS_RESPONSE_BUCKETS */
+  }
 }
 
 request_rec *h2_create_request_rec(const h2_request *req, conn_rec *c,
                                    int no_body)
 {
     int access_status = HTTP_OK;
+    int is_connect = !apr_strnatcasecmp("CONNECT", req->method);
```
ylavic: Why "natural" comparison is required here and not ap_cstr_casecmp()?

modules/http2/h2_request.c
```diff
@@ -362,24 +371,43 @@ request_rec *h2_create_request_rec(const h2_request *req, conn_rec *c,
...
                                     req->method, req->authority);
     }
-    else if (req->scheme && ap_cstr_casecmp(req->scheme, "http")
-             && ap_cstr_casecmp(req->scheme, "https")) {
-        /* FIXME: we also need to create absolute uris when we are
-         * in a forward proxy configuration! But there is currently
-         * no way to detect that. */
+    else if (req->protocol) {
+      ap_log_cerror(APLOG_MARK, APLOG_DEBUG, 0, c, APLOGNO()
+                    "':protocol: %s' header present in %s request",
```
ylavic, line 401: Stray `'` in `%s'`?

modules/http2/h2_stream.c
```diff
@@ -897,7 +900,14 @@ apr_status_t h2_stream_end_headers(h2_stream *stream, int eos, size_t raw_bytes)
      *      of CONNECT requests (see [RFC7230], Section 5.3)).
      */
     if (!ap_cstr_casecmp(req->method, "CONNECT")) {
-        if (req->scheme || req->path) {
+        if (req->protocol && !strcmp("websocket", req->protocol)) {
+            if (!req->scheme || !req->path) {
+                ap_log_cerror(APLOG_MARK, APLOG_INFO, 0, stream->session->c1,
+                              H2_STRM_LOG(APLOGNO(), stream, "Request to websocket CONNECT "
+                              "without :scheme or :path, sending 400 answer"));
+            }
```
ylavic, line 908: Hm, what is a websocket CONNECT? And why does it require a scheme and path?
Also, it seems it's not really "sending 400 answer" here. (not set_error_response() like in the other error cases below).

modules/http2/h2_ws.c
```diff
@@ -0,0 +1,320 @@
...
+    h2_conn_ctx_t *conn_ctx = h2_conn_ctx_get(c2);
+    h2_request *wsreq;
+    unsigned char key_raw[16];
+    const char *key_base64, *accept_base64;
+    struct ws_filter_ctx *ws_ctx;
+
+    if (!conn_ctx || !req->protocol || strcmp("websocket", req->protocol))
+        return req;
+
+    if (apr_strnatcasecmp("CONNECT", req->method)) {
```
ylavic: Likewise, why apr_strnatcasecmp() and not ap_cstr_casecmp()?

modules/http2/h2_ws.c
```diff
@@ -0,0 +1,320 @@
...
+    if (!req->path) {
+        ap_log_cerror(APLOG_MARK, APLOG_TRACE1, 0, c2,
+                      "h2_c2(%s-%d): websocket CONNECT without :path",
+                      conn_ctx->id, conn_ctx->stream_id);
+        return req;
+    }
+
+    ap_log_cerror(APLOG_MARK, APLOG_TRACE1, 0, c2,
+                  "h2_c2(%s-%d): websocket CONNECT for %s",
+                  conn_ctx->id, conn_ctx->stream_id, req->path);
```
ylavic, line 107: Ah, do you mean websocket upgrade/handshake rather than CONNECT? I find CONNECT rather confusing w.r.t. the HTTP CONNECT method, but maybe it's just me..

modules/http2/h2_ws.c
```diff
@@ -0,0 +1,320 @@
...
+    /* Transform the HTTP/2 extended CONNECT to an internal GET using
+     * the HTTP/1.1 version of websocket connection setup. */
+    wsreq = h2_request_clone(c2->pool, req);
+    wsreq->method = "GET";
+    wsreq->protocol = NULL;
+    apr_table_set(wsreq->headers, "Upgrade", "websocket");
+    /* add Sec-WebSocket-Key header */
+    ap_random_insecure_bytes(key_raw, sizeof(key_raw));
```
ylavic: The system RNG through apr_generate_random_bytes() is probably not slower than ap_random_insecure_bytes() nowadays.

modules/http2/h2_ws.c
```diff
@@ -0,0 +1,320 @@
...
+
+    if (ws_ctx->has_final_response) {
+        /* already did, nop */
+        return;
+    }
+
+    ap_log_cerror(APLOG_MARK, APLOG_TRACE2, 0, c2,
+                  "h2_c2(%s-%d): H2_C2_WS_OUT inspecting response %d",
+                  conn_ctx->id, conn_ctx->stream_id, resp->status);
+    if (resp->status == 101) {
```
ylavic: HTTP_SWITCHING_PROTOCOLS?

modules/http2/h2_ws.c
```diff
@@ -0,0 +1,320 @@
...
+            }
+            b_override = make_invalid_resp(c2, HTTP_BAD_GATEWAY, resp->notes);
+            override_body = is_final = 1;
+        }
+    }
+    else if (resp->status < 200) {
+        /* other intermediate response, pass through */
+    }
+    else if (resp->status < 300) {
+        /* Failure, we might be talking to a plain http resource */
```
ylavic, line 250: I'm not sure we should consider a 2xx response as a failure, can't a backend server respond with anything to a websocket upgrade request?
At least mod_proxy_http simply forwards any non-101 response, without switching of course, maybe we shouldn't turn it to a HTTP_BAD_GATEWAY here.

server/core.c
```diff
@@ -5971,6 +5977,30 @@ static int core_upgrade_storage(request_rec *r)
...
+{
+    if (c && !c->master) {
+        pfd->desc_type = APR_POLL_SOCKET;
+        pfd->desc.s = ap_get_conn_socket(c);
+        if (ptimeout) {
+            apr_socket_timeout_get(pfd->desc.s, ptimeout);
+        }
+        return APR_SUCCESS;
+    }
+    return DECLINED;
```
ylavic: APR_ENOTIMPL?

server/core.c
```diff
@@ -103,6 +104,11 @@ AP_IMPLEMENT_HOOK_RUN_FIRST(apr_status_t, insert_network_bucket,
                              apr_socket_t *socket),
                             (c, bb, socket), AP_DECLINED)
 
+AP_IMPLEMENT_HOOK_RUN_FIRST(apr_status_t, get_conn_in_pollfd,
+                            (conn_rec *c, struct apr_pollfd_t *pfd,
+                             apr_interval_time_t *ptimeout),
+                              (c, pfd, ptimeout), AP_DECLINED)
```
ylavic: APR_ENOTIMPL instead of AP_DECLINED?

server/core.c
```diff
@@ -5971,6 +5977,30 @@ static int core_upgrade_storage(request_rec *r)
...
+    return DECLINED;
+}
+
+AP_CORE_DECLARE(apr_status_t) ap_get_conn_in_pollfd(conn_rec *c,
+                                      struct apr_pollfd_t *pfd,
+                                      apr_interval_time_t *ptimeout)
+{
+    apr_status_t rv;
+    rv = ap_run_get_conn_in_pollfd(c, pfd, ptimeout);
+    return (rv == DECLINED)? APR_ENOTIMPL : rv;
```
ylavic: If get_conn_in_pollfd hooks return APR_ENOTIMPL when they don't implement the thing we could use ap_run_get_conn_in_pollfd() directly possibly.

modules/proxy/proxy_util.c
icing: First, thanks for reviewing this.

POLLOUT on the fd will not work, imo. It is just for reading the signal that new input is available (pipes are in one direction). And you are correct that the strategy is not as good as the one with 2 sockets.

Passing a chunk from origin to client may block in the worker thread, but that does not block the TCP socket to the client in the main connection thread. Data from the client will be accepted, up to the HTTP/2 flow control window, and buffered. There is still a chance of messing up if the client does stop reading output on a stream. But if it really does not want any more output, it should RST the whole stream, I think. On a RST, our blocked output would CONNABORT.

Even if we add a second pipe for output, I have no good idea how POLLOUT signalling on that one would work.

include/http_core.h
icing: I was mentally prepared for that. And I am not emotionally involved in the proposed name. Your propposal works for me.

modules/http2/h2_request.c
icing: Will change.

modules/http2/h2_request.c
icing: its the quote for the header, which is ':protocol: name'.

modules/http2/h2_stream.c
icing: RFC 8441 defines how to bootstrap websockets over HTTP/2. Will check on the error response setting.

modules/http2/h2_ws.c
icing: Will change.

modules/http2/h2_ws.c
icing: It is a HTTP/2 request with ':method: CONNECT'. That is the way it is defined in the standard.

modules/http2/h2_ws.c
icing: I was scanning for what we already use in other parts of the server. Can switch the call if that one is "better".

modules/http2/h2_ws.c
icing: Will change.

modules/http2/h2_ws.c
icing: Again, RFC 8441 defines 200 as the one positive answer to a CONNECT with ':protocol: websocket' and anything else must indicate a failure (could be 401 for auth negotiation).

mod_http2 gets a CONNECT with a path for a websocket resource, it needs to reply 200 *only if* that protocol has been negotiated. It does not know, before inspecting the answer, if the internal resource at path does websocket or is "just" HTTP.

server/core.c
icing: Can do, yes. I find DECLINED better, since hooks commonly use that for calls where they do not engage. But that would work as well.

modules/proxy/proxy_util.c
ylavic: Yes, I think we need two pipes, c1[out]->c2[in] for inbound and c2[out]->c1[in] for outbound (in/out is h1 point of vue, you may see it differently :) ).
Or maybe better a socketpair() which is bidirectional on both sides, available on unix/mac platforms and which we could emulate on Windows (something like the pollset wakeup_socket in APR which is using a pair of AF_INET sockets on 127.0.0.1, or we could use a pair of AF_UNIX sockets which seem to be implemented since Windows 10, e.g. https://gitlab.com/openconnect/openconnect/-/merge_requests/320).

If both h2 and h1 have their own socket to communicate with each other I think we can go full async/nonblocking on both sides (the socket is the buffering/limiting/pollable mechanism between the two), and both could even defer to the MPM for polling.
We could use the current ap_bucket_type_{request,response,headers} to exchange the HTTP header/trailers between h2 and h1 still (it's already parsed/bufferized/full so it would be unfortunate to use the socket/pipes for that), but the request/response bodies could be chunked and received/sent normally by the core/network filters on the h1 side, while h2 (mplx?) could have a pollset of all the h1 sockets/pipes to POLLIN/POLLOUT from/to or even defer to the MPM and be called back (ap_mpm_register_poll_callback_timeout).

Faster to talk about than implement though.. Does it sound like a (realistic) plan still?
Maybe you saw #294 (wip) already, there the h1 header fetching part is fully async already and on the mod_proxy_http side we are not very far from going async/nonblocking for everything (like we do for websocket in trunk already). I'm (slowly) moving on this..

icing: You are talking about a decent chunk of work. And I love the idea of making everything pollable and shifting it to mpm_event. However, that seems a long term thing.

For this PR, I am looking at adding H2 websockets to 2.4.x later and need to limit the scope just a tiny bit. :)

ylavic: OK, let's do that for now then :)
So this hunk looks good, the proxy_tunnel_transfer() one below wants an ap_assert() IMHO (see there).

modules/proxy/proxy_util.c
```diff
@@ -5053,8 +5085,22 @@ static int proxy_tunnel_transfer(proxy_tunnel_rec *tunnel,
...
+        }
+        else {
+            /* Unable to use POLLOUT in this direction, out is not a socket */
+            if (in->down_in) {
+                del_pollset(tunnel->pollset, in->pfd, APR_POLLIN);
+                ap_log_rerror(APLOG_MARK, APLOG_TRACE3, 0, tunnel->r,
+                              "proxy: %s: %s write shutdown",
+                              tunnel->scheme, out->name);
+                out->down_out = 1;
+            }
```
ylavic: Here I think we should remove POLLIN for in->pfd unconditionally and ap_assert(APR_STATUS_IS_EOF(rv)) for the APR_POLL_FILE case.
IIUC, there is no reason that APR_STATUS_IS_INCOMPLETE(rv) is returned above because h2_c2_filter_out() will block to flush the data if necessary thus it will never grow any buffer indefinitely nor set pending data aside to trigger ap_filter_should_yield().
If so we are good here too, in 2.4.x we are not nonblocking anyway for now.

I would just make the assumptions clear, writing this part like:
```
        del_pollset(tunnel->pollset, in->pfd, APR_POLLIN);
        if (out->pfd->desc_type == APR_POLL_SOCKET) {
            /* If the output is a SOCKET, we stop polling the input
             * until the output signals POLLOUT again.
             */
            add_pollset(tunnel->pollset, out->pfd, APR_POLLOUT);
        }
        else {
            /* We can't use POLLOUT in this direction for the only APR_POLL_FILE case
             * we have so far (mod_h2's "signal" pipe), we assume that the client's
             * ouput filters chain will block/flush if necessary (i.e. no pending data),
             * hence that the origin is EOF when reaching here. This direction is over.
             */
            ap_assert(in->down_in && APR_STATUS_IS_EOF(rv));
            ap_log_rerror(APLOG_MARK, APLOG_TRACE3, 0, tunnel->r,
                          "proxy: %s: %s write shutdown (pipe)",
                          tunnel->scheme, out->name);
            out->down_out = 1;
        }
```

modules/proxy/proxy_util.c
icing: Made the changes as suggested.

include/http_core.h
icing: Done

include/http_core.h
icing: Changed to ap_get_pollfd_from_conn()

modules/http2/h2_request.c
icing: Done

modules/http2/h2_ws.c
icing: Done

modules/http2/h2_ws.c
icing: Changed to apr_generate_random_bytes()

modules/http2/h2_ws.c
icing: Done

server/core.c
icing: Changed to APR_ENOTIMPL.

server/core.c
icing: Changed to APR_ENOTIMPL.

server/core.c
icing: Changed to APR_ENOTIMPL.

modules/proxy/proxy_util.c
icing: Changed to that suggestion. Thanks.

ylavic (approved):
+1

icing:
Added to trunk as r1910507.

PR #363, opened 2023-06-18
Update README

Mayank-2545:


rbowen:
This looks reasonable to me. This patch changes the README from plain text to markdown, which will make it look nicer on Github, but makes no other functional or textual difference. I'm +1 on this, but would appreciate other eyes on this.

covener:
maybe personal preference, but I think the long lines are a step backwards. They should be wrapped.

rbowen:
FWIW, I've updated my local version of the patch accordingly, but would still give credit to @Mayank-2545 when merged.

PR #364, opened 2023-06-20
H2 backports 2.4.58

icing:
Merge 1907696,1907697 from trunk:

      *) mod_http2: new directive 'H2MaxDataFrameLen n' to limit the maximum
         amount of response body bytes put into a single HTTP/2 DATA frame.
         Setting this to 0 places no limit (but the max size allowed by the
         protocol is observed).
         The module, by default, tries to use the maximum size possible, which is
         somewhat around 16KB. This sets the maximum. When less response data is
         available, smaller frames will be sent.

    Merge r1909769 from trunk (plus sync mod_http2.xml)

      *) mod_http2: v2.0.15 with the following fixes and improvements
         - New directive 'H2EarlyHint name value' to add headers to a response,
           picked up already when a "103 Early Hints" response is sent. 'name' and
           'value' must comply to the HTTP field restrictions.
           This directive can be repeated several times and header fields of the
           same names add. Sending a 'Link' header with 'preload' relation will
           also cause a HTTP/2 PUSH if enabled and supported by the client.
         - Fixed an issue where requests were not logged and accounted in a timely
           fashion when the connection returns to "keepalive" handling, e.g. when
           the request served was the last outstanding one.
           This led to late appearance in access logs with wrong duration times
           reported.
         - Accurately report the bytes sent for a request in the '%O' Log format.
           This addresses #203, a long outstanding issue where mod_h2 has reported
           numbers over-eagerly from internal buffering and not what has actually
           been placed on the connection.
           The numbers are now the same with and without H2CopyFiles enabled.

    Merge r1910157 from trunk

     * mod_proxy_http2: fixed using the wrong "bucket_alloc" from the backend
       connection when sending data on the frontend one. This caused crashes
       or infinite loops in rare situations.
     * mod_proxy_http2: fixed a bug in retry/response handling that could lead
       to wrong status codes or HTTP messages send at the end of response bodies
       exceeding the announced content-length.
     * mod_proxy_http2: fix retry handling to not leak temporary errors.
       On detecting that that an existing connection was shutdown by the other
       side, a 503 response leaked even though the request was retried on a
       fresh connection.
     * mod_http2: fixed a bug that did cleanup of consumed and pending buckets in
       the wrong order when a bucket_beam was destroyed.

    Merge of 1910331,1910386,1910441 from trunk

      *) mod_http2: fixed a bug in flushing pending data on an already closed
         connection that could lead to a busy loop, preventing the HTTP/2 session
         to close down successfully. Fixed PR 66624.
         Fixed a bug that could lead to a crash in main connection
         output handling. This occured only when the last request on a HTTP/2
         connection had been processed and the session decided to shut down.
         This could lead to an attempt to send a final GOAWAY while the previous
         write was still in progress. See PR 66646.

### DIFF
--- a/changes-entries/h2_early_hints_and_fixes.txt
+++ b/changes-entries/h2_early_hints_and_fixes.txt
@@ -1,0 +1,18 @@
+  *) mod_http2: v2.0.15 with the following fixes and improvements
+     - New directive 'H2EarlyHint name value' to add headers to a response,
+       picked up already when a "103 Early Hints" response is sent. 'name' and
+       'value' must comply to the HTTP field restrictions.
+       This directive can be repeated several times and header fields of the
+       same names add. Sending a 'Link' header with 'preload' relation will
+       also cause a HTTP/2 PUSH if enabled and supported by the client.
+     - Fixed an issue where requests were not logged and accounted in a timely
+       fashion when the connection returns to "keepalive" handling, e.g. when
+       the request served was the last outstanding one.
+       This led to late appearance in access logs with wrong duration times
+       reported.
+     - Accurately report the bytes sent for a request in the '%O' Log format.
+       This addresses #203, a long outstanding issue where mod_h2 has reported
+       numbers over-eagerly from internal buffering and not what has actually
+       been placed on the connection.
+       The numbers are now the same with and without H2CopyFiles enabled.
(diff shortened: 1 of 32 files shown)


icing:
Merged into 2.4.x as r1910699.

PR #366, opened 2023-06-29
h2 websockets backport for 2.4.x

icing:
Based on #364 

* `core`: add `ap_get_pollfd_from_conn()`, increment `mmn` number
* mod_http2:
  - add support for bootstrapping WebSockets via HTTP/2, as  described 
     in RFC 8441. 
  - new directive 'H2WebSockets on|off' has been  added. The feature is by default not enabled.
         As also discussed in the manual, this feature should work for setups
         using "ProxyPass backend-url upgrade=websocket" without further changes.
         Special server modules for WebSockets will have to be adapted,
         most likely, as the handling if IO events is different with HTTP/2.
  - HTTP/2 WebSockets are supported on platforms with native pipes. This
         excludes Windows.
  - small optimisation for output passing when sending DATA frames
  - new directive `H2ProxyRequests on|off` to enable handling
         of HTTP/2 requests in a forward proxy configuration.
         General forward proxying is enabled via `ProxyRequests`. If the
         HTTP/2 protocol is also enabled for such a server/host, this new
         directive is needed in addition.

* tests: add websocket tests, requires python `websockets` module version 10.4 or newer


modules/proxy/proxy_util.c
```diff
@@ -4765,17 +4773,12 @@ PROXY_DECLARE(apr_status_t) ap_proxy_tunnel_create(proxy_tunnel_rec **ptunnel,
...
-    apr_socket_timeout_get(tunnel->origin->pfd->desc.s, &tunnel->timeout);
-    if (timeout >= 0 && (tunnel->timeout < 0 || tunnel->timeout < timeout)) {
-        tunnel->timeout = timeout;
-    }
-
-    /* We should be nonblocking from now on the sockets */
-    apr_socket_opt_set(tunnel->client->pfd->desc.s, APR_SO_NONBLOCK, 1);
-    apr_socket_opt_set(tunnel->origin->pfd->desc.s, APR_SO_NONBLOCK, 1);
+    tunnel->timeout = (origin_timeout >= 0 && origin_timeout > client_timeout)?
+                      origin_timeout : client_timeout;
```
ylavic: I'm wondering if we shouldn't use the smaller of the two timeouts instead (i.e. change the original behaviour). It could be easier for the user to control the lifetime/ending of ws connections by doing this..

icing: I was wondering about the original choice. I think it makes sense to use the smaller timeout, since this *should* tear down the tunnel, I believe.

ylavic: Yeah, makes more sense to me.

icing: Like this?

```
    tunnel->timeout = (client_timeout > 0 && client_timeout < origin_timeout)?
                      client_timeout : origin_timeout;
```

icing: The question really is what a timeout of `0` actually should do? Would the tunnel not exit after the first poll without an event?

ylavic: A zero timeout is probably not possible/useful but strictly speaking it would be:
```
    tunnel->timeout = (client_timeout >= 0 && client_timeout < origin_timeout ?
                       client_timeout : origin_timeout);
```
(using _my_ codestyle to please _my_ vim indent :)

ylavic: > The question really is what a timeout of `0` actually should do? Would the tunnel not exit after the first poll without an event?

I think it would timeout immediately if data on any side are not available immediately (not very useful but an admin issue..).

icing: Added as r1910704 to trunk.

ylavic: :+1: thanks Stefan!

include/ap_mmn.h
```diff
@@ -595,15 +596,15 @@
  * 20120211.125 (2.4.55-dev) Export mod_http2.h as public header
  * 20120211.126 (2.4.55-dev) Add additional hcmethod_t enums and PROXY_WORKER_IS_ERROR
  * 20120211.127 (2.4.56-dev) Add ap_proxy_canonenc_ex
- * 20120211.128 (2.4.55-dev) Add AP_CTIME_OPTION_GMTOFF to util_time.h
+ * 20120211.129 (2.4.58-dev) Add ap_get_pollfd_from_conn()
```
rpluem, line 599: Why removing the 128 minor bump from comments?

modules/http2/h2_c2_filter.c
```diff
@@ -108,20 +109,39 @@ apr_status_t h2_c2_filter_request_in(ap_filter_t *f,
...
+        if (req->http_status != H2_HTTP_STATUS_UNSET) {
             /* error was encountered preparing this request */
-            b = ap_bucket_error_create(conn_ctx->request->http_status, NULL, f->r->pool,
+            ap_log_cerror(APLOG_MARK, APLOG_TRACE2, 0, f->c,
+                          "h2_c2_filter_request_in(%s): adding error bucket %d",
+                          conn_ctx->id, req->http_status);
+            b = ap_bucket_error_create(req->http_status, NULL, f->r->pool,
                                        f->c->bucket_alloc);
             APR_BRIGADE_INSERT_TAIL(bb, b);
             return APR_SUCCESS;
```
rpluem, line 137: In the old code we only had the error bucket in `bb`in case that `req->http_status != H2_HTTP_STATUS_UNSET`. Now we always have a request bucket in `bb` even if `bb`in case that `req->http_status != H2_HTTP_STATUS_UNSET`. Does this matter?

modules/http2/h2_c2_filter.c
```diff
@@ -184,7 +204,7 @@ static int uniq_field_values(void *d, const char *key, const char *val)
          */
         for (i = 0, strpp = (char **) values->elts; i < values->nelts;
              ++i, ++strpp) {
-            if (*strpp && apr_strnatcasecmp(*strpp, start) == 0) {
+            if (*strpp && ap_cstr_casecmp(*strpp, start) == 0) {
```
rpluem: This gives different results for strings e.g. that consist only of spaces, but off different length, or that start with a different amount of spaces as prefix and are otherwise the same.

modules/http2/h2_c2_filter.c
```diff
@@ -292,7 +312,7 @@ static h2_headers *create_response(request_rec *r)
 
         while (field && (token = ap_get_list_item(r->pool, &field)) != NULL) {
             for (i = 0; i < r->content_languages->nelts; ++i) {
-                if (!apr_strnatcasecmp(token, languages[i]))
+                if (!ap_cstr_casecmp(token, languages[i]))
```
rpluem: Same as above

modules/http2/h2_proxy_util.c
```diff
@@ -381,7 +381,7 @@ static int iq_bubble_down(h2_proxy_iqueue *q, int i, int bottom,
  * h2_proxy_ngheader
  ******************************************************************************/
 #define H2_HD_MATCH_LIT_CS(l, name)  \
-    ((strlen(name) == sizeof(l) - 1) && !apr_strnatcasecmp(l, name))
+    ((strlen(name) == sizeof(l) - 1) && !ap_cstr_casecmp(l, name))
```
rpluem: Same as above on this topic.

modules/http2/h2_proxy_util.c
```diff
@@ -500,7 +500,7 @@ static int ignore_header(const literal *lits, size_t llen,
     
     for (i = 0; i < llen; ++i) {
         lit = &lits[i];
-        if (lit->len == nlen && !apr_strnatcasecmp(lit->name, name)) {
+        if (lit->len == nlen && !ap_cstr_casecmp(lit->name, name)) {
```
rpluem: Same as above on this topic.

modules/http2/h2_proxy_util.c
```diff
@@ -542,7 +542,7 @@ void h2_proxy_util_camel_case_header(char *s, size_t len)
 
 /** Match a header value against a string constance, case insensitive */
 #define H2_HD_MATCH_LIT(l, name, nlen)  \
-    ((nlen == sizeof(l) - 1) && !apr_strnatcasecmp(l, name))
+    ((nlen == sizeof(l) - 1) && !ap_cstr_casecmp(l, name))
```
rpluem: Same as above on this topic.

modules/http2/h2_push.c
```diff
@@ -426,7 +426,7 @@ static void inspect_link(link_ctx *ctx, const char *s, size_t slen)
 
 static int head_iter(void *ctx, const char *key, const char *value) 
 {
-    if (!apr_strnatcasecmp("link", key)) {
+    if (!ap_cstr_casecmp("link", key)) {
         inspect_link(ctx, value, strlen(value));
```
rpluem, line 430: Same as above on this topic.

modules/proxy/proxy_util.c
```diff
@@ -4799,14 +4802,43 @@ PROXY_DECLARE(apr_status_t) ap_proxy_tunnel_create(proxy_tunnel_rec **ptunnel,
...
+    else if (tunnel->client->pfd->desc_type == APR_POLL_FILE) {
+        /* Input is a PIPE fd, the poll strategy is:
+         * - always POLLIN on origin
+         * - use socket strategy described above for client only
+         * otherwise the same
+         */
+        tunnel->client->pfd->reqevents = 0;
+        tunnel->origin->pfd->reqevents = APR_POLLIN | APR_POLLHUP |
+                                         APR_POLLOUT | APR_POLLERR;
+        if ((rv = apr_pollset_add(tunnel->pollset, tunnel->origin->pfd))) {
```
rpluem, line 4830: Don't we add the client fd aka the pipe to the pollset?

modules/http2/h2_ws.h
```diff
@@ -0,0 +1,35 @@
...
+#ifndef __mod_h2__h2_ws__
+#define __mod_h2__h2_ws__
+
+#include "h2.h"
+
+/**
+ * Rewrite a websocket request.
+ *
+ * @param req the h2 request to rewrite
+ * @param conn the connection to process the request on
```
rpluem: Nitpick: Either `c2` here in the docs or `conn` below.

modules/http2/h2_util.c
```diff
@@ -1706,7 +1706,7 @@ static apr_status_t req_add_header(apr_table_t *headers, apr_pool_t *pool,
         return APR_SUCCESS;
     }
     else if (nv->namelen == sizeof("cookie")-1
-             && !apr_strnatcasecmp("cookie", (const char *)nv->name)) {
+             && !ap_cstr_casecmp("cookie", (const char *)nv->name)) {
```
rpluem, line 1709: Same comment as above

modules/http2/h2_util.c
```diff
@@ -1725,7 +1725,7 @@ static apr_status_t req_add_header(apr_table_t *headers, apr_pool_t *pool,
         }
     }
     else if (nv->namelen == sizeof("host")-1
-             && !apr_strnatcasecmp("host", (const char *)nv->name)) {
+             && !ap_cstr_casecmp("host", (const char *)nv->name)) {
```
rpluem, line 1728: Same comment as above

modules/http2/h2_util.c
```diff
@@ -1650,7 +1650,7 @@ static int contains_name(const literal *lits, size_t llen, nghttp2_nv *nv)
     for (i = 0; i < llen; ++i) {
         lit = &lits[i];
         if (lit->len == nv->namelen
-            && !apr_strnatcasecmp(lit->name, (const char *)nv->name)) {
+            && !ap_cstr_casecmp(lit->name, (const char *)nv->name)) {
             return 1;
```
rpluem, line 1654: Same comment as above

modules/http2/h2_util.h
```diff
@@ -337,7 +337,7 @@ apr_size_t h2_util_table_bytes(apr_table_t *t, apr_size_t pair_extra);
 
 /** Match a header value against a string constance, case insensitive */
 #define H2_HD_MATCH_LIT(l, name, nlen)  \
-    ((nlen == sizeof(l) - 1) && !apr_strnatcasecmp(l, name))
+    ((nlen == sizeof(l) - 1) && !ap_cstr_casecmp(l, name))
```
rpluem: Same comment as above

modules/http2/h2_ws.c
```diff
@@ -0,0 +1,361 @@
...
+                              "h2_c2(%s-%d): websocket CONNECT, got 101 response "
+                              "without Sec-WebSocket-Accept header",
+                              conn_ctx->id, conn_ctx->stream_id);
+            }
+            else {
+                /* This points to a bug, either in our WebSockets negotiation
+                 * or in the request processings implementation of WebSockets */
+                ap_log_cerror(APLOG_MARK, APLOG_ERR, 0, c2, APLOGNO(10463)
+                              "h2_c2(%s-%d): websocket CONNECT, 101 response "
+                              "without 'Sec-WebSocket-Accept: %s' but expected %s",
```
rpluem:
```suggestion
                              "with 'Sec-WebSocket-Accept: %s' but expected %s",
```

rpluem (changes requested):

include/ap_mmn.h
icing: Ah, my mistake. 128 was merged when the PR already existed and I messed this up during the rebase.

include/ap_mmn.h
```diff
@@ -412,6 +412,7 @@
  *                         ap_condition_if_none_match(),
  *                         ap_condition_if_modified_since(),
  *                         ap_condition_if_range()
+<<<<<<< HEAD
```
covener: leftover line from merge conflict

modules/http2/h2_c2_filter.c
icing: Oh, then I misunderstood Yann's comment that `ap_cstr_casecmp` is the preferred use.

I will revert this instance in header merging. There *should* be no leading spaces, but...

modules/http2/h2_c2_filter.c
icing: Will revert.

modules/http2/h2_proxy_util.c
icing: Wil revert.

modules/http2/h2_proxy_util.c
icing: Will revert.

modules/http2/h2_proxy_util.c
icing: Will revert.

modules/http2/h2_push.c
icing: here it should be the better choice.

modules/proxy/proxy_util.c
icing: No, we cannot POLLOUT the pipe fd. It is ony way, only for input pending notifications.

modules/http2/h2_ws.h
icing: Will fix.

modules/http2/h2_util.c
icing: here it is the better choice. We checked the length already.

modules/http2/h2_util.c
icing: Better choice, length already checked.

modules/http2/h2_util.c
icing: Better choice. `nghttp2_nv` will not take field names with spaces.

modules/http2/h2_util.h
icing: Will revert for better-safe-than-sorry reasons since this macro is used in several places.

modules/http2/h2_ws.c
icing: Will fix.

modules/http2/h2_c2_filter.c
icing: This is the case with `AP_HAS_RESPONSE_BUCKETS`, so `trunk`. 

In trunk, we need to add a request bucket to the input, so the context for processing the error bucket can be established, e.g. if the error is a 400 response, what this response is for.

This case did not trigger in trunk before websockets and their tests. Now in `test_h2_800_01`, we send a `CONNECT` with an unknown `procotol:` header. That will be answered with a 400.

error buckets without an established request lead to a crash in trunk. I think we can discuss there how to improve the situation and if that could possibly apply to HTTP/1.1 processing as well.

icing:
Many thanks for reviewing and finding mistakes. Just pushed the changes.

include/ap_mmn.h
icing: Thanks, will push fix.

modules/http2/h2_c2_filter.c
rpluem: Thanks for explaining

modules/http2/h2_config.c
```diff
@@ -363,6 +375,11 @@ static void h2_srv_config_seti(h2_config *conf, h2_config_var_t var, int val)
         case H2_CONF_MAX_DATA_FRAME_LEN:
             H2_CONFIG_SET(conf, max_data_frame_len, val);
             break;
+        case H2_CONF_PROXY_REQUESTS:
+            H2_CONFIG_SET(conf, proxy_requests, val);
+        case H2_CONF_WEBSOCKETS:
```
tititiou36, line 381: Blind guess: break missing?

modules/http2/h2_c2.c
```diff
@@ -668,11 +718,21 @@ static apr_status_t c2_process(h2_conn_ctx_t *conn_ctx, conn_rec *c)
 {
     const h2_request *req = conn_ctx->request;
     conn_state_t *cs = c->cs;
-    request_rec *r;
+    request_rec *r = NULL;
     const char *tenc;
     apr_time_t timeout;
+    apr_status_t rv = APR_SUCCESS;
+
+    if(req->protocol && !strcmp("websocket", req->protocol)) {
```
tititiou36: Nit: Space after if

modules/http2/h2_config.c
icing: Nice catch. Will fix this silly mistake.

modules/http2/h2_c2.c
icing: Thanks. Switch between projects with different coding styles poses challenges. ;)

modules/http2/h2_config.c
icing: Pushed.

modules/http2/h2_c2.c
icing: Pushed.

rpluem-vf:
As this PR touches the HTTP/2 implementation I think it should pass the HTTP/2 test suite.

rpluem:
As this PR touches the HTTP/2 implementation I think it should pass the HTTP/2 test suite.

icing:
> As this PR touches the HTTP/2 implementation I think it should pass the HTTP/2 test suite.

The HTTP/2 test suite should be more reliable, even when CPU resources are scarce. Pushed a fix.

icing:
github runner seem to have a Monday. Fixed the failure in test_h2_600_32 in trunk r1911964.

icing:
Merged to 2.4.x in [r1911979](http://svn.apache.org/viewvc?view=revision&revision=1911979).

PR #367, opened 2023-07-05
Proxy address reuse and expiry

ylavic:


modules/proxy/mod_proxy.h
```diff
@@ -755,12 +765,13 @@ typedef __declspec(dllimport) const char *
 /* Connection pool API */
 /**
  * Return the user-land, UDS aware worker name
- * @param p        memory pool used for displaying worker name
+ * @param unused   memory pool unused
  * @param worker   the worker
  * @return         name
+ * @note Even though the returned name is non constant char*, the string
+ *       it points to is shared and should *not* be modified by the caller!
  */
```
rpluem, line 774: How are the above changes related?

modules/proxy/mod_proxy_ftp.c
```diff
@@ -1176,23 +1126,34 @@ static int proxy_ftp_handler(request_rec *r, proxy_worker *worker,
...
     }
 
+    /*
+     * get all the possible IP addresses for the destname and loop through
+     * them until we get a successful connection
+     */
+    err = ap_proxy_determine_address("FTP", backend, connectname, connectport,
+                                     r, r->server);
+    if (APR_SUCCESS != err) {
+        proxy_ftp_cleanup(r, backend);
```
rpluem: Why only `proxy_ftp_cleanup` and not `ftp_proxyerror` with  a  `DNS lookup failure for:` error message?

modules/proxy/mod_proxy.c
```diff
@@ -227,6 +227,18 @@ static const char *set_worker_param(apr_pool_t *p,
...
+         */
+        char *endptr;
+        long ttl = strtol(val, &endptr, 10);
+        if (endptr == val || *endptr || ttl < -1L || ttl > APR_INT32_MAX) {
+            return "AddressTTL must be a number between -1 and "
+                    APR_STRINGIFY(APR_INT32_MAX);
+        }
+        worker->s->address_ttl = apr_time_from_sec(ttl);
+        worker->s->address_ttl_set = 1;
+    }
```
rpluem, line 247: Why not using `ap_timeout_parameter_parse`?

modules/proxy/mod_proxy_balancer.c
```diff
@@ -142,20 +142,18 @@ static int proxy_balancer_canon(request_rec *r, char *url)
     return OK;
 }
 
-static void init_balancer_members(apr_pool_t *p, server_rec *s,
-                                 proxy_balancer *balancer)
+static void init_balancer_members(proxy_balancer *balancer,
+                                  server_rec *s, apr_pool_t *p)
```
rpluem, line 146: Why changing the order of parameters?

modules/proxy/proxy_util.c
```diff
@@ -1492,92 +1515,168 @@ static void socket_cleanup(proxy_conn_rec *conn)
...
+             || worker->s->disablereuse) {
         socket_cleanup(conn);
         conn->close = 0;
     }
     else if (conn->is_ssl) {
-        /* Unbind/reset the SSL connection dir config (sslconn->dc) from
-         * r->per_dir_config, r will likely get destroyed before this proxy
-         * conn is reused.
+        /* The current ssl section/dir config of the conn is not necessarily
+         * the one it will be reused for, so while the conn is in the reslit
```
rpluem:
```suggestion
         * the one it will be reused for, so while the conn is in the reslist
```

modules/proxy/proxy_util.c
```diff
@@ -1684,14 +1765,11 @@ static apr_status_t connection_destructor(void *resource, void *params,
...
                                            proxy_worker *worker)
 {
-    if (!(*worker->s->uds_path) || !p) {
-        /* just in case */
-        return worker->s->name;
-    }
-    return apr_pstrcat(p, "unix:", worker->s->uds_path, "|", worker->s->name, NULL);
+    (void)unused;
+    return worker->uds_name ? (char *)worker->uds_name : worker->s->name;
 }
```
rpluem, line 1779: Why the above change?

modules/proxy/proxy_util.c
```diff
@@ -3544,12 +3823,10 @@ static int proxy_connection_create(const char *proxy_function,
          * the peer reset the connection already; ap_create_connection()
          * closed the socket
          */
-        ap_log_error(APLOG_MARK, APLOG_DEBUG, 0,
+        ap_log_error(APLOG_MARK, APLOG_ERR, 0,
                      s, APLOGNO(00960) "%s: an error occurred creating a "
                      "new connection to %pI (%s)", proxy_function,
                      backend_addr, conn->hostname);
-        /* XXX: Will be closed when proxy_conn is closed */
-        socket_cleanup(conn);
```
rpluem, line 3552: Why isn't this needed any longer?

modules/proxy/proxy_util.c
```diff
@@ -2620,6 +2697,248 @@ PROXY_DECLARE(int) ap_proxy_release_connection(const char *proxy_function,
...
+                apr_pool_t *pool = NULL;
+                apr_sockaddr_t *addr;
+
+                apr_pool_create(&pool, worker->cp->dns_pool);
+                rv = apr_sockaddr_info_get(&addr, hostname, APR_UNSPEC,
+                                           hostport, 0, pool);
+                if (rv != APR_SUCCESS) {
+                    PROXY_THREAD_UNLOCK(worker);
+                    apr_pool_destroy(pool);
+                    if (r && !s) {
```
rpluem: Is this the healthcheck worker case?

modules/proxy/proxy_util.c
```diff
@@ -2620,6 +2697,248 @@ PROXY_DECLARE(int) ap_proxy_release_connection(const char *proxy_function,
...
+        }
+
+        address = AP_VOLATILIZE_T(proxy_address *, worker->address);
+        if (!address
+            || conn->address != address
+            || apr_atomic_read32(&address->expiry) <= now) {
+            PROXY_THREAD_LOCK(worker);
+
+            /* Re-check under the mutex, things may have changed. */
+            address = worker->address;
```
rpluem: Why don't we need to use `AP_VOLATILIZE_T` here?

modules/proxy/proxy_util.c
```diff
@@ -2620,6 +2697,248 @@ PROXY_DECLARE(int) ap_proxy_release_connection(const char *proxy_function,
...
+                if (old_address) {
+                    proxy_address_dec(old_address);
+                }
+
+                address_changed = 1;
+            }
+
+            /* Update worker->cp->addr if it's not set (or reset) */
+            if (!AP_VOLATILIZE_T(apr_sockaddr_t *, worker->cp->addr)) {
+                worker->cp->addr = address->addr;
```
rpluem: How do we ensure it never expires even if it got reset in the meantime?

modules/proxy/mod_proxy.c
ylavic: This must be in seconds <= APR_INT32_MAX, but yes it could benefit from the "h" or "mi" suffixes still (but not "ms"..). I will use the below instead:
```
        /* Address TTL in seconds
         */
        apr_interval_time_t ttl;
        if (strcmp(val, "-1") == 0) {
            worker->s->address_ttl = -1;
        }
        else if (ap_timeout_parameter_parse(val, &ttl, "s") == APR_SUCCESS
                 && (ttl <= apr_time_from_sec(APR_INT32_MAX))
                 && (ttl % apr_time_from_sec(1)) == 0) {
            worker->s->address_ttl = ttl;
        }
        else {
            return "AddressTTL must be -1 or a number of seconds not "
                   "exceeding " APR_STRINGIFY(APR_INT32_MAX);
        }
        worker->s->address_ttl_set = 1;
```

modules/proxy/mod_proxy.h
ylavic: They are not :) As I said on dev@, I just happened to change them "by the way".
Only the first commit of this PR is really related actually.

modules/proxy/mod_proxy_balancer.c
ylavic: Yeah, I changed this interface several times, passing the `proxy_server_conf*` etc to finally coming back to the initial parameters. Like this it aligns with ap_proxy_initialize_balancer/worker(), so I went with it..

modules/proxy/mod_proxy_ftp.c
ylavic: Argh, yes, either that or use NULL instead of r->server in the ap_proxy_determine_address() call to get an APLOG_ERR from there.
I'll go with the latter if my explanation below on why `(r && !s)` is not too twisted :)

modules/proxy/proxy_util.c
ylavic: Thanks, will fix.

modules/proxy/proxy_util.c
ylavic: Well, it's not related, but not very wise either to race or leak on p here when we know the uds_name from the start.

I will not squash all the commits of this PR in the same svn commit (should this PR be fine) if that reassures you :)

modules/proxy/proxy_util.c
ylavic: It'd probably be consistent/clearer to do so, though worker->address can only change under the mutex and I don't think that the compiler is allowed to reorder this instruction much here since it's preceded by a barrier (mutex lock) and followed by a control/address dependency (the `&address->expiry`).

modules/proxy/proxy_util.c
ylavic: It's for when the user wants an ERR log (and some error details in r->notes) generated by ap_proxy_dermine_address() directly. In this PR, it applies to the calls in ap_proxy_determine_connection() and ftp (well I missed that one actually as you noticed above), while the calls in hcheck and ap_proxy_connect_backend() have no `r` so use `s` only.
So `(r && !s)` calls ap_proxyerror() (thus APLOG_ERR and notes), while `r` or `s` alone issue an APLOG_DEBUG only. Yeah I know it's not a pretty interface, but I didn't want/feel to use an explicit parameter for this, and it's in the dox! :p

modules/proxy/proxy_util.c
ylavic: The first address/DNS lookup will never expire because it has a refcount of 2, but if some (third party) module plays by resetting worker->cp->addr arbitrarily after the startup it will be set to the current address, which will expire when a new one is needed and the worker or last connection using it releases it.
I don't think it's ever been safe to reset worker->cp->addr, so we should be good?

modules/proxy/proxy_util.c
ylavic: Every caller of `ap_proxy_connection_create[_ex]()` should release/invalidate the connection on failure already, like with any other ap_proxy_ function taking a proxy_conn_rec which fails? We do this upstream already afaict.

modules/proxy/mod_proxy.c
ylavic: I should add that with this PR, address_ttl=0 disables address and connection reuse completely for the worker, while an explicit address_ttl=-1|>0 (where -1 is infinite) allows for address reuse regardless of enable/disablereuse (which then relates to connection/socket reuse only).
IOW, this new parameter allows to distinguish betwwen enable/disablereuse and is_address_reusable, still when is_address_reusable is zero (either explicitely or implicitely for the generic workers) both the addresses and connections are never reused, same when enablereuse=0 and address_ttl= is not set, but an explicit address_ttl= to something not zero will reuse addresses and in this case enable/disablereuse (default or set) will apply to connections/sockets only.
I thought it was something desirable..

modules/proxy/proxy_util.c
rpluem: Good point, but in order to be save shouldn't we do at least a `conn->close = 1` here in case the caller only releases the connection but forgets to update the `close` flag before? But I agree that the current consumers in our code do it properly.

modules/proxy/proxy_util.c
rpluem: My worry is that the compiler stored worker->address in a register when doing the assignment in line 2791 and reuse this value from the register after the lock. But this worry might be unfounded.

modules/proxy/proxy_util.c
rpluem: I agree that using setting `worker->cp->addr` to `NULL` as an indicator was not a good idea and given that this possibility is probably only available in one patch release of 2.4. I am personally happy to drop this. I only feel unsure if we could do it for formal reasons. But probably as long as no one complains we get away with it :-p.

modules/proxy/proxy_util.c
```diff
@@ -403,8 +419,12 @@ PROXY_DECLARE(char *)
     return NULL;
 }
 
-PROXY_DECLARE(int) ap_proxyerror(request_rec *r, int statuscode, const char *message)
+static int ap_proxyerror_ex(request_rec *r, int statuscode, const char *message,
```
rpluem: Is this the correct name for a `static` function? I think it should be either `proxyerror_core` (at least dropping the `ap_` prefix, the further part of the name I don't care that much) or we should make it part of the API and properly declare it via `PROXY_DECLARE` as non static function and add it to `mod_proxy.h`.

modules/proxy/proxy_util.c
ylavic: I don't think that the compiler is allowed to do that because it'd move an instruction from inside to outside the critical section, which is quite against the principle of critical sections. The compiler can move instructions from outside to inside, but not here either for `address` because it's used in the condition to access the CS (data and address dependency) and it's volatilized anyway to keep the instruction just before the condition.
But to be clear/consistent I'll change to use volatile access inside the CS too.

modules/proxy/proxy_util.c
ylavic: Fixed

modules/proxy/mod_proxy_ftp.c
ylavic: Fixed in [52126bf](https://github.com/apache/httpd/pull/367/commits/52126bff6facd33ca10f03401565a09aa182018a), using ftp_proxyerror() finally.

modules/proxy/mod_proxy.c
ylavic: Fixed in [8ad4da6](https://github.com/apache/httpd/pull/367/commits/8ad4da692850d480bf2fd85518485596872e1b3b)

modules/proxy/proxy_util.c
ylavic: Added [97ce1e9](https://github.com/apache/httpd/pull/367/commits/97ce1e996b29d8666ae6778ebe4e1feff0986650) to be consistent on how we close the socket on reuse failure.

modules/proxy/proxy_util.c
ylavic: Done in [8ad4da6](https://github.com/apache/httpd/pull/367/commits/8ad4da692850d480bf2fd85518485596872e1b3b)

modules/proxy/proxy_util.c
ylavic: Done in [8ad4da6](https://github.com/apache/httpd/pull/367/commits/8ad4da692850d480bf2fd85518485596872e1b3b)

modules/proxy/proxy_util.c
ylavic: I don't see where we do this in 2.4, it's trunk changes only for now and addressed by this PR so we should be good ;)

rpluem: My bad. Missed that it was never part of 2.4.x.

rpluem:
Let's go for it.

rpluem:
@ylavic : Do you care to commit to trunk?

rpluem:
@ylavic : Do you care to commit to trunk?

ylavic:
> Do you care to commit to trunk?

Yeah, sorry about the delay (quite busy times..).
I just pushed some more commits for what I could find (re-)reading the changes a few months later, the plan is to squash the many commits in bigger related ones with better commit messages (I'll do this locally) and then push that to trunk. Hopefully tonight..

ylavic:
The last commit ([176de10](https://github.com/apache/httpd/pull/367/commits/176de10e2bd63b7d3d025010a3f8f7072f487a29)) is to avoid expiring the current backend address forcibly and maybe spuriously if ap_proxy_connect_backend() fails on the existing address(es) for something else than some address(es) change. WDYT?

darmbrust:
Is there an easy way to figure out what version this patch landed in?  I see the docs for it are already advertised, but the docs don't say when it became a valid option.  I don't see it mentioned in the changelog.

Also, if I'm understanding this feature correctly, it will be a better solution to a problem folks have where load balancers rotate IPs:
https://serverfault.com/questions/870026/apache-honouring-the-dns-ttl-in-proxy-pass
The current suggestion is to disablereuse=On

Does this section of the workers doc need to be updated with this feature?
https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#workers

Specifically:

**
DNS resolution for origin domains

DNS resolution happens when the socket to the origin domain is created for the first time. When connection reuse is enabled, each backend domain is resolved only once per child process, and cached for all further connections until the child is recycled. This information should to be considered while planning DNS maintenance tasks involving backend domains. Please also check ProxyPass parameters for more details about connection reuse. 
**

How does the above interact with this new feature?  Will it remove connections when at age address_ttl?  Or, will it take a failure on a connection, and then resolve dns if address_ttl is expired before creating a new connection?

rpluem:
> Is there an easy way to figure out what version this patch landed in? I see the docs for it are already advertised, but the docs don't say when it became a valid option. I don't see it mentioned in the changelog.

It will be part of the next 2.4.x release.

> 
> Also, if I'm understanding this feature correctly, it will be a better solution to a problem folks have where load balancers rotate IPs: https://serverfault.com/questions/870026/apache-honouring-the-dns-ttl-in-proxy-pass The current suggestion is to disablereuse=On
> 
> Does this section of the workers doc need to be updated with this feature? https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#workers
> 
> Specifically:
> 
> ** DNS resolution for origin domains
> 
> DNS resolution happens when the socket to the origin domain is created for the first time. When connection reuse is enabled, each backend domain is resolved only once per child process, and cached for all further connections until the child is recycled. This information should to be considered while planning DNS maintenance tasks involving backend domains. Please also check ProxyPass parameters for more details about connection reuse. **
> 

This would need some update.

> How does the above interact with this new feature? Will it remove connections when at age address_ttl? Or, will it take a failure on a connection, and then resolve dns if address_ttl is expired before creating a new connection?

It will close connections when the TTL of the DNS entry is expired or if the the address changed in the meantime as another connection triggered a new lookup which resulted in a different address and recreate them.

PR #370, opened 2023-07-06
mod_ldap: Use the LDAP API directly to implement the rebind callback

notroj:
```
mod_ldap: Use the LDAP API directly to implement the rebind callback
for modern versions of OpenLDAP, avoiding the overhead of the apr-util implementation.

* modules/ldap/util_ldap.c: Define USE_APR_LDAP_REBIND unless a modern version of OpenLDAP is used. (uldap_rebind_proc): New function. (uldap_rebind_init, uldap_rebind_add): Define, using either the callback or the (bad) APR-util versions. (uldap_connection_unbind): Clear the rebind pool to remove rebind references prior to destroying the LDAP *. Omit for !USE_APR_LDAP_REBIND. (uldap_connection_init): Use new wrappers, only create the rebind pool if USE_APR_LDAP_REBIND.

* include/util_ldap.h: Don't include apr_ldap_rebind.h here.

PR: 64414
```

### DIFF
--- a/changes-entries/ldap-rebind.txt
+++ b/changes-entries/ldap-rebind.txt
@@ -1,0 +1,2 @@
+  *) mod_ldap: Avoid performance overhead of APR-util rebind cache for
+     OpenLDAP 2.2+.  PR 64414.  [Joe Orton]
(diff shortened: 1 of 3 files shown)


notroj:
Merged in 85be5353986d676dd75a424e8d4094128e9a20d9

PR #371, opened 2023-07-07
pytests: Ignore log errors/warnings per function.

ylavic:
It's easier to control errors at the function scope and it avoids ignored errors/warnings to propagate to the following tests (i.e. mask further unexpected errors).


ylavic:
@icing Does this work for you?

icing:
Good idea, excellent!

PR #372, opened 2023-07-15
Add MPEG-TS file extensions

mohd-akram:
[Extensions (per FFmpeg)](https://github.com/FFmpeg/FFmpeg/blob/n6.0/libavformat/mpegtsenc.c#L2345)

The mts extension is predominantly used for AVCHD files recorded by camcorders. I cannot find any evidence of `model/vnd.mts` being used for anything and it seems to be for a defunct file format.

### DIFF
--- a/docs/conf/mime.types
+++ b/docs/conf/mime.types
@@ -1645,7 +1645,7 @@ model/vnd.gdl					gdl
 # model/vnd.gs.gdl
 model/vnd.gtw					gtw
 # model/vnd.moml+xml
-model/vnd.mts					mts
+# model/vnd.mts
 # model/vnd.opengex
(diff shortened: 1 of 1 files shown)


mohd-akram:
Thoughts on this? Is it good to merge?

mohd-akram:
@icing Friendly ping. These `mime.types` are used by many projects so it would be great to have this merged as many currently have an incorrect mime type in the case of `mts` and missing mime types in the case of the other extensions, despite their popularity.

notroj:
Seems reasonable to me, thanks for the PR.

mohd-akram:
@notroj Thank you for merging. It would be great if this could also be cherry-picked for the 2.4.60 release.

PR #373, opened 2023-07-15
mod_log_config: Add log_formatter support and a JSON log_formatter

thomasmey:
Based on https://github.com/apache/httpd/pull/353 but based on trunk

### DIFF
--- a/modules/loggers/mod_log_config.c
+++ b/modules/loggers/mod_log_config.c
@@ -167,6 +167,10 @@
 #include "ap_mpm.h"
 #include "ap_provider.h"
 
+#if APU_MAJOR_VERSION > 1 || (APU_MAJOR_VERSION == 1 && APU_MINOR_VERSION >= 7)
+#include "apr_json.h"
+#endif
+
 #if APR_HAVE_UNISTD_H
 #include <unistd.h>
 #endif
@@ -1761,15 +1765,152 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
     return old;
 }
 
+#if APU_MAJOR_VERSION > 1 || (APU_MAJOR_VERSION == 1 && APU_MINOR_VERSION >= 7)
+static ap_log_formatted_data * ap_json_log_formatter( request_rec *r,
+                           void *handle,
+                           ap_log_formatted_data *lfd,
+                           const void *itms)
+
+{
+    json_log_formatter_options * formatter_options = (json_log_formatter_options *) handle;
+    log_format_item *items = (log_format_item *) itms;
+
+    const char* attribute_name;
+    const char* attribute_value;
+
+    ap_log_formatted_data *lfdj = apr_palloc(r->pool, sizeof(ap_log_formatted_data));
+
+    /* TODO: how to determine max number of vectors? */
+    int nvec = lfd->nelts * 3;
+    struct iovec *vec = apr_palloc(r->pool, sizeof(struct iovec*) * nvec );
+
+    apr_bucket_alloc_t *ba;
+    apr_bucket_brigade *bb;
+
+    apr_json_value_t *sub_log_object;
+    apr_json_value_t *json_null = apr_json_null_create(r->pool);
+    apr_json_value_t *json_empty_string = apr_json_string_create(r->pool, "\"\"", APR_JSON_VALUE_STRING);
+
+    /* build log object */
+    apr_json_value_t *log_object = apr_json_object_create(r->pool);
+
+    for (int i = 0; i < lfd->nelts; ++i) {
+        if(!items[i].tag) {
+            continue;
+        }
+
+        if (formatter_options->use_short_attribute_names) {
+            attribute_name = NULL;
+        }
+        else {
+            attribute_name = apr_hash_get(json_hash, items[i].tag, APR_HASH_KEY_STRING);
+        }
+
+        if (!attribute_name) {
+            attribute_name = items[i].tag; /* fallback: use tag as attribute name */
+        }
+
+        /* if the current tag has arguments, i.e. "i" (requestHeaders)
+         * create a sub object with items[i].arg as key and lfd->portions[i] values.
+         *
+         * as multiple same tags with different arguments can exist,
+         * the log_format_item is sorted before, so we can easily check for
+         * gruppenwechsel
+         */
+        if(items[i].arg != NULL && strlen(items[i].arg) > 0) {
+            /* start sub object */
+            sub_log_object = apr_json_object_create(r->pool);
+
+            apr_json_object_set(log_object,
+                                attribute_name, APR_JSON_VALUE_STRING,
+                                sub_log_object,
+                                r->pool);
+
+            for (int j = i; j < lfd->nelts; ++j) {
+                if(!items[j].tag) {
+                    continue;
+                }
+                if(strcmp(items[j].tag, items[i].tag) != 0) {
+                    i = j - 1;
+                    break;
+                }
+
+                attribute_name = items[j].arg;
+                if (!lfd->portions[j]) {
+                    apr_json_object_set(sub_log_object,
+                                        attribute_name, APR_JSON_VALUE_STRING,
+                                        json_null, r->pool);
+                }
+                else if (!*lfd->portions[j]) {
+                    apr_json_object_set(sub_log_object,
+                                        attribute_name, APR_JSON_VALUE_STRING,
+                                        json_empty_string,
+                                        r->pool);
+                }
+                else {
+                    attribute_value = lfd->portions[j];
+                    apr_json_object_set(sub_log_object,
+                                        attribute_name, APR_JSON_VALUE_STRING,
+                                        apr_json_string_create(r->pool, attribute_value,
+                                                               APR_JSON_VALUE_STRING),
+                                        r->pool);
+                }
+            }
+
+            continue;
+        }
+
+        if (!lfd->portions[i]) {
+            apr_json_object_set(log_object,
+                                attribute_name, APR_JSON_VALUE_STRING,
+                                json_null, r->pool);
+
+        }
+        else if (!*lfd->portions[i]) {
+            apr_json_object_set(log_object,
+                                attribute_name, APR_JSON_VALUE_STRING,
+                                json_empty_string,
+                                r->pool);
+        }
+        else {
+            attribute_value = lfd->portions[i];
+            apr_json_object_set(log_object,
+                                attribute_name, APR_JSON_VALUE_STRING,
+                                apr_json_string_create(r->pool, attribute_value,
+                                                       APR_JSON_VALUE_STRING),
+                                r->pool);
+
+        }
+    }
+
+    ba = apr_bucket_alloc_create(r->pool);
+    bb = apr_brigade_create(r->pool, ba);
+
+    /* encode into bucket brigade */
+    apr_json_encode(bb, NULL, NULL, log_object, APR_JSON_FLAGS_WHITESPACE, r->pool);
+    apr_brigade_puts(bb, NULL, NULL, APR_EOL_STR);
+
+    // TODO: detect overflow nvec
+    apr_brigade_to_iovec(bb, vec, &nvec);
+
+    lfdj->portions = apr_palloc(r->pool, sizeof(char*) * nvec);
+    lfdj->lengths = apr_palloc(r->pool, sizeof(int) * nvec);
+    lfdj->total_len = 0;
+    for(int i = 0; i < nvec; i++) {
+        lfdj->portions[i] = vec[i].iov_base;
+        lfdj->lengths[i] = vec[i].iov_len;
+        lfdj->total_len += vec[i].iov_len;
+    }
+    lfdj->nelts = nvec;
+    return lfdj;
+}
+#else
 static apr_size_t add_str(apr_array_header_t * ah_strs, apr_array_header_t * ah_strl, const char *str)
 {
-	char** strs;
-	int* strl;
+    char** strs = apr_array_push(ah_strs);
+    int* strl = apr_array_push(ah_strl);
 
-    strs = apr_array_push(ah_strs);
     *strs = (char*) str;
-
-    strl = apr_array_push(ah_strl);
     *strl = strlen(str);
 
     return *strl;
@@ -1787,105 +1928,56 @@ static ap_log_formatted_data * ap_json_log_formatter( request_rec *r,
     const char* attribute_name;
     const char* attribute_value;
 
+    ap_log_formatted_data *lfdj = apr_palloc(r->pool, sizeof(ap_log_formatted_data));
     apr_array_header_t *strs = apr_array_make(r->pool, lfd->nelts * 3, sizeof(char *)); /* array of pointers to char */
     apr_array_header_t *strl = apr_array_make(r->pool, lfd->nelts * 3, sizeof(int));    /* array of int (strlen) */
 
-    ap_log_formatted_data *lfdj = apr_palloc(r->pool, sizeof(ap_log_formatted_data));
     lfdj->total_len = 0;
 
     /* build json object */
     lfdj->total_len += add_str(strs, strl, "{");
     for (int i = 0; i < lfd->nelts; ++i) {
-        if(!items[i].tag) {
-            continue;
+        if(items[i].tag == NULL) {
+                continue;
         }
 
         if (formatter_options->use_short_attribute_names) {
-            attribute_name = NULL;
+                attribute_name = NULL;
         }
         else {
             attribute_name = apr_hash_get(json_hash, items[i].tag, APR_HASH_KEY_STRING);
         }
 
         if (!attribute_name) {
-            attribute_name = ap_escape_logjson(r->pool, items[i].tag, NULL, 0); /* fallback: use tag as attribute name */
+            attribute_name = ap_escape_logjson(r->pool, items[i].tag, NULL, 0); /* use tag as attribute name as fallback */
         }
 
         lfdj->total_len += add_str(strs, strl, "\"");
         lfdj->total_len += add_str(strs, strl, attribute_name);
-        lfdj->total_len += add_str(strs, strl, "\":");
 
-        /* if the current tag has arguments, i.e. "i" (requestHeaders)
-         * create a sub object with items[i].arg as key and lfd->portions[i] values.
-         *
-         * as multiple same tags with different arguments can exist,
-         * the log_format_item is sorted before, so we can easily check for
-         * gruppenwechsel
-         */
+        /* process any arguments as attribute name extension */
         if(items[i].arg != NULL && strlen(items[i].arg) > 0) {
-            // start sub object
-            lfdj->total_len += add_str(strs, strl, "{");
-            for (int j = i; j < lfd->nelts; ++j) {
-                if(!items[j].tag) {
-                    continue;
-                }
-                if(strcmp(items[j].tag, items[i].tag) != 0) {
-                    i = j - 1;
-                    break;
-                }
-                attribute_name = ap_escape_logjson(r->pool, items[j].arg, NULL, 0);
-
-                lfdj->total_len += add_str(strs, strl, "\"");
-                lfdj->total_len += add_str(strs, strl, attribute_name);
-                lfdj->total_len += add_str(strs, strl, "\":");
-
-                if (!lfd->portions[j]) {
-                    lfdj->total_len += add_str(strs, strl, "null");
-                }
-                else if (!*lfd->portions[j]) {
-                    lfdj->total_len += add_str(strs, strl, "\"\"");
-                }
-                else {
-                    lfdj->total_len += add_str(strs, strl, "\"");
-                    attribute_value = ap_escape_logjson(r->pool, lfd->portions[j], NULL, 0);
-                    lfdj->total_len += add_str(strs, strl, attribute_value);
-                    lfdj->total_len += add_str(strs, strl, "\"");
-                }
-                lfdj->total_len += add_str(strs, strl, ",");
-            }
-            // remove last ","
-            apr_array_pop(strs);
-            lfdj->total_len -= *(int *)apr_array_pop(strl);
-
-            // end sub object
-            lfdj->total_len += add_str(strs, strl, "},");
-            continue;
+            attribute_name = ap_escape_logjson(r->pool, items[i].arg, NULL, 0);
+            lfdj->total_len += add_str(strs, strl, " ");
+            lfdj->total_len += add_str(strs, strl, attribute_name);
         }
+        lfdj->total_len += add_str(strs, strl, "\":\"");
 
-        if (!lfd->portions[i]) {
-            lfdj->total_len += add_str(strs, strl, "null");
-        }
-        else if (!*lfd->portions[i]) {
-            lfdj->total_len += add_str(strs, strl, "\"\"");
-        }
-        else {
-            lfdj->total_len += add_str(strs, strl, "\"");
-            attribute_value = ap_escape_logjson(r->pool, lfd->portions[i], NULL, 0);
-            lfdj->total_len += add_str(strs, strl, attribute_value);
-            lfdj->total_len += add_str(strs, strl, "\"");
-        }
-        lfdj->total_len += add_str(strs, strl, ",");
+        attribute_value = ap_escape_logjson(r->pool, lfd->portions[i], NULL, 0);
+        lfdj->total_len += add_str(strs, strl, attribute_value);
+        lfdj->total_len += add_str(strs, strl, "\",");
     }
-    /* replace last ',' with '}' */
+    /* replace last '",' with '"}' */
     apr_array_pop(strs);
     lfdj->total_len -= *(int *)apr_array_pop(strl);
-    lfdj->total_len += add_str(strs, strl, "}" APR_EOL_STR);
+    lfdj->total_len += add_str(strs, strl, "\"}" APR_EOL_STR);
 
     lfdj->portions = (const char **)strs->elts;
     lfdj->lengths = (int *)strl->elts;
     lfdj->nelts = strs->nelts;
     return lfdj;
 }
+#endif
 
 static apr_status_t ap_default_log_writer( request_rec *r,
                            void *handle,
@@ -2074,7 +2166,7 @@ static apr_status_t ap_buffered_log_writer(request_rec *r,
 
 static void json_register_attribute(apr_pool_t *p, const char *tag, const char* attribute_name)
 {
-    apr_hash_set(json_hash, tag, strlen(tag), (const void *)ap_escape_logjson(p, attribute_name, NULL, 0));
+    apr_hash_set(json_hash, tag, strlen(tag), attribute_name);
 }
 
 static int log_pre_config(apr_pool_t *p, apr_pool_t *plog, apr_pool_t *ptemp)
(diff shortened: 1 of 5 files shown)


thomasmey:
@ylavic May you have a look at this?

server/util.c
```diff
@@ -2201,6 +2201,219 @@ AP_DECLARE(char *) ap_escape_html2(apr_pool_t *p, const char *s, int toasc)
     x[j] = '\0';
     return x;
 }
+
+/* TODO: not sure about copyright&license of below code */
+
```
minfrin, line 2475: Digging further this appears to be MIT licensed.

minfrin:
One observation - there is a json encoder and decoder in apr-util that is Apache licensed.

https://github.com/apache/apr-util/blob/1.7.x/json/apr_json_encode.c

It is not released yet, and it's reasonable not to have to wait around until it is, but if you're looking for code to bring in that's already Apache, this should do.

The encoder code is designed to meet the strictness requirements of JSON parsing in JWT.

thomasmey:
> One observation - there is a json encoder and decoder in apr-util that is Apache licensed.
> 
> https://github.com/apache/apr-util/blob/1.7.x/json/apr_json_encode.c
> 
> It is not released yet, and it's reasonable not to have to wait around until it is, but if you're looking for code to bring in that's already Apache, this should do.
> 
> The encoder code is designed to meet the strictness requirements of JSON parsing in JWT.

so, i'm not sure what to do now, the code for utf-8 parsing was initially propsed by @ylavic 

should I rewrite to use above apr_json_encode.c which is available only in apr-util 1.7.x?
how to even back port this to 2.4.x which is my original goal to provide this functionality on 2.4.x?

thomasmey:
@minfrin see original discussion here: https://github.com/apache/httpd/pull/353

minfrin:
> so, i'm not sure what to do now, the code for utf-8 parsing was initially propsed by @ylavic
> 
> should I rewrite to use above apr_json_encode.c which is available only in apr-util 1.7.x? how to even back port this to 2.4.x which is my original goal to provide this functionality on 2.4.x?

In the short term to get it into 2.4.x, depend on the external library, but in the longer term it would be better if the external dependency could be avoided. Ideally if you can herd the json code together so that it can be #ifdef'ed easily either to support apr_json now or perhaps in future if you don't have the resources to do something like this right now.

thomasmey:
Hi @minfrin I added a ifdef, one implementation using apr json and the other one the custom utf8 escaping. for apr json i added some TODO comments, not sure how to best cope with max numer of iovec elements, any adivse welcome,

what do you think?

thomasmey:
can someone please approve the github workflow?

thomasmey:
would @ylavic also please review again?

server/util.c
minfrin: One possible thing to do is rework the escaping code from APR, which is Apache licensed and already belongs to the ASF:

https://github.com/apache/apr-util/blob/1.7.x/json/apr_json_encode.c#L63

modules/loggers/mod_log_config.c
```diff
@@ -1449,9 +1560,10 @@ static const char *set_buffered_logs_on(cmd_parms *parms, void *dummy, int flag)
     }
     return NULL;
 }
+
 static const command_rec config_log_cmds[] =
 {
-AP_INIT_TAKE23("CustomLog", add_custom_log, NULL, RSRC_CONF,
+AP_INIT_TAKE_ARGV("CustomLog", add_custom_log, NULL, RSRC_CONF,
```
minfrin, line 1571: Need to make the same change to GlobalLog to match.

modules/loggers/mod_log_config.c
```diff
@@ -1666,13 +1782,229 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+        /* process any arguments as attribute name extension */
+        if(items[i].arg != NULL && strlen(items[i].arg) > 0) {
+            attribute_name = ap_escape_logjson(r->pool, items[i].arg, NULL, 0);
+            lfdj->total_len += add_str(strs, strl, " ");
+            lfdj->total_len += add_str(strs, strl, attribute_name);
+        }
+        lfdj->total_len += add_str(strs, strl, "\":\"");
+
+        attribute_value = ap_escape_logjson(r->pool, lfd->portions[i], NULL, 0);
+        lfdj->total_len += add_str(strs, strl, attribute_value);
```
minfrin: I'm getting a segfault here, as in my case attribute_value is NULL, which is strlen()'d and then it crashes.

It gets safely through ap_escape_logjson(), lfd->portions[i] is NULL in my case.

Looks like NULL is a valid value in `config_log_transaction()`, it's not handled later.

modules/loggers/mod_log_config.c
```diff
@@ -1902,6 +2239,51 @@ static int log_pre_config(apr_pool_t *p, apr_pool_t *plog, apr_pool_t *ptemp)
...
     }
 
+    /*
+     * should align with tomcat attribute names:
+     * https://github.com/apache/tomcat/blob/9.0.x/java/org/apache/catalina/valves/JsonAccessLogValve.java#L78
+     * also interessting is fluentd resp. fluent-bit:
+     * - https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/parser_apache2.rb#L72
+     */
+    if (log_pfn_register) {
+        json_register_attribute(p, "a", "remoteAddr");
```
minfrin: Rather than `json_register_attribute()`, it would be cleaner to implement an extended `log_pfn_register()` function, something like `log_pfn_register_ex()`, that takes the extra long form of the name.

`log_pfn_register()` can then be kept for backwards compatibility, with the long form name being NULL. Any module that doesn't provide a long form parameter is just ignored by the json formatter.

This will be safe to backport to v2.4.

modules/loggers/mod_log_config.c
```diff
@@ -1666,13 +1782,229 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+    return lfdj;
+}
+#else
+static apr_size_t add_str(apr_array_header_t * ah_strs, apr_array_header_t * ah_strl, const char *str)
+{
+    char** strs = apr_array_push(ah_strs);
+    int* strl = apr_array_push(ah_strl);
+
+    *strs = (char*) str;
+    *strl = strlen(str);
```
minfrin, line 1993: Worked around the segfault with:

`    *strl = str ? strlen(str) : 0;
`

modules/loggers/mod_log_config.c
```diff
@@ -1666,13 +1782,229 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+
+        if (formatter_options->use_short_attribute_names) {
+                attribute_name = NULL;
+        }
+        else {
+            attribute_name = apr_hash_get(json_hash, items[i].tag, APR_HASH_KEY_STRING);
+        }
+
+        if (!attribute_name) {
+            attribute_name = ap_escape_logjson(r->pool, items[i].tag, NULL, 0); /* use tag as attribute name as fallback */
```
minfrin: The `ap_escape_logjson()` is only needed in the not-apr-1.7 case, as the APR code does all escaping internally.

minfrin:
In the test, the following got logged, which is a win:

```
{"remoteAddr":"127.0.0.1","size":"236","logicalUserName":"","request":"GET / HTTP/1.1","statusCode":"404","time":"[17/Dec/2023:19:52:58 +0000]","user":""}
```

In the test there was no "logicalUserName" or "user", but these were returned as an empty string. I suspect we need to alter this to not add the parameter if the parameter doesn't exist. Something like this:

```
    /* build json object */
    lfdj->total_len += add_str(strs, strl, "{");
    for (i = 0; i < lfd->nelts; ++i) {
        if (items[i].tag == NULL) {
            continue;
        }

        attribute_value = ap_escape_logjson(r->pool, lfd->portions[i], NULL, 0);

        if (!attribute_value) {
            continue;
        }

        if (formatter_options->use_short_attribute_names) {
            attribute_name = NULL;
        }
        else {
            attribute_name = apr_hash_get(json_hash, items[i].tag, APR_HASH_KEY_STRING);
        }

        if (!attribute_name) {
            attribute_name = ap_escape_logjson(r->pool, items[i].tag, NULL, 0); /* use tag as attribute name as fallback */
        }

        lfdj->total_len += add_str(strs, strl, "\"");
        lfdj->total_len += add_str(strs, strl, attribute_name);

        /* process any arguments as attribute name extension */
        if(items[i].arg != NULL && strlen(items[i].arg) > 0) {
            attribute_name = ap_escape_logjson(r->pool, items[i].arg, NULL, 0);
            lfdj->total_len += add_str(strs, strl, " ");
            lfdj->total_len += add_str(strs, strl, attribute_name);
        }
        lfdj->total_len += add_str(strs, strl, "\":\"");

        lfdj->total_len += add_str(strs, strl, attribute_value);
        lfdj->total_len += add_str(strs, strl, "\",");
    }
```

thomasmey:
> In the test, the following got logged, which is a win:
> 
> ```
> {"remoteAddr":"127.0.0.1","size":"236","logicalUserName":"","request":"GET / HTTP/1.1","statusCode":"404","time":"[17/Dec/2023:19:52:58 +0000]","user":""}
> ```
> 
> In the test there was no "logicalUserName" or "user", but these were returned as an empty string. I suspect we need to alter this to not add the parameter if the parameter doesn't exist. Something like this:
> 
oops, sorry I somehow managed to add in an old version of the non-apr-json code, let me fix.

But the question remains: should we add NULL or empty string values to the resulting JSON?

minfrin:
> But the question remains: should we add NULL or empty string values to the resulting JSON?

I would leave out the whole kv pair completely, no need to take up space logging something that represents it wasn't there.

An empty string would be confusing, as you wouldn't be able to distinguish between the case where a user was missing or a user was the empty string. This was be nasty to troubleshoot.

modules/loggers/mod_log_config.c
```diff
@@ -395,49 +405,45 @@ static const char *log_request_line(request_rec *r, char *a)
...
 {
     if (a && !strcmp(a, "c")) {
-        return r->connection->log_id ? r->connection->log_id : "-";
+        return r->connection->log_id ? r->connection->log_id : NULL;
     }
     else {
-        return r->log_id ? r->log_id : "-";
+        return r->log_id ? r->log_id : NULL;
     }
 }
```
ylavic, line 442: Could return r->[connection->]log_id directly.

modules/loggers/mod_log_config.c
```diff
@@ -961,7 +968,7 @@ static char *parse_log_misc_string(apr_pool_t *p, log_format_item *it,
      */
     it->arg = apr_palloc(p, s - *sa + 1);
 
-    d = it->arg;
+    d = (char *)it->arg;
```
ylavic: `d = apr_palloc(...); it->arg = d;` or `it->arg = d = apr_palloc(...);` would avoid the const away cast.

modules/loggers/mod_log_config.c
```diff
@@ -1078,22 +1086,24 @@ static char *parse_log_item(apr_pool_t *p, log_format_item *it, const char **sa)
...
-                return apr_pstrcat(p, "Unrecognized LogFormat directive %",
-                               dummy, NULL);
+                handler = (ap_log_handler *)apr_hash_get(log_hash, s, 1);
+                if (!handler) {
+                    char dummy[2];
+
+                    dummy[0] = s[0];
+                    dummy[1] = '\0';
+                    return apr_pstrcat(p, "Unrecognized LogFormat directive %",
+                                   dummy, NULL);
```
ylavic, line 1106:
```
    return apr_psprintf(p, "Unrecognized LogFormat directive %%%.*s", 1, s);
```
?

modules/loggers/mod_log_config.c
```diff
@@ -1433,7 +1527,23 @@ static const char *add_global_log(cmd_parms *cmd, void *dummy, const char *fn,
...
+./modules/http2/h2.h
+
+#define H2_ALEN(a)          (sizeof(a)/sizeof((a)[0]))
+./modules/http2/h2_proxy_session.h
+
+#define TLS_DIM(a)      (sizeof(a)/sizeof(a[0]))
+    apr_size_t len, bmax = sizeof(buffer)/sizeof(buffer[0]); \
+./modules/tls/tls_util.h
+
+     */
```
ylavic, line 1551: Yeah, we probably need an `AP_ARRAY_LEN()` in httpd.h or so.

modules/loggers/mod_log_config.c
```diff
@@ -1513,6 +1624,10 @@ static int open_multi_logs(server_rec *s, apr_pool_t *p)
                 format = apr_table_get(mls->formats, cls->format_string);
                 if (format) {
                     cls->format = parse_log_string(p, format, &dummy);
+                    if(cls->format && cls->log_formatter == ap_json_log_formatter /* sort_log_format_items */ ) {
+                        /* sort log_format_items array, to help json formatter */
+                        qsort(cls->format->elts, cls->format->nelts, cls->format->elt_size, compare_log_format_item_by_tag);
```
ylavic: Do we need to do this both here and in add_custom_log()?

modules/loggers/mod_log_config.c
```diff
@@ -1666,13 +1781,279 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
 
+#if APU_MAJOR_VERSION > 1 || (APU_MAJOR_VERSION == 1 && APU_MINOR_VERSION >= 7)
+static ap_log_formatted_data * ap_json_log_formatter( request_rec *r,
+                           void *handle,
+                           ap_log_formatted_data *lfd,
+                           const void *itms)
+
+{
+    json_log_formatter_options * formatter_options = (json_log_formatter_options *) handle;
+    log_format_item *items = (log_format_item *) itms;
```
ylavic: I don't think an `ap_log_formatter` function should modify its `handle` parameter so it could be `const void *handle`?
Maybe the `lfd` could be returned as-is and not copied by some implementation (not here) so it makes sense to leave it non-const, or return a `const ap_log_formatted_data*` dunno.
Anyway, I think these locals should be const here:
```
    const json_log_formatter_options *formatter_options = handle;
    const log_format_item *items = itms;
```

modules/loggers/mod_log_config.c
```diff
@@ -1666,13 +1781,279 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+
+    ba = apr_bucket_alloc_create(r->pool);
+    bb = apr_brigade_create(r->pool, ba);
+
+    /* encode into bucket brigade */
+    apr_json_encode(bb, NULL, NULL, log_object, APR_JSON_FLAGS_WHITESPACE, r->pool);
+    apr_brigade_puts(bb, NULL, NULL, APR_EOL_STR);
+
+    /* TODO: detect overflow nvec */
+    apr_brigade_to_iovec(bb, vec, &nvec);
```
ylavic: This dance makes me wonder whether we actually should use the apr_json API here (when it's available).
All we need is to build a json string (we won't work with or process the json values) so having to build an object value with key+value strings then serializing/encoding it in a bucket brigade then splitting the brigade in an iovec looks overkill[1] to me for the usage here.
If we need to keep ap_escape_logjson() anyway because apr-util < 1.7 or because it's usefull on its own (or elsewhere[2]) we possibly can keep the code simple(r) by just encoding the json string as we go so keep the `#else` ap_json_log_formatter() only.
If we keep this code we should avoid apr_brigade_to_iovec(), it seems indeed quite hard to use.. Using apr_brigade_pflatten() would create yet another buffer (possibly unbounded..) so maybe have our own helper like:
```
static apr_status_t brigade_to_iovec(apr_bucket_brigade *bb,
                                     struct iovec **pvec, int *nvec,
                                     apr_pool_t *p)
{
    apr_status_t rv = APR_SUCCESS;
    apr_array_header_t *a;
    const char *str;
    apr_size_t len;
    apr_bucket *b;
    int n = 0;

    for (b = APR_BRIGADE_FIRST(bb);
         b != APR_BRIGADE_SENTINEL(bb);
         b = APR_BUCKET_NEXT(b)) {
         n++;
    }
    a = apr_array_make(p, n, sizeof(struct iovec));
    for (b = APR_BRIGADE_FIRST(bb);
         b != APR_BRIGADE_SENTINEL(bb);
         b = APR_BUCKET_NEXT(b)) {
        rv = apr_bucket_read(b, &str, &len, APR_NONBLOCK_READ);
        if (rv != APR_SUCCESS) {
            break;
        }
        if (len) {
            struct iovec *iov = &APR_ARRAY_PUSH(a, struct iovec);
            iov->iov_base = (void *)str;
            iov->iov_len = len;
        }
    }
    *pvec = (void *)a->elts;
    *nvec = a->nelts;

    return rv;
}
```
?

[1] I'm not saying that the API is overkill since it's very well suited when one has to build or parse json and then work with the values (like mod_autht_jwk), but it's not what we do here (just produce a json string..). Btw I think that `apr_json_encode()` should take flags for what to do with non-utf8 strings (as-is/nocheck, replacement char, assume ISO-8859-1, error out, make tea..), likewise for `apr_json_string_create[_ex]()` because the closer to where the strings come from, the better (IMHO).

[2] mod_log_json is using libjansson (yet something else..) but with the `_nocheck` setters, ap_escape_logjson() could "enforce" valid json there too.

modules/loggers/mod_log_config.c
```diff
@@ -1666,13 +1781,279 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+}
+
+static ap_log_formatted_data * ap_json_log_formatter( request_rec *r,
+                           void *handle,
+                           ap_log_formatted_data *lfd,
+                           const void *itms)
+
+{
+    json_log_formatter_options * formatter_options = (json_log_formatter_options *) handle;
+    log_format_item *items = (log_format_item *) itms;
```
ylavic: Likewise about constantness here.

modules/loggers/mod_log_config.c
```diff
@@ -1666,13 +1781,279 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+         */
+        if(items[i].arg != NULL && strlen(items[i].arg) > 0) {
+            /* start sub object */
+            lfdj->total_len += add_str(strs, strl, "{");
+            for (j = i; j < lfd->nelts; ++j) {
+                if(items[j].tag == NULL) {
+                    continue;
+                }
+                if(strcmp(items[j].tag, items[i].tag) != 0) {
+                    i = j; /* -1 ?*/
```
ylavic: Hm, yes I think `i = j - 1` is needed (otherwise it would skip this item on the next/upper ++i loop), but this `i = j - 1` needs to go before the `continue` below (i.e. in any case) because if it doesn't `break` here we still need to exit both loops when/if `j` reaches `lfd->nelts`, no?
Also maybe this test should be `if (j > i && strcmp(items[j].tag, items[i].tag) != 0)` to make it clear that the first value is always processed (even though if j == i the strcmp() is always 0 anyway), meaning that this will never loop infinitely because `j` will always advance at least by 1 here.

modules/loggers/mod_log_config.c
```diff
@@ -1666,13 +1781,279 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+                    lfdj->total_len += add_str(strs, strl, "\"");
+                    attribute_value = ap_escape_logjson(r->pool, lfd->portions[i], NULL, 0);
+                    lfdj->total_len += add_str(strs, strl, attribute_value);
+                    lfdj->total_len += add_str(strs, strl, "\"");
+                }
+                lfdj->total_len += add_str(strs, strl, ",");
+            }
+            /* remove last "," */
+            apr_array_pop(strs);
+            lfdj->total_len -= *(int *)apr_array_pop(strl);
```
ylavic: There is nothing to remove if all the items->tag were NULL in the `j` loop, is it possible?
Rather than removing maybe we could add if/when needed only, like:
```
            int found_j = 0;
            for (j = i; j < lfd->nelts; ++j) {
                if (items[j].tag == NULL) {
                    continue;
                }
                if (found_j) {
                    lfdj->total_len += add_str(strs, strl, ",");
                }
                else {
                    found_j = 1;
                }
                ...
           }
```

modules/loggers/mod_log_config.c
```diff
@@ -1666,13 +1781,279 @@ static ap_log_writer *ap_log_set_writer(ap_log_writer *handle)
...
+            lfdj->total_len += add_str(strs, strl, "\"");
+            attribute_value = ap_escape_logjson(r->pool, lfd->portions[i], NULL, 0);
+            lfdj->total_len += add_str(strs, strl, attribute_value);
+            lfdj->total_len += add_str(strs, strl, "\"");
+        }
+        lfdj->total_len += add_str(strs, strl, ",");
+    }
+    /* replace last ',' with '}' */
+    apr_array_pop(strs);
+    lfdj->total_len -= *(int *)apr_array_pop(strl);
```
ylavic: Likewise here with some `found_i`.

server/util.c
ylavic: As explained in the added code for using apr_json, I'm not sure it's worth the added complexity here.

modules/loggers/mod_log_config.c
ylavic: This is the not-apr-util-1.7 case here IIUC.

modules/loggers/mod_log_config.c
thomasmey: should be fixed now

modules/loggers/mod_log_config.c
thomasmey: should be fixed now

modules/loggers/mod_log_config.c
thomasmey: yes, I agree this code is in the apr-util < 1.7 ifdef

ylavic:
> > In the test, the following got logged, which is a win:
> > ```
> > {"remoteAddr":"127.0.0.1","size":"236","logicalUserName":"","request":"GET / HTTP/1.1","statusCode":"404","time":"[17/Dec/2023:19:52:58 +0000]","user":""}
> > ```
> >   
> > In the test there was no "logicalUserName" or "user", but these were returned as an empty string. I suspect we need to alter this to not add the parameter if the parameter doesn't exist. Something like this:
> 
> oops, sorry I somehow managed to add in an old version of the non-apr-json code, let me fix.
> 
> But the question remains: should we add NULL or empty string values to the resulting JSON?

I think the empty values should be `""` and the missing ones `null`, when configured a missing value should not be omitted.

minfrin:
> > But the question remains: should we add NULL or empty string values to the resulting JSON?
> 
> I think the empty values should be `""` and the missing ones `null`, when configured a missing value should not be omitted.

Is this not going to waste a lot of disk space and cycles?

Think of a time when we might be logging say 20 different attributes, and 10 of them are null. Seems to be wasteful.

ylavic:
> Is this not going to waste a lot of disk space and cycles?
> 
> Think of a time when we might be logging say 20 different attributes, and 10 of them are null. Seems to be wasteful.

Yeah, but then the LogFormat is explicit in the log, no need to know it for figuring out what's missing/expected. Not a strong opinion though, maybe space is more of a concern with json, there is no parsing/column issue like with plain lines (and the need for `-`).
Let's see what others think..

covener:
>  there is no parsing/column issue like with plain lines (and the need for -).

Seems like a good argument for omitting.

modules/loggers/mod_log_config.c
thomasmey: I added log_pfn_register_ex()

modules/loggers/mod_log_config.c
thomasmey: This is the original code, from trunk/2.4, I just did change some indentation, but I can adjust to above suggesion if you want

modules/loggers/mod_log_config.c
thomasmey: I did think about this one a bit today and pushed the commit https://github.com/apache/httpd/pull/373/commits/9f8355c35977ab7811afb60555b2e9303517bf80 - what do you think?

thomasmey:
At least for json null values we could add a config option like "omitnulls", to not serialize those attributes. not sure what default for this options should be.

modules/loggers/mod_log_config.c
thomasmey: thinking a bit more about this I think the lookup of log formats happening currently in "ap_hook_open_logs(init_config_log" should IMHO actually be done in "ap_hook_check_config(log_check_config" as part of consistency check of the current configuration.

So for example having a:
CustomLog "logs/access_log" combind

Will get silently parsed as an empty format string, and not rejected as an error of misspelled format nickname "combined", which is probably not what the user wants.

Also the check in "config_log_transaction":
format = cls->format ? cls->format : default_format;
can actually only happen in TransferLog case in which case fallback to default format is done.

Maybe rejecting format string which doesn't contain any format specifier, if it's not a format nickname is a sensible change?

and moving all those checks/consolidating into log_check_config let's probably get rid of some invariants.
but thats probably stuff for another PR...

covener: > Maybe rejecting format string which doesn't contain any format specifier, if it's not a format nickname is a sensible change?

+1, I have seen it bite people many times and I don't see a reason to preserve it.

modules/loggers/mod_log_config.c
thomasmey: should be fixed now

minfrin:
As an aside, I added support for JSON escaping to APR, which I plan to backport to v1.x: https://github.com/apache/apr/commit/3646493027dbf0dcc1523d1ebf182a5792aa70a3

thomasmey:
> As an aside, I added support for JSON escaping to APR, which I plan to backport to v1.x: [apache/apr@3646493](https://github.com/apache/apr/commit/3646493027dbf0dcc1523d1ebf182a5792aa70a3)

one obersavation: The esapce functions do return "NULL" if the input is NULL, but JSON null is lower case, or doesn't it matter?

modules/loggers/mod_log_config.c
thomasmey: this should be fixed with latest commit which does implement omitting of null values.
also regarding your first question "There is nothing to remove if all the items->tag were NULL in the j loop, is it possible?" this should be also fixed with latest commit, but generall answer is: as all item[*].tags are sorted before usage in ap_json_log_formatter anyway, because of Gruppenwechsel logic, all NULL tags will be first and then later all non-nulls, so this check was in the code before the gruppenwechsel and sub object logic, and can probaly get removed now.

modules/loggers/mod_log_config.c
thomasmey: should be fixed with latest commit which implements omit null values

modules/loggers/mod_log_config.c
thomasmey: fixed

thomasmey:
> As an aside, I added support for JSON escaping to APR, which I plan to backport to v1.x: [apache/apr@3646493](https://github.com/apache/apr/commit/3646493027dbf0dcc1523d1ebf182a5792aa70a3)

I did port your code into httpd, one questions regarding handling of those new functions:
The apr-trunk builds do fail now, probably because of duplicate function definition, see:
https://github.com/thomasmey/httpd/actions/runs/7287742760/job/19858988307

how to cope with that in general? is there a way to make those function definitons conditional on apr version or if linked against an apr version that already has those methods? how would such thing be done in C/autogen/configure/#ifdef?

minfrin:
> one obersavation: The esapce functions do return "NULL" if the input is NULL, but JSON null is lower case, or doesn't it matter?

It does matter - thank you for finding that. Fixed.

minfrin:
> > As an aside, I added support for JSON escaping to APR, which I plan to backport to v1.x: [apache/apr@3646493](https://github.com/apache/apr/commit/3646493027dbf0dcc1523d1ebf182a5792aa70a3)
> 
> I did port your code into httpd, one questions regarding handling of those new functions: The apr-trunk builds do fail now, probably because of duplicate function definition, see: https://github.com/thomasmey/httpd/actions/runs/7287742760/job/19858988307
> 
> how to cope with that in general? is there a way to make those function definitons conditional on apr version or if linked against an apr version that already has those methods? how would such thing be done in C/autogen/configure/#ifdef?

Ah, to bring anything in as a public function to httpd, give it the prefix ap_something(). The apr_something prefix is for APR stuff only.

In the long term, all of the encoding/decoding code in httpd is being deprecated and will be switched over to use the APR functions exclusively, but because of backwards compatibility the old functions are likely to be around for a long time.

The new functions use the tiniest memory footprint possible, rather than assuming a worst case scenario large output string (example: 3*strlen(in) for urlencoding), and then perform conversions on data that doesn't need escaping (that's most data), we work out the actual length of the string and whether conversion is needed at all as a separate step. We can then allocate the precise amount of memory needed, or nothing if not needed.

modules/loggers/mod_log_config.c
thomasmey: fixed

thomasmey:
Happy new year!
How to proceed with this pr?
Should I delete the now obsolete MIT licence utf8 handling code and squash all commits in the PR into one, so the code is really gone from git?

thomasmey:
Hi, @ylavic @minfrin @covener 

I did clean-up and rebased/squashed this code into a new branch, please have a look here:
https://github.com/apache/httpd/compare/trunk...thomasmey:httpd:json-trunk-clean?expand=1

If you like I can force push to thomasmey:json-trunk or create a new PR.

PR #377, opened 2023-08-16
mod_dav: add DavBasePath

notroj:


docs/manual/mod/mod_dav.xml
```diff
@@ -194,6 +194,36 @@ Alias "/php-source" "/home/gstein/php_files"
 </usage>
 </directivesynopsis>
 
+<directivesynopsis>
+<name>DavBasePath</name>
+<description>Configure repository root path</description>
+<syntax>DavBasePath <var>root-path</var></syntax>
+<default>DavBasePath /repos/</default>
```
tititiou36: Is "/repos/" as the default value really meaningful?

It is not hard-coded in the module and does not fit well with the comment "Take the repos root from DAVBasePath if configured, **else the path of the enclosing section**" in the [cb4eb29](https://github.com/apache/httpd/pull/377/commits/cb4eb2940eaab5103d9055e2fe25e5a3e717a132)

modules/dav/main/mod_dav.c
```diff
@@ -765,11 +779,27 @@ DAV_DECLARE(dav_error *) dav_get_resource(request_rec *r, int label_allowed,
...
-    err = (*conf->provider->repos->get_resource)(r, conf->dir,
+    err = (*conf->provider->repos->get_resource)(r, base,
                                                  label, use_checked_in,
                                                  res_p);
     if (err != NULL) {
+        /* In the error path, give a hint that DavBasePath needs to be
+         * used if the location was configured via a regex match. */
+        if (!conf->base) {
+            core_dir_config *conf = ap_get_core_module_config(r->per_dir_config);
+
```
rpluem, line 795: Is it a good idea to use the same name `conf`for the `core_dir_config` variable as for the `dav_dir_conf` variable and thus mask it in the block?

docs/manual/mod/mod_dav.xml
rpluem: Maybe

```suggestion
<default>None</default>
```

modules/dav/main/mod_dav.c
notroj: Definitely a bad idea! Fixed in 706af916933612750c1cf8f165e23d37c9b865b2

docs/manual/mod/mod_dav.xml
notroj: Thanks for catching this, fixed in 59c725a

notroj: Nope, definitely wouldn't be... thanks for reviewing.

notroj:
Thanks a lot for the reviews on this PR - updated now.

notroj:
Merged in ddd173bc63477cbbf45586f68c951e09659c7c4e

PR #378, opened 2023-08-24
Remove redundant slash

nisbet-hubbard:
The trailing slash duplicates opening slash from `DocumentRoot`.


notroj:
Thanks for the patch, but the change needed to merge here is to `mod_proxy.xml` rather than the generated HTML file.

nisbet-hubbard:
@notroj Done. Thanks for the pointer!

rbowen:
Thanks!

PR #379, opened 2023-08-24
Improve documentation on regex

nisbet-hubbard:
This PR addresses two issues:
1. `mod_setenvif` has different behaviour to `mod_rewrite` when it comes to backslash, since it relies on `ap_getword_conf` for parsing. This tends to create surprises when porting things between these modules, and is not documented.
2.  The docs are mostly silent on the subject of quoting regex. The practice of examples is haphazard: double quote seems to be preferred in general, but there are also instances of unquoted regex.


nisbet-hubbard:
Moved changes to xml files.

docs/manual/mod/mod_setenvif.xml
```diff
@@ -59,6 +59,16 @@ BrowserMatch MSIE !netscape
    are not separately evaluated in the subrequest due to the API phases
    <module>mod_setenvif</module> takes action in.</p>
 
+   <p>A regular expression only needs quoting when it contains space, 
+   in which case single and double quotes are equivalent. Unlike 
+   <code class="module"><a
```
rbowen: Hey, @nisbet-hubbard , I really like this PR, but have a couple of nitpicks.

Please use <module> (see line 60 above for example) rather than <code class="module">

rbowen:
Overall +1 to this patch. Thank you! I had one comment (use <module> rather than HTML <class> tags for module references.) Other than that, this PR looks ready to merge. I can make this change, if you like, but would rather you get credit for the patch thana me. Thanks again!

rbowen:
Also, general comment - it would be cool if we were consistent across all the docs re " vs '   

I tend to prefer " for the simple pragmatic reason that it's easier to see in various fonts.

docs/manual/mod/mod_setenvif.xml
nisbet-hubbard: Fixed. Thanks!

PR #383, opened 2023-09-01
Arrange the bybusyness logic and prevent bad busy values.

jfclere:
Use atomic to prevent busy value increasing and preventing bybusyness balancer working correctly.

### DIFF
--- a/modules/proxy/balancers/mod_lbmethod_bybusyness.c
+++ b/modules/proxy/balancers/mod_lbmethod_bybusyness.c
@@ -29,11 +29,11 @@ static APR_OPTIONAL_FN_TYPE(proxy_balancer_get_best_worker)
 static int is_best_bybusyness(proxy_worker *current, proxy_worker *prev_best, void *baton)
 {
     int *total_factor = (int *)baton;
+    apr_size_t current_busy = getbusy_count(current);
+    apr_size_t prev_best_busy = 0;
 
     current->s->lbstatus += current->s->lbfactor;
     *total_factor += current->s->lbfactor;
-    apr_size_t current_busy = getbusy_count(current);
-    apr_size_t prev_best_busy = 0;
     if (prev_best)
         prev_best_busy = getbusy_count(prev_best);
      
--- a/modules/proxy/balancers/mod_lbmethod_bybusyness.c
+++ b/modules/proxy/balancers/mod_lbmethod_bybusyness.c
@@ -25,18 +25,24 @@ module AP_MODULE_DECLARE_DATA lbmethod_bybusyness_module;
 static APR_OPTIONAL_FN_TYPE(proxy_balancer_get_best_worker)
                             *ap_proxy_balancer_get_best_worker_fn = NULL;
 
+
 static int is_best_bybusyness(proxy_worker *current, proxy_worker *prev_best, void *baton)
 {
     int *total_factor = (int *)baton;
 
     current->s->lbstatus += current->s->lbfactor;
     *total_factor += current->s->lbfactor;
+    apr_size_t current_busy = getbusy_count(current);
+    apr_size_t prev_best_busy = 0;
+    if (prev_best)
+        prev_best_busy = getbusy_count(prev_best);
+     
 
     return (
         !prev_best
-        || (current->s->busy < prev_best->s->busy)
+        || (current_busy < prev_best_busy)
         || (
-            (current->s->busy == prev_best->s->busy)
+            (current_busy == prev_best_busy)
             && (current->s->lbstatus > prev_best->s->lbstatus)
         )
     );
@@ -65,7 +71,7 @@ static apr_status_t reset(proxy_balancer *balancer, server_rec *s)
     worker = (proxy_worker **)balancer->workers->elts;
     for (i = 0; i < balancer->workers->nelts; i++, worker++) {
         (*worker)->s->lbstatus = 0;
-        (*worker)->s->busy = 0;
+        setbusy_count(*worker, 0); 
     }
     return APR_SUCCESS;
 }
--- a/modules/proxy/balancers/mod_lbmethod_bybusyness.c
+++ b/modules/proxy/balancers/mod_lbmethod_bybusyness.c
@@ -15,6 +15,7 @@
  */
 
 #include "mod_proxy.h"
+#include "proxy_util.h"
 #include "scoreboard.h"
 #include "ap_mpm.h"
 #include "apr_version.h"
--- a/modules/proxy/mod_proxy.h
+++ b/modules/proxy/mod_proxy.h
@@ -467,7 +467,7 @@ typedef struct {
     apr_size_t      recv_buffer_size;
     apr_size_t      io_buffer_size;
     apr_size_t      elected;    /* Number of times the worker was elected */
-    volatile apr_size_t      busy;       /* busyness factor */
+    apr_size_t      busy;       /* busyness factor */
     apr_size_t      response_field_size; /* Size of proxy response buffer in bytes. */
     apr_port_t      port;
     apr_off_t       transferred;/* Number of bytes transferred to remote */
@@ -1577,38 +1577,6 @@ PROXY_DECLARE(apr_status_t) ap_proxy_transfer_between_connections(
                                                        int *sent,
                                                        apr_off_t bsize,
                                                        int flags);
-/*
- * Get the busy counter from the shared worker memory
- *
- * @param worker_ Pointer to the worker structure.
- * @return      apr_size_t value atomically read for the worker.
- */
-PROXY_DECLARE(apr_size_t) getbusy_count(void *worker_);
-
-/*
- * Set the busy counter from the shared worker memory
- *
- * @param worker_ Pointer to the worker structure.
- * @param busy value to set the busy counter.
- * @return      void
- */
-PROXY_DECLARE(void) setbusy_count(void *worker_, apr_size_t busy);
-
-/*
- * decrement the busy counter from the shared worker memory
- *
- * @param worker_ Pointer to the worker structure.
- * @return      apr_status_t returns APR_SUCCESS.
- */
-PROXY_DECLARE(apr_status_t) decrement_busy_count(void *worker_);
-
-/*
- * increment the busy counter from the shared worker memory
- *
- * @param worker_ Pointer to the worker structure.
- * @return      void
- */
-PROXY_DECLARE(void) increment_busy_count(void *worker_);
 
 extern module PROXY_DECLARE_DATA proxy_module;
 
--- a/modules/proxy/mod_proxy.h
+++ b/modules/proxy/mod_proxy.h
@@ -467,7 +467,7 @@ typedef struct {
     apr_size_t      recv_buffer_size;
     apr_size_t      io_buffer_size;
     apr_size_t      elected;    /* Number of times the worker was elected */
-    apr_size_t      busy;       /* busyness factor */
+    volatile apr_size_t      busy;       /* busyness factor */
     apr_size_t      response_field_size; /* Size of proxy response buffer in bytes. */
     apr_port_t      port;
     apr_off_t       transferred;/* Number of bytes transferred to remote */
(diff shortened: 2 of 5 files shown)


modules/proxy/balancers/mod_lbmethod_bybusyness.c
```diff
@@ -25,18 +25,24 @@ module AP_MODULE_DECLARE_DATA lbmethod_bybusyness_module;
...
                             *ap_proxy_balancer_get_best_worker_fn = NULL;
 
+
 static int is_best_bybusyness(proxy_worker *current, proxy_worker *prev_best, void *baton)
 {
     int *total_factor = (int *)baton;
 
     current->s->lbstatus += current->s->lbfactor;
     *total_factor += current->s->lbfactor;
+    apr_size_t current_busy = getbusy_count(current);
```
rainerjung: To make the automated checks happy, you need to move the declarations in line 35 and 36 to the beginning of the block, at least before the first non-declaration line.

modules/proxy/mod_proxy.h
```diff
@@ -467,7 +467,7 @@ typedef struct {
     apr_size_t      recv_buffer_size;
     apr_size_t      io_buffer_size;
     apr_size_t      elected;    /* Number of times the worker was elected */
-    apr_size_t      busy;       /* busyness factor */
+    volatile apr_size_t      busy;       /* busyness factor */
```
ylavic: It's public interface so I think that it can't change without a MMN major bump. But I don't think `volatile` is actually needed here if worker->s->busy is **always** accessed with the new atomic functions (i.e. we shouldn't use bare worker->s->busy anywhere). Looks like one case is missing in proxy_balancer_get_best_worker(), not sure if the other accesses to worker->s->busy in the other lbmethods (!bybusiness) should use the new helpers for consistency too, looks sensible if the common proxy_balancer_get_best_worker() has to.

modules/proxy/mod_proxy.h
```diff
@@ -1577,6 +1577,38 @@ PROXY_DECLARE(apr_status_t) ap_proxy_transfer_between_connections(
...
+ */
+PROXY_DECLARE(apr_status_t) decrement_busy_count(void *worker_);
+
+/*
+ * increment the busy counter from the shared worker memory
+ *
+ * @param worker_ Pointer to the worker structure.
+ * @return      void
+ */
+PROXY_DECLARE(void) increment_busy_count(void *worker_);
```
ylavic: Couldnt' these busy_count helpers go to "proxy_util.h" if they are used internally only?
They should have a `proxy_` prefix in the name and use `proxy_worker*` instead of `void*` for the pointer to worker argument (for better type checking) in any case I think.

modules/proxy/proxy_util.c
```diff
@@ -4984,6 +4985,52 @@ PROXY_DECLARE(apr_status_t) ap_proxy_tunnel_create(proxy_tunnel_rec **ptunnel,
...
+        apr_uint32_t val = apr_atomic_add32((volatile apr_uint32_t *)&(worker->s->busy), -1);
+        if (val == 0 || val > INT_MAX)
+            apr_atomic_inc32((volatile apr_uint32_t *)&(worker->s->busy));
+    } else {
+        apr_uint64_t val = apr_atomic_add64((volatile apr_uint64_t *)&(worker->s->busy), -1);
+        if (val == 0 || val > INT_MAX)
+            apr_atomic_inc64((volatile apr_uint64_t *)&(worker->s->busy));
+    }
+    return APR_SUCCESS;
+}
```
ylavic, line 5029: Rolling back a "dec" with "inc" is racy, a reader could still see worker->s->busy < 0 in the meantime. This probably needs a compare and swap (CAS) loop like:
```
void decrement_busy_count(proxy_worker *worker)
{
    apr_size_t val;

#if APR_SIZEOF_VOIDP == 4
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(apr_uint32_t));
    val = apr_atomic_read32(&worker->s->busy);
    while (val > 0) {
        apr_size_t old = val;
        val = apr_atomic_cas32(&worker->s->busy, val - 1, old);
        if (val == old) {
            break;
        }
    }
#elif APR_VERSION_AT_LEAST(1,7,4) /* APR 64bit atomics not safe before 1.7.4 */
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(apr_uint64_t));
    val = apr_atomic_read64(&worker->s->busy);
    while (val > 0) {
        apr_size_t old = val;
        val = apr_atomic_cas64(&worker->s->busy, val - 1, old);
        if (val == old) {
            break;
        }
    }
#else /* Use atomics for (64bit) pointers */
    void *volatile *busy_p = (void *)&worker->s->busy;
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(void*));
    AP_DEBUG_ASSERT((apr_uintptr_t)busy_p % sizeof(void*) == 0);
    val = (apr_uintptr_t)apr_atomic_casptr((void *)busy_p, NULL, NULL);
    while (val > 0) {
        apr_size_t old = val;
        val = (apr_uintptr_t)apr_atomic_casptr((void *)busy_p,
                                               (void *)(apr_uintptr_t)(val - 1),
                                               (void *)(apr_uintptr_t)old);
        if (val == old) {
            break;
        }
    }
#endif
}
```
Note that APR 64bit atomics are available since 1.7.0 only, and really working on all platforms since 1.7.4, so we need something for earlier versions supported by http-2.4 still. The code above uses apr_atomic_casptr() which is always available and should work because I think we can assume sizeof(size_t) == sizeof(void*) and 8-bytes alignment of 64bit integral/pointer types on all 64bit platforms..

modules/proxy/proxy_util.c
```diff
@@ -4984,6 +4985,52 @@ PROXY_DECLARE(apr_status_t) ap_proxy_tunnel_create(proxy_tunnel_rec **ptunnel,
...
+
+PROXY_DECLARE(void) increment_busy_count(void *worker_)
+{
+    proxy_worker *worker = worker_;
+    if (sizeof(worker->s->busy) == 4) {
+       apr_atomic_inc32((volatile apr_uint32_t *)&(worker->s->busy));
+    } else {
+        apr_atomic_inc64((volatile apr_uint64_t *)&(worker->s->busy));
+    }
+}
```
ylavic, line 5069: Likewise:
```
void increment_busy_count(proxy_worker *worker)
{
    apr_size_t val;

#if APR_SIZEOF_VOIDP == 4
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(apr_uint32_t));
    val = apr_atomic_read32(&worker->s->busy);
    while (val < APR_INT32_MAX) {
        apr_size_t old = val;
        val = apr_atomic_cas32(&worker->s->busy, val + 1, old);
        if (val == old) {
            break;
        }
    }
#elif APR_VERSION_AT_LEAST(1,7,4) /* APR 64bit atomics not safe before 1.7.4 */
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(apr_uint64_t));
    val = apr_atomic_read64(&worker->s->busy);
    while (val < APR_INT32_MAX) {
        apr_size_t old = val;
        val = apr_atomic_cas64(&worker->s->busy, val + 1, old);
        if (val == old) {
            break;
        }
    }
#else /* Use atomics for (64bit) pointers */
    void *volatile *busy_p = (void *)&worker->s->busy;
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(void*));
    AP_DEBUG_ASSERT((apr_uintptr_t)busy_p % sizeof(void*) == 0);
    val = (apr_uintptr_t)apr_atomic_casptr((void *)busy_p, NULL, NULL);
    while (val < APR_INT32_MAX) {
        apr_size_t old = val;
        val = (apr_uintptr_t)apr_atomic_casptr((void *)busy_p,
                                               (void *)(apr_uintptr_t)(val + 1),
                                               (void *)(apr_uintptr_t)old);
        if (val == old) {
            break;
        }
    }
#endif
}
```

modules/proxy/proxy_util.c
```diff
@@ -4984,6 +4985,52 @@ PROXY_DECLARE(apr_status_t) ap_proxy_tunnel_create(proxy_tunnel_rec **ptunnel,
...
+
+PROXY_DECLARE(apr_size_t) getbusy_count(void *worker_)
+{
+    proxy_worker *worker = worker_;
+    if (sizeof(worker->s->busy) == 4) {
+       return(apr_atomic_read32((volatile apr_uint32_t *)&(worker->s->busy)));
+    } else {
+        return(apr_atomic_read64((volatile apr_uint64_t *)&(worker->s->busy)));
+    }
+}
```
ylavic, line 5088: And:
```
apr_size_t get_busy_count(proxy_worker *worker)
{
    apr_size_t val;

#if APR_SIZEOF_VOIDP == 4
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(apr_uint32_t));
    val = apr_atomic_read32(&worker->s->busy);
#elif APR_VERSION_AT_LEAST(1,7,4) /* APR 64bit atomics not safe before 1.7.4 */
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(apr_uint64_t));
    val = apr_atomic_read64(&worker->s->busy);
#else /* Use atomics for (64bit) pointers */
    void *volatile *busy_p = (void *)&worker->s->busy;
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(void*));
    AP_DEBUG_ASSERT((apr_uintptr_t)busy_p % sizeof(void*) == 0);
    val = (apr_uintptr_t)apr_atomic_casptr((void *)busy_p, NULL, NULL);
#endif

    return val;
}
```

modules/proxy/proxy_util.c
```diff
@@ -4984,6 +4985,52 @@ PROXY_DECLARE(apr_status_t) ap_proxy_tunnel_create(proxy_tunnel_rec **ptunnel,
...
+
+PROXY_DECLARE(void) setbusy_count(void *worker_, apr_size_t busy)
+{
+    proxy_worker *worker = worker_;
+    if (sizeof(worker->s->busy) == 4) {
+        apr_atomic_set32((volatile apr_uint32_t *)&worker->s->busy, (apr_uint32_t) busy); 
+    } else {
+        apr_atomic_set64((volatile apr_uint64_t *)&worker->s->busy, (apr_uint64_t) busy);
+    }
+}
```
ylavic, line 5104: And:
```
void set_busy_count(proxy_worker *worker, apr_size_t to)
{
#if APR_SIZEOF_VOIDP == 4
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(apr_uint32_t));
    apr_atomic_set32(&worker->s->busy, to);
#elif APR_VERSION_AT_LEAST(1,7,4) /* APR 64bit atomics not safe before 1.7.4 */
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(apr_uint64_t));
    apr_atomic_set64(&worker->s->busy, to);
#else /* Use atomics for (64bit) pointers */
    void *volatile *busy_p = (void *)&worker->s->busy;
    AP_DEBUG_ASSERT(sizeof(apr_size_t) == sizeof(void*));
    AP_DEBUG_ASSERT((apr_uintptr_t)busy_p % sizeof(void*) == 0);
    apr_atomic_xchgptr((void *)busy_p, (void *)(apr_uintptr_t)to);
#endif
}
```

modules/proxy/mod_proxy.h
jfclere: APR says: 
apr_uint32_t apr_atomic_read32 (volatile apr_uint32_t * mem  )
that is why I used volatile...
I know I have to bump   MMN before merging.

modules/proxy/proxy_util.c
jfclere: Do we need something specific for platforms that don't have atomics?

modules/proxy/mod_proxy.h
ylavic: Since the function parameter is declared volatile, any passed in memory (whether or not declared volatile itself) will be considered volatile at the function call and within the function itself (i.e. the volatile qualifier is additive, not part of the type checking). But volatility is not enough (per se) to provide atomicity on all platforms, that's why we can't use worker->s->busy directly and need the *_busy_count() helpers to read/change it. So we'll never use the volatile qualifier on proxy_worker_shared::busy anyway if it existed, IOW if the **accesses** to worker->s->busy are all made volatile then the memory itself needs not be declared volatile.
And I think that adding the volatile qualifier to some public struct member is an API change which requires an MMN **major** bump (some third party code using it might not compile anymore because the compiler could then warn about existing code "removing" the volatile qualification somehow), meaning it's not backportable to 2.4.x. That's why I'm trying to avoid it..

modules/proxy/proxy_util.c
ylavic: The apr_atomic_*() functions will use a mutex for platforms that don't have atomics, it needs not be handled by httpd.

modules/proxy/mod_proxy.h
jfclere: fixed

modules/proxy/mod_proxy.h
jfclere: fixed

modules/proxy/proxy_util.c
jfclere: thanks fixed.

modules/proxy/proxy_util.c
jfclere: thanks fixed.

modules/proxy/proxy_util.c
jfclere: thanks fixed.

whoami-anoint (approved):

PR #384, opened 2023-09-07
acinclude.m4: test for OPENSSL_NO_ENGINE

orbea:
With `>= LibreSSL 3.8.1` the engine code has been removed and `OPENSSL_NO_ENGINE` has been defined, testing for this define during configure will prevent implicit function declarations when using an ssl implementation with the engine code disabled or removed.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -607,11 +607,23 @@ AC_DEFUN([APACHE_CHECK_OPENSSL],[
 
       dnl Run library and function checks
       liberrors=""
+      AC_TRY_COMPILE([#include <openssl/opensslconf.h>],[
+#if defined(OPENSSL_NO_ENGINE)
+#error "Engine support disabled in <openssl/opensslconf.h>"
+#endif],
+      [AC_MSG_RESULT(OK)
+       ac_cv_openssl_engine=yes],
+      [AC_MSG_RESULT(FAILED)])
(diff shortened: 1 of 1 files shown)


notroj:
Can you test whether this works with current trunk? I think this was likely fixed in 1e06568a28e02b477eb660cc18a42420360103cf

orbea:
@notroj I tested the current trunk (https://github.com/apache/httpd/commit/c9028584a89a761e2804b596ce2c9f961f5b6fb8) which worked, then I tested that commit (https://github.com/apache/httpd/commit/1e06568a28e02b477eb660cc18a42420360103cf) which failed....

I then checked the `2.4.58` tag which matches the version in Gentoo, this also fails to build. Lastly I bisected between the trunk and the 1e06568a28e02b477eb660cc18a42420360103cf commit and found the commit that fixed this was 922d6bbc423f1684c002031dabc98c7ffabc8e08. This is a similar fix to what I was doing here so I'll close this now.

Thanks for the reply!

For reference my build command was:
```
export CFLAGS="-O2 -pipe -Werror=implicit-function-declaration"
./buildconf --with-apr=../apr-1.7.4 --with-apr-util=../apr-util-1.6.3 && ./configure && make
```

notroj:
Thanks a lot for testing it out.

PR #389, opened 2023-10-30
Merge r1881790, r1904513, r1909400, r1909401, r1909402, r1909451, r1912459 from trunk

rpluem:
Merge r1881790, r1904513, r1909400, r1909401, r1909402, r1909451, r1912459 from trunk


rpluem:
Missed r1365001

PR #393, opened 2023-11-17
mod_xml2enc: remove dependency on xmlstring header

ttachi:
Include of libxml/xmlstring.h was removed from libxml/encoding.h in libxml2 v2.12.0
https://github.com/GNOME/libxml2/commit/7909ff08e2e755e8e7c99c3e12ad7254ef728a1f

### DIFF
--- a/modules/filters/mod_xml2enc.c
+++ b/modules/filters/mod_xml2enc.c
@@ -209,7 +209,7 @@ static void sniff_encoding(request_rec* r, xml2ctx* ctx)
   
     /* to sniff, first we look for BOM */
     if (ctx->xml2enc == XML_CHAR_ENCODING_NONE) {
-        ctx->xml2enc = xmlDetectCharEncoding((const xmlChar*)ctx->buf,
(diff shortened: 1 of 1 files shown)


covener:
I guess both casts are ugly, but should we be including libxml/parser.h here?

ttachi:
libxml/encoding.h is included in mod_xml2enc.c.
https://github.com/apache/httpd/blob/trunk/modules/filters/mod_xml2enc.c#L38

actually, libxml/encoding.h has no dependencies on libxml/xmlstring.h.
https://github.com/GNOME/libxml2/blob/v2.12.0/include/libxml/encoding.h

instead of adding more headers, eliminate reference to xmlChar defined in libxml/xmlstring.h to reduce unnecessary dependencies.
https://github.com/apache/httpd/blob/trunk/modules/filters/mod_xml2enc.c#L212

covener:
Thanks, I see now, the type has never been in the signature of xmlDetectCharEncoding

PR #396, opened 2023-11-28
Implement [ap_]atomics for [u]int/[u]long/size_t and use them for (some) proxy_worker_shared fields

ylavic:
Follow up to https://lists.apache.org/thread/np1xzvq6yq5onmtqqovyx0doncowvp3f

This is a proposal/POC to implement [ap_]atomics on system dependent signed/unsigned types, using the uint32/uint66/ptr atomics from the APR as primitives.

This currently implements:
- ap_atomic_{int,long,size}_get()
- ap_atomic_{int,long,size}_set()
- ap_atomic_{int,long,size}_xchg()
- ap_atomic_{int,long,size}_cas()
- ap_atomic_{int,long,size}_add()
- ap_atomic_{int,long,size}_add_sat()
- ap_atomic_{int,long,size}_sub()
- ap_atomic_{int,long,size}_sub_sat()

The add/sub functions wrap on under/overflow (safely for signed types), while the add_sat/sub_sat ones saturate instead (e.g. at INT_MIN/INT_MAX for ints, at 0/ULONG_MAX for unsigned longs, ...).

This allows to replace/remove the `ap_proxy_{get,set,dec,inc}_busy_count()` specialized functions by using the new `ap_atomic_size_*()` helpers directly on worker->s->busy, and likewise to use the the new `ap_atomic_int_*()` directly on worker->s->lbstatus to switch to atomic ops there too.

TODO: dox, MMNs..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -732,6 +732,7 @@ SET(LIBHTTPD_SOURCES
   server/ssl.c
   server/scoreboard.c
   server/util.c
+  server/util_atomics.c
   server/util_cfgtree.c
   server/util_cookies.c
   server/util_debug.c
--- a/NWGNUmakefile
+++ b/NWGNUmakefile
@@ -268,6 +268,7 @@ FILES_nlm_objs = \
 	$(OBJDIR)/request.o \
 	$(OBJDIR)/scoreboard.o \
 	$(OBJDIR)/util.o \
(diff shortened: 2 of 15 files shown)


jfclere:
Oops I caused the conflicts.
Note I have created https://github.com/apache/httpd/pull/421 that works around a problem when building with windows.
I think that is due to the changes in httpd.h.

PR #397, opened 2023-12-04
mod_ssl: Add support for loading keys from OpenSSL 3.x providers via STORE

ifranzki:
The mod_ssl module has support for loading keys and certificates from OpenSSL engines via PKCS#11 URIs at SSLCertificateFile and SSLCertificateKeyFile, e.g. using the PKCS#11 engine part of libp11 (https://github.com/OpenSC/libp11).

This works fine, but with OpenSSL 3.0 engines got deprecated, and a new provider concept is used. OpenSSL 1.1.1 is no longer supported by the OpenSSL organization (https://www.openssl.org/blog/blog/2023/09/11/eol-111/), and newer distributions all have OpenSSL 3.x included. Currently, engines do still work, bit since they are deprecated, they will at some point in time no longer be working.

With OpenSSL 3.x providers one can implements loading of keys and certificates by implementing a STORE method. With this, keys and certificates can be loaded for example from PKCS#11 modules via PKCS#11 URIs, just like it was possible with an PKCS#11 engine.

This commit contains code changes required to support loading the server private key and certificates from a PKCS#11 provider using OpenSSL STORE providers.

The usage is very similar to how it was with engines. You can specify a PKCS#11 URI with SSLCertificateFile and SSLCertificateKeyFile, exactly how it is with engines. The only difference is that you must specify 'SSLCryptoDevice provider' as crypto device, instead of specifying the engine name.

That way, the code continues to support working with engines. So SSLCryptoDevice accepts either 'builtin' or an engine name as before, but now also 'provider' to enable the OpenSSL provider STORE API. Instead of choosing this approach, we could just replace the engine support by the provider support, but this might break existing installations that are still using engines.

The provider(s) to be used with httpd must be configured via the OpenSSL config file in the provider section. Most providers need additional, provider specific settings that can only be supplied via the OpenSSL config file.

If one does not like to configure the providers globally, one can have a separate OpenSSL config file and use environment variable OPENSSL_CONF to specify the config file to use. That way one can have an OpenSSL config file just for httpd.

Currently there exist 2 PKCS#11 provider projects:
- https://github.com/latchset/pkcs11-provider
- https://github.com/opencryptoki/openssl-pkcs11-sign-provider

Both do support loading keys via PKCS#11 URI via their STORE support, but the code below is not limited to just those two.
Any provider that supports a STORE implementation for URIs with the 'pkcs11' scheme can be used.

### DIFF
--- a/modules/ssl/ssl_engine_config.c
+++ b/modules/ssl/ssl_engine_config.c
@@ -689,11 +689,6 @@ const char *ssl_cmd_SSLCryptoDevice(cmd_parms *cmd,
     if (strcEQ(arg, "builtin")) {
         mc->szCryptoDevice = NULL;
     }
-#if MODSSL_HAVE_OPENSSL_STORE
-    else if (strcEQ(arg, "provider")) {
-        mc->szCryptoDevice = arg;
-    }
-#endif
 #if MODSSL_HAVE_ENGINE_API
     else if ((e = ENGINE_by_id(arg))) {
         mc->szCryptoDevice = arg;
@@ -702,11 +697,7 @@ const char *ssl_cmd_SSLCryptoDevice(cmd_parms *cmd,
 #endif
     else {
         err = "SSLCryptoDevice: Invalid argument; must be one of: "
-              "'builtin' (none)"
-#if MODSSL_HAVE_OPENSSL_STORE
-              ", 'provider' (use OpenSSL >= 3.0 provider STORE)"
-#endif
-              ;
+              "'builtin' (none)";
 #if MODSSL_HAVE_ENGINE_API
         e = ENGINE_get_first();
         while (e) {
--- a/modules/ssl/ssl_engine_config.c
+++ b/modules/ssl/ssl_engine_config.c
@@ -689,6 +689,11 @@ const char *ssl_cmd_SSLCryptoDevice(cmd_parms *cmd,
     if (strcEQ(arg, "builtin")) {
         mc->szCryptoDevice = NULL;
     }
+#if MODSSL_HAVE_OPENSSL_STORE
+    else if (strcEQ(arg, "provider")) {
+        mc->szCryptoDevice = arg;
+    }
+#endif
 #if MODSSL_HAVE_ENGINE_API
     else if ((e = ENGINE_by_id(arg))) {
         mc->szCryptoDevice = arg;
@@ -697,7 +702,11 @@ const char *ssl_cmd_SSLCryptoDevice(cmd_parms *cmd,
 #endif
     else {
         err = "SSLCryptoDevice: Invalid argument; must be one of: "
-              "'builtin' (none)";
+              "'builtin' (none)"
+#if MODSSL_HAVE_OPENSSL_STORE
+              ", 'provider' (use OpenSSL >= 3.0 provider STORE)"
+#endif
+              ;
 #if MODSSL_HAVE_ENGINE_API
         e = ENGINE_get_first();
         while (e) {
--- a/modules/ssl/ssl_engine_init.c
+++ b/modules/ssl/ssl_engine_init.c
@@ -506,7 +506,11 @@ apr_status_t ssl_init_Engine(server_rec *s, apr_pool_t *p)
     SSLModConfigRec *mc = myModConfig(s);
     ENGINE *e;
 
+#if MODSSL_HAVE_OPENSSL_STORE
+    if (mc->szCryptoDevice && !strcEQ(mc->szCryptoDevice, "provider")) {
+#else
     if (mc->szCryptoDevice) {
+#endif
         if (!(e = ENGINE_by_id(mc->szCryptoDevice))) {
             ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(01888)
                          "Init: Failed to load Crypto Device API `%s'",
@@ -1476,8 +1480,8 @@ static apr_status_t ssl_init_server_certs(server_rec *s,
             if (cert) {
                 if (SSL_CTX_use_certificate(mctx->ssl_ctx, cert) < 1) {
                     ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10137)
-                                 "Failed to configure engine certificate %s, check %s",
-                                 key_id, certfile);
+                                 "Failed to configure certificate %s from %s, check %s",
+                                 key_id, mc->szCryptoDevice, certfile);
                     ssl_log_ssl_error(SSLLOG_MARK, APLOG_EMERG, s);
                     return APR_EGENERAL;
                 }
@@ -1488,8 +1492,8 @@ static apr_status_t ssl_init_server_certs(server_rec *s,
             
             if (SSL_CTX_use_PrivateKey(mctx->ssl_ctx, pkey) < 1) {
                 ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10130)
-                             "Failed to configure private key %s from engine",
-                             keyfile);
+                             "Failed to configure private key %s from %s",
+                             keyfile, mc->szCryptoDevice);
                 ssl_log_ssl_error(SSLLOG_MARK, APLOG_EMERG, s);
                 return APR_EGENERAL;
             }
--- a/modules/ssl/ssl_engine_init.c
+++ b/modules/ssl/ssl_engine_init.c
@@ -506,11 +506,7 @@ apr_status_t ssl_init_Engine(server_rec *s, apr_pool_t *p)
     SSLModConfigRec *mc = myModConfig(s);
     ENGINE *e;
 
-#if MODSSL_HAVE_OPENSSL_STORE
-    if (mc->szCryptoDevice && !strcEQ(mc->szCryptoDevice, "provider")) {
-#else
     if (mc->szCryptoDevice) {
-#endif
         if (!(e = ENGINE_by_id(mc->szCryptoDevice))) {
             ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(01888)
                          "Init: Failed to load Crypto Device API `%s'",
@@ -1481,7 +1477,9 @@ static apr_status_t ssl_init_server_certs(server_rec *s,
                 if (SSL_CTX_use_certificate(mctx->ssl_ctx, cert) < 1) {
                     ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10137)
                                  "Failed to configure certificate %s from %s, check %s",
-                                 key_id, mc->szCryptoDevice, certfile);
+                                 key_id, mc->szCryptoDevice ?
+                                             mc->szCryptoDevice : "provider",
+                                 certfile);
                     ssl_log_ssl_error(SSLLOG_MARK, APLOG_EMERG, s);
                     return APR_EGENERAL;
                 }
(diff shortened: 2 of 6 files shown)


ifranzki:
Rebased to latest trunc.

notroj:
Very nice, thanks. Could you factor out the while() loop which does the extraction from the STORE based on a type argument, it looks like the same both times?

ifranzki:
> Very nice, thanks. Could you factor out the while() loop which does the extraction from the STORE based on a type argument, it looks like the same both times?

Its the same except for `OSSL_STORE_INFO_get1_PKEY()` vs `OSSL_STORE_INFO_get1_CERT()`. 
Still I could use a common routine, that would on input get the type to retrieve, and then return the OSSL_STORE_INFO to the caller.

modules/ssl/ssl_engine_config.c
```diff
@@ -689,6 +689,11 @@ const char *ssl_cmd_SSLCryptoDevice(cmd_parms *cmd,
     if (strcEQ(arg, "builtin")) {
         mc->szCryptoDevice = NULL;
     }
+#if MODSSL_USE_OPENSSL_STORE
+    else if (strcEQ(arg, "provider")) {
+        mc->szCryptoDevice = arg;
```
ylavic: I'm not familiar with the new provider API, if there are other ways to use a provider than through the OpenSSL config file and OSSL_STORE (like with ENGINE_by_id()?) maybe we should name the crypto device more specifically here, e.g. "provider-pkcs11" or "provider-store" or something?

modules/ssl/ssl_engine_config.c
```diff
@@ -697,7 +702,11 @@ const char *ssl_cmd_SSLCryptoDevice(cmd_parms *cmd,
 #endif
     else {
         err = "SSLCryptoDevice: Invalid argument; must be one of: "
+#if MODSSL_USE_OPENSSL_STORE
+              "'builtin' (none), 'provider' (use OpenSSL >= 3.0 provider STORE)";
+#else
               "'builtin' (none)";
+#endif
```
ylavic: Maybe:
```
        err = "SSLCryptoDevice: Invalid argument; must be one of: "
              "'builtin' (none)"
#if MODSSL_USE_OPENSSL_STORE
              ", 'provider' (use OpenSSL >= 3.0 provider STORE)"
#endif
              ;
```
for better "evolutivity"?

modules/ssl/ssl_engine_init.c
```diff
@@ -1476,7 +1480,11 @@ static apr_status_t ssl_init_server_certs(server_rec *s,
             if (cert) {
                 if (SSL_CTX_use_certificate(mctx->ssl_ctx, cert) < 1) {
                     ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10137)
+#if MODSSL_USE_OPENSSL_STORE
+                                 "Failed to configure engine/provider certificate %s, check %s",
+#else
                                  "Failed to configure engine certificate %s, check %s",
+#endif
                                  key_id, certfile);
```
ylavic: Maybe:
```
    SSLModConfigRec *mc = myModConfig(s);
    ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10137)
                 "Failed to configure certificate %s from %s, check %s",
                 key_id, mc->szCryptoDevice, certfile);
```

modules/ssl/ssl_engine_init.c
```diff
@@ -1488,7 +1496,11 @@ static apr_status_t ssl_init_server_certs(server_rec *s,
             
             if (SSL_CTX_use_PrivateKey(mctx->ssl_ctx, pkey) < 1) {
                 ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10130)
+#if MODSSL_USE_OPENSSL_STORE
+                             "Failed to configure private key %s from engine/provider",
+#else
                              "Failed to configure private key %s from engine",
+#endif
                              keyfile);
```
ylavic: Likewise, maybe:
```
    SSLModConfigRec *mc = myModConfig(s);
    ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10130)
                 "Failed to configure private key %s from %s",
                 keyfile, mc->szCryptoDevice);
```

modules/ssl/ssl_engine_pphrase.c
```diff
@@ -873,6 +877,133 @@ apr_status_t modssl_load_engine_keypair(server_rec *s, apr_pool_t *p,
...
+    ppcb.s = s;
+    ppcb.p = p;
+    ppcb.bPassPhraseDialogOnce = TRUE;
+    ppcb.key_id = vhostid;
+    ppcb.pkey_file = keyid;
+
+    *privkey = NULL;
+    sctx = OSSL_STORE_open(keyid, ui_method, &ppcb, NULL, NULL);
+    if (!sctx) {
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10131)
```
ylavic: It's better to use an empty `APLOGNO()` than reusing an existing one (here and below), a script can take care of the empty ones when this PR is merged.

modules/ssl/ssl_engine_pphrase.c
```diff
@@ -873,6 +877,133 @@ apr_status_t modssl_load_engine_keypair(server_rec *s, apr_pool_t *p,
...
+        break;
+    }
+
+    OSSL_STORE_close(sctx);
+
+    if (!*privkey) {
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10131)
+                     "Init: OSSL_STORE_INFO_PKEY lookup failed for private key identifier `%s'",
+                     keyid);
+        OSSL_STORE_close(sctx);
```
ylavic: Double close it seems.

modules/ssl/ssl_engine_config.c
ifranzki: You don't enable a specific provider (as it is with engines), you just enable that it goes the Provider STORE path.
Other than with engines, you can't load a specific provider programmatically. Well you can, but most providers do need additional configuration options which you can't supply when loading it programmatically.

So whatever we choose should be generic enough to not be confused with naming a specific provider.

modules/ssl/ssl_engine_config.c
ifranzki: sure, will change.
I didn't like the lonely semicolon at the end :-)

modules/ssl/ssl_engine_pphrase.c
```diff
@@ -873,6 +877,133 @@ apr_status_t modssl_load_engine_keypair(server_rec *s, apr_pool_t *p,
...
+        return ssl_die(s);
+    }
+
+    if (certid) {
+        *pubkey = NULL;
```
ylavic: This could go with the early `*privkey = NULL;` above for the error cases and/or !certid. Maybe there should a cleanup code for `*privkey` and `*pubkey` if not NULL and an error happens at some point (`goto cleanup` in a common pattern in httpd), unless it's a fatal error and httpd is about to exit anyway.

modules/ssl/ssl_private.h
```diff
@@ -118,6 +118,13 @@
 #define MODSSL_HAVE_ENGINE_API 0
 #endif
 
+/* Use OpenSSL 3.x STORE for loading URI keys and certificates starting with
+ * OpenSSL 3.0
+ */
+#if OPENSSL_VERSION_NUMBER >= 0x30000000
+#define MODSSL_USE_OPENSSL_STORE 1
```
ylavic: Nit: I think it should be MODSSL_HAVE_OPENSSL_STORE, _USE_is usually for when we _HAVE_ it but it can disabled by other means.

modules/ssl/ssl_engine_init.c
ifranzki: yes, will change.

modules/ssl/ssl_engine_init.c
ifranzki: yes, will change.

modules/ssl/ssl_engine_pphrase.c
ifranzki: OK, will change

modules/ssl/ssl_engine_pphrase.c
ifranzki: Right. will remove.

modules/ssl/ssl_engine_pphrase.c
ifranzki: OK, will move op the `*pubkey = NULL;`
All error cases use `ssl_die()` so I guess a cleanup is not needed. 
It was that way for the engine case already, so I guess it really should die when it can't load the key/cert for any reason.

modules/ssl/ssl_private.h
ifranzki: OK, will change.

ylavic: Thanks, it's possibly missing an:
```
#else
#define MODSSL_HAVE_OPENSSL_STORE 0
```
too.

modules/ssl/ssl_engine_config.c
ylavic: All right, if it's the (most) common way to use an openssl provider I'm fine with the simple "provider" name.
EDIT: or even better if it can be left to "builtin" per your discussion with Joe :)

modules/ssl/ssl_private.h
ifranzki: right. will fix, too.

ifranzki:
I fixed the findings and force-pushed. Please take a look. 
I also factored out the while() loop which does the extraction from the STORE based on a type argument.

As a proposal, I changed to code to no longer require the `SSLCryptoDevice provider` special name. If `SSLCryptoDevice` is not specified or is `builtin`, and the key or certificate uses a PKCS#11 URI, then it automatically goes the STORE path. If `SSLCryptoDevice` does specify an engine, then it still goes the engine path. If no PKCS#11 URI is used, then it anyways goes the ordinal file path, because `modssl_load_engine_keypair()` is only called if an URI is used.
However, it allows to use OpenSSL providers just by specifying an URI for the key and/or certificate, without any further configuration required.

I have added this proposal as a fixup commit on top. Let me know if you like it. If so, I can squash the 2 commits together.

modules/ssl/ssl_engine_pphrase.c
```diff
@@ -873,6 +877,117 @@ apr_status_t modssl_load_engine_keypair(server_rec *s, apr_pool_t *p,
...
+    if (certid) {
+        info = modssl_load_store_uri(s, p, vhostid, certid, OSSL_STORE_INFO_CERT);
+        if (!info) {
+            ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO()
+                         "Init: OSSL_STORE_INFO_CERT lookup failed for certificate identifier `%s'",
+                         keyid);
+            return ssl_die(s);
+        }
+
+        *pubkey = OSSL_STORE_INFO_get1_CERT(info);
```
ylavic, line 963: OSSL_STORE_INFO_free(info); ?

ylavic (approved):
+1 (let's see if the ci starts with an approval..)

modules/ssl/ssl_engine_pphrase.c
ifranzki: Ah yes. Fixed and force pushed.

ifranzki:
> +1 (let's see if the ci starts with an approval..)

I guess some maintainer needs to approve the CI run....

notroj:
The permissions are still broken for this AFAIK. I've pushed a fork of this fork under my user, maybe that works.

notroj:
And by some magic the CI results are presented here too :)

LGTM. I will merge later unless there's any other feedback here. Don't bother to squash it since we have to translate it to SVN anyway + add the APLOGNOs.

ifranzki:
> Don't bother to squash it since we have to translate it to SVN anyway

OK, how about the commit message text and SVN? The text from the 1st commit still mentions the SSLCryptoDevice setting with 'provider' special name. You might need to adjust that to reflect the squashed state of the code.

notroj:
I wrote something a bit more concise for the commit message - hope this doesn't misinterpret anything you wrote! Big thanks for the contribution, this is awesome :+1:

soeda-san (commented):
.

PR #401, opened 2024-01-04
ARRAY_LEN: Introduce small helper macro

thomasmey:


notroj:
Thanks for the PR, I think this is a good idea though it doesn't need a whole new header. Maybe add the macro to `include/ap_config.h` instead, that is included everywhere?

PR #402, opened 2024-01-16
Check SSL_CTX_new() return value

StephenWall:
SSL_CTX_new() will return NULL if there was an error creating a new SSL context.

### DIFF
--- a/modules/ssl/ssl_engine_init.c
+++ b/modules/ssl/ssl_engine_init.c
@@ -738,6 +738,10 @@ static apr_status_t ssl_init_ctx_protocol(server_rec *s,
         TLS_server_method();  /* server */
 #endif
     ctx = SSL_CTX_new(method);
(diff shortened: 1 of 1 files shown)


notroj:
In what cases does it fail other than an out-of-memory condition?

StephenWall:
I saw it returning NULL while attempting to get FIPS enabled in OpenSSL system-wide.  I had some configuration/installation problems to resolve, but an error message from apache saying "SSL_CTX_new() failed" would have been much more useful than a segfault.
The man page simply says "The creation of a new SSL_CTX object failed. Check the error stack to find out the reason."  There are many places that could trigger a NULL return in the code. A lot are no doubt memory exhaustion, but there is some dependency on engines, and I saw a comment about races between threads, so it's hard to say what else might trigger it.

notroj:
Thanks, that's good to know.  Applying this now.

PR #410, opened 2024-02-07
2.4.x: mod_xml2enc media type fixes

notroj:
```Merge r1884505, r1915625 from trunk:
The Microsoft OOXML format uses xml packaged into a zip file, and has mimetypes like:

application/vnd.openxmlformats-officedocument.spreadsheetml.sheet

This mimetypes contains 'xml', but is unfortunately not an xml file.

xml2enc processes these files (in particular, when mod_proxy_html is used), typically resulting in them being corrupted as it seems to attempt to perform a ISO-8859-1 to UTF-8 conversion on them.

* modules/filters/mod_xml2enc.c (xml2enc_ffunc): Restrict test for XML types to matching "+xml".

Submitted by: Joseph Heenan <joseph.heenan fintechlabs.io>, jorton
PR: 64339
Github: closes #150
```

### DIFF
--- a/modules/filters/mod_xml2enc.c
+++ b/modules/filters/mod_xml2enc.c
@@ -323,7 +323,7 @@ static apr_status_t xml2enc_ffunc(ap_filter_t* f, apr_bucket_brigade* bb)
     apr_bucket* bstart;
     apr_size_t insz = 0;
     int pending_meta = 0;
-    char *ctype;
+    char *mtype;
     char *p;
 
     if (!ctx || !f->r->content_type) {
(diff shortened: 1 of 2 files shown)


modules/filters/mod_xml2enc.c
```diff
@@ -332,13 +332,17 @@ static apr_status_t xml2enc_ffunc(ap_filter_t* f, apr_bucket_brigade* bb)
...
-
-    /* only act if starts-with "text/" or contains "xml" */
-    if (strncmp(ctype, "text/", 5) && !strstr(ctype, "xml"))  {
+    /* Extract the media type, ignoring parameters in content-type. */
+    mtype = apr_pstrdup(f->r->pool, f->r->content_type);
+    if ((p = ap_strchr(mtype, ';')) != NULL) *p = '\0';
+    ap_str_tolower(mtype);
+
+    /* Accept text/ types, plus any XML media type per RFC 7303. */
+    if (!(strncmp(mtype, "text/", 5) == 0
```
covener, line 341: @notroj Is this correct? It looks like it mixes two flavors of negating strncmp result and negates twice

notroj: I struggled to make this readable, hoped it was easier after De Morgan-ing it, but it's still somehow not...

intuition: if you look just at the first clause it turns `if strncmp() != 0` into `if !(strncmp() == 0)` which is equivalent

covener: I see, I just realized the block inside is removing the filter, not taking action on the body.

PR #411, opened 2024-02-09
Don't reset the balancer when we create a new child

jfclere:
and don't change shared memory the load balancer is not using.


jfclere:
merged by r1916004.

PR #412, opened 2024-02-20
Merge r1825124 from trunk

notroj:
```
Merge r1825124 from trunk:

Follow-up to r1825120:

* modules/ssl/ssl_engine_init.c (ssl_init_ca_cert_path): Since SSL_add_file_cert_subjects_to_stack() internally replaces the comparison callback with one equivalent to ssl_init_FindCAList_X509NameCmp, there's no point in using that here. (ssl_init_FindCAList_X509NameCmp): Removed.

```


notroj:
Merged in 534de0a39d32b2403efdfb0d1c5f67be70b2858d

PR #419, opened 2024-03-05
Arrange error page

jfclere:
When the error is broken (like missing error file) make sure the
error page displays the right method and the correct Additionally message.


jfclere:
merged

PR #421, opened 2024-03-06
A little mor for https://github.com/apache/httpd/pull/396

jfclere:
Change ERROR into DAVERROR, because Windows has ERROR in assert.h...


jfclere:
Just that windoze doesn't like the changes in httpd.h...

PR #432, opened 2024-04-03
Use AP_DECLARE_MODULE in Developing modules and mod_so guides

jajik:
The macro calls `AP_MODULE_DECLARE_DATA` macro that is used currently, but in addition it sets the module name for logging. This should be compatible with versions 2.3.6 (almost 14 years old) and newer so I believe it should be preferred.


jajik:
(force push because I forgot to include the language mutations, sorry about that)

notroj:
This is great, thanks. FWIW there is no need to include language mutations in future PRs, these tend to get regenerated periodically anyway.

jajik:
Thank you for the merge. I'll keep that in mind.

PR #435, opened 2024-04-13
Modernization of index.html

Pandastrange:


bigio:
this file is part of tests, there is no need to improve the html page.

PR #437, opened 2024-04-18
Add git commands for apr checkout to INSTALL

jajik:
Since git/Github si getting more exposure and apr is mirrored there as well, it seems reasonable to add a git way of checking out apr.


cpu (approved):

jfclere (approved):

notroj:
Missed this one, good idea, thanks.

PR #441, opened 2024-05-01
fix the Japanese translation docs of the LimitRequestBody

vinyl-umbrella:
I fixed the Japanese docs of the LimitRequestBody.

Because it had not been updated, I assumed that the limit of request body was unlimited and wasted several hours.

I would like to correct this so that Japanese engineers do not make the same mistake.


notroj:
Thanks for the PR.

PR #447, opened 2024-05-23
mod_ssl: Fix reading custom DH parameters from CertificateFile with OpenSSL 3.X

o8vm:
This commit addresses an issue where Apache HTTPD with OpenSSL 3.0 or later is unable to load custom DH parameters specified in the CertificateFile configuration option. PEM_read_bio_Parameters() is unable to extract just the DH parameters and returns NULL when a CertificateFile contains signed certificates, intermediate certificates, and DH parameters all together in one file. A new approach is needed for OpenSSL 3.0. This patch implements a solution using OpenSSL's decoding framework to selectively load only the custom DH parameters from a composite CertificateFile.

Testing confirmed custom DH params can now be properly extracted from CertificateFile as intended.

### DIFF
--- a/modules/ssl/ssl_private.h
+++ b/modules/ssl/ssl_private.h
@@ -102,6 +102,7 @@
 #include <openssl/dh.h>
 #if OPENSSL_VERSION_NUMBER >= 0x30000000
 #include <openssl/core_names.h>
+#include <openssl/decoder.h>
 #endif
 
 /* Avoid tripping over an engine build installed globally and detected
--- a/modules/ssl/ssl_util_ssl.c
+++ b/modules/ssl/ssl_util_ssl.c
@@ -493,10 +493,20 @@ EVP_PKEY *modssl_dh_pkey_from_file(const char *file)
 {
     EVP_PKEY *pkey;
     BIO *bio;
+    OSSL_DECODER_CTX *dctx;
 
     if ((bio = BIO_new_file(file, "r")) == NULL)
         return NULL;
-    pkey = PEM_read_bio_Parameters(bio, NULL);
+    if ((dctx =
+         OSSL_DECODER_CTX_new_for_pkey(&pkey, "PEM", NULL, "DH",
+                                       EVP_PKEY_KEY_PARAMETERS, NULL,
+                                       NULL)) == NULL)
+        return NULL;
+    while (!OSSL_DECODER_from_bio(dctx, bio) || pkey == NULL) {
+        if (BIO_eof(bio))
+            break;
+    }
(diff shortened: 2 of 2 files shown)


o8vm:
https://bz.apache.org/bugzilla/show_bug.cgi?id=69045

notroj:
Thank you for sending in the PR but I think this is already addressed by dee1eb37d787d34cb37df7eab535240e1774293a on trunk, can you confirm that also fixes this issue for you?

o8vm:
Thank you for checking, but this issue is still not fixed in the latest HEAD of the trunk branch. I tested it, and the issue is also not addressed by the dee1eb3 commit, which seems to be effective only for OPENSSL_VERSION_NUMBER < 0x30000000L. My patch included in the PR specifically addresses this issue in OpenSSL 3.0 and later versions.

notroj:
dee1eb37d787d34cb37df7eab535240e1774293a definitely affects the OpenSSL 3.0+ code path as well as the pre-3.0 code path.

I am still not sure what your patch here fixes. Can you explain why using `PEM_read_bio_Parameters` is an inappropriate  way to load the DH parameters? Does it fail? Can you add logging on the error path, if so, to show the failure? Or is the call to `SSL_CTX_set0_tmp_dh_pkey` failing?

Testing with trunk and a custom DH parameter appended to the file referenced from `SSLCertificateFile`  now works as expected for me with OpenSSL 3.2.1.

```
t/logs/error_log:[Mon Jun 17 14:02:14.950767 2024] [ssl:debug] [pid 63509:tid 63509] ssl_engine_init.c(1611): AH02540: Custom DH parameters (4096 bits) for localhost:8553 loaded from ... /conf/ssl/ca/asf/certs/server.crt
```

notroj:
I would really like to fix this if there is an outstanding issue here. Can you confirm whether you see this issue in 2.4.61?

notroj:
Per https://bz.apache.org/bugzilla/show_bug.cgi?id=69045 closing since I believe this is resolved via dee1eb37d787d34cb37df7eab535240e1774293a

kmacm99:
In my test environment(httpd 2.4.62 and openssl 3.4.0), it can load custom DH parameters only when setting it at the beginning of CertificateFile and it can't when setting it at the end of CertificateFile.
I think it's better it can load at any position.
I'm not sure but this issue is similar to the following openssl fix and the suggested approach in this PR is also similar to that.

https://github.com/openssl/openssl/commit/5585d0da5f27971d39050c213a064ca6b3249e4a#diff-3da744bcd889cccc9c57bd5ec5629e71d70e6a275f6f5d55214dcebc18023b5dR600

PR #452, opened 2024-06-03
Fix missing initialization, follow up to r1918098

jajik:
One of the latest changes in https://github.com/apache/httpd/commit/d821182d76394cb38d5c80d447a66c1be23dc46c triggers a warning that causes compilation failure with `-Werror`. (It broke our CI and the CI here as well.)


notroj:
Thanks!

ylavic:
Oups, thanks for the report, and thanks Joe for the merge.

PR #462, opened 2024-07-17
Backport 2.4.x: mod_rewrite: Better question mark tracking to avoid UnsafeAllow3F

ylavic:
BZ 69197.

Track in do_expand() whether a '?' in the uri-path comes from a literal in the substitution string or from an expansion (variable, lookup, ...). In the former case it's safe to assume that it's the query-string separator but for the other case it's not (could be a decoded %3f from r->uri).

This allows to avoid [UnsafeAllow3F] for most cases.

git-svn-id: https://svn.apache.org/repos/asf/httpd/httpd/trunk@1919325 13f79535-47bb-0310-9956-ffa450edef68

### DIFF
--- a/changes-entries/pr69197.txt
+++ b/changes-entries/pr69197.txt
@@ -1,0 +1,2 @@
+  *) mod_rewrite: Better question mark tracking to avoid UnsafeAllow3F.
+     PR 69197. [Yann Ylavic, Eric Covener]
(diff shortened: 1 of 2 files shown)


ylavic:
Backported in [r1919545](https://github.com/apache/httpd/commit/0f6944358cf468d7d80bd63f33fbed7c4c0c6311)

PR #470, opened 2024-08-01
BZ-69203: backport to 2.4.x

ylavic:
- r1919620
- r1919621
- r1919623
- r1919628
- r1921237

### DIFF
--- a/modules/proxy/mod_proxy_fcgi.c
+++ b/modules/proxy/mod_proxy_fcgi.c
@@ -59,10 +59,13 @@ static int proxy_fcgi_canon(request_rec *r, char *url)
 {
     char *host, sport[7];
     const char *err;
-    char *path, *pc;
+    char *path;
     apr_port_t port, def_port;
     fcgi_req_config_t *rconf = NULL;
     const char *pathinfo_type = NULL;
+    fcgi_dirconf_t *dconf = ap_get_module_config(r->per_dir_config,
+                                                 &proxy_fcgi_module);
+    int from_handler;
 
     if (ap_cstr_casecmpn(url, "fcgi:", 5) == 0) {
         url += 5;
@@ -71,12 +74,11 @@ static int proxy_fcgi_canon(request_rec *r, char *url)
         return DECLINED;
     }
 
-    path = url;
     port = def_port = ap_proxy_port_of_scheme("fcgi");
 
     ap_log_rerror(APLOG_MARK, APLOG_TRACE1, 0, r,
                  "canonicalising URL %s", url);
-    err = ap_proxy_canon_netloc(r->pool, &path, NULL, NULL, &host, &port);
+    err = ap_proxy_canon_netloc(r->pool, &url, NULL, NULL, &host, &port);
     if (err) {
         ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, r, APLOGNO(01059)
                       "error parsing URL %s: %s", url, err);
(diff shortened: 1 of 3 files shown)


modules/proxy/mod_proxy_fcgi.c
```diff
@@ -106,16 +129,6 @@ static int proxy_fcgi_canon(request_rec *r, char *url)
             return HTTP_BAD_REQUEST;
         }
     }
-    /*
-     * If we have a raw control character or a ' ' in nocanon path,
-     * correct encoding was missed.
-     */
-    if (path == url && *ap_scan_vchar_obstext(path)) {
```
covener, line 113: have we intentionally lost failing on spaces here?  I guess the original intent was about splitting with CRLF?

modules/proxy/mod_proxy_fcgi.c
```diff
@@ -348,10 +360,11 @@ static apr_status_t send_environment(proxy_conn_rec *conn, request_rec *r,
...
     fcgi_dirconf_t *dconf = ap_get_module_config(r->per_dir_config, &proxy_fcgi_module);
 
-    if (rconf) {
-       if (rconf->need_dirwalk) {
-          ap_directory_walk(r);
-       }
+    if (rconf && rconf->need_dirwalk) {
+        char *saved_filename = r->filename;
+        r->filename = r->uri;
+        ap_directory_walk(r);
```
rpluem: See https://lists.apache.org/thread/4gw6tsp0xlpdf3l9gtz1jnr5wzycsfg8:

Why not using the result of the below Strip proxy: prefixes for `r->filename`? In case of a `Rewriterule` I guess `r->uri` and `r->filename` could be fundamentally different path wise.

modules/proxy/mod_proxy_fcgi.c
```diff
@@ -92,9 +95,29 @@ static int proxy_fcgi_canon(request_rec *r, char *url)
         host = apr_pstrcat(r->pool, "[", host, "]", NULL);
     }
 
-    if (apr_table_get(r->notes, "proxy-nocanon")
+    from_handler = apr_table_get(r->notes, "proxy-sethandler") != NULL;
+    if (from_handler
```
rpluem: See https://lists.apache.org/thread/rjh5t17512yxt44cogxwqzscgo4t2m09

Why do we need the `from_handler` variable?

modules/proxy/mod_proxy_fcgi.c
covener: Maybe it helps to say a little about this obscure option to try to calculate PATH_INFO. It's meant/doc'ed for ProxyPass/ProxyPassMatch configurations where there was no mapping to FS hence no split off for PATH_INFO.  

I don't think any of these proxy-fcgi-pathinfo  options are widely used.  I guess if r->filename had proxy: prefix stuff in it this was maybe useless since inception hence no users?

OTOH like you say, just peeling it all off as we will already do below would seem to give a better result assuming the user doesn't intend to keep some "original" PATH_INFO.

modules/proxy/mod_proxy_fcgi.c
ylavic: Yes, IIUC that's the issue reported in BZ-69203.
Before 2.4.60+ we didn't call proxy_fcgi_canon() for SetHandler URLs, so SCRIPT_FILENAME was never encoded. Now that we do, this PR applies "nocanon" for SetHandler URLs, and lets spaces through because supposedly they are not harmful in the SCRIPT_FILENAME env var (unlike in the uri-path for an HTTP request..).

modules/proxy/mod_proxy_fcgi.c
ylavic: This was taken from a larger patch in this thread [1] where there is another use of it.

[1] https://lists.apache.org/thread/vmz22tv2j8bz7vhw98hy3xwvqyp68cdc

modules/proxy/mod_proxy_fcgi.c
ylavic: > Why not using the result of the below Strip proxy: prefixes for `r->filename`? In case of a `Rewriterule` I guess `r->uri` and `r->filename` could be fundamentally different path wise.

Done in [1] too, but this patch is not ready yet and will probably need more discussions (including with fpm maintainers..). I can take this bit from there though, will do!

[1] https://lists.apache.org/thread/vmz22tv2j8bz7vhw98hy3xwvqyp68cdc

ylavic: Added 7441bb425008cd00d58047b818f131d9f301b217 for this, thanks!

modules/proxy/mod_proxy_fcgi.c
ylavic: Axed in b207f2026748fef281b95c6f2e33a958f2eac991

modules/proxy/mod_proxy_fcgi.c
ylavic: I'll create another PR for this because this has nothing to do with BZ-69203 afterall..

ylavic: See #489

bastien-roucaries:
What is the ETA for this pull request ?

modules/proxy/mod_proxy_fcgi.c
```diff
@@ -92,9 +94,30 @@ static int proxy_fcgi_canon(request_rec *r, char *url)
...
+        }
+        else {
+            while (!apr_iscntrl(*c))
+                c++;
+        }
+        if (*c) {
+            ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, r, APLOGNO(10414)
+                          "To be forwarded path contains control characters%s (%s)",
+                          FCGI_MAY_BE_FPM(dconf) ? " or '?'" : "", url);
+            return HTTP_FORBIDDEN;
```
bukka, line 117: Just a note that this changed from 404 to 403 (just tested the question mark case) compare to 2.4.59. But I think it's actually better error code now so should be fine.

bukka (commented):
I have done some testing with FPM (those are the test cases: https://github.com/wstool/wst-php-fpm/blob/bf390dd1dda196ea9339b4d3d26caf5fbf306226/spec/instances/httpd-proxy-fcgi-handler-basic.yaml ) and the only change required is the different status code for `?` in path which is actually for better as I mentioned in the comment. So this looks like a reasonable change to me. I haven't tested UDS yet though as I need to extend my testing framework a little bit for that.

bukka (approved):
I added [similar tests for UDS](https://github.com/wstool/wst-php-fpm/blob/563e061dbb7aa6fbe43979e78869b9b107c1b1b2/spec/instances/httpd-proxy-fcgi-handler-uds-basic.yaml) and they all pass too. It seems to work fine so I think this is good to go from the PHP-FPM PoV.

notroj:
Merged in 2aa446f5b08a10c37e952daf96d0c80d3460873a

PR #471, opened 2024-08-12
Update suexec.c

bobicloudvision:
Better message understanding of what exactly needs to be changed for suexec permissions.

Reason:
I compiled my own version of httpd with suexec, but when I get to use suexec it returns a permissions error.
But the error doesn't tell you exactly what you need to change

### DIFF
--- a/support/suexec.c
+++ b/support/suexec.c
@@ -633,6 +633,30 @@ int main(int argc, char *argv[])
         exit(119);
     }
 
+    /**
+    * Better understanding what exactly needs to be changed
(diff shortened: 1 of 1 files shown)


notroj:
It is kind of explicit in the existing message at https://github.com/apache/httpd/blob/trunk/support/suexec.c#L640 isn't it?

notroj:
No feedback, closing; if nothing else we tend to be extremely careful to never touch `suexec.c` unless it's really necessary.

PR #476, opened 2024-08-28
Multipath TCP patches from BZ 69292

notroj:


modules/proxy/mod_proxy_connect.c
```diff
@@ -244,6 +244,9 @@ static int proxy_connect_handler(request_rec *r, proxy_worker *worker,
      * For now we do nothing, ie we get DNS round robin.
      * XXX FIXME
      */
+
+    /* set the socket protocol to use */
+    conf->sock_proto = worker->s->sock_proto;
```
ylavic: This is not thread-safe, maybe ignoring MPTCP in mod_proxy_connect is acceptable for now because ap_proxy_connect_to_backend() has no worker configuration, and moving to ap_proxy_connect_backend() is more work..

Aperence: I don't mind dropping it in this case, from what I understood, this method is deprecated anyway so it should be replaced with ap_proxy_connect_backend in the future (which does support MPTCP) ?

ylavic: Yes, I think dropping the changes in mod_proxy_connect and ap_proxy_connect_to_backend() is simpler for now, if one wants to use mod_proxy_connect with MPTCP we can look at how moving to ap_proxy_connect_backend() there.

Aperence: Dropped those in the latest version available on [bugzilla](https://bz.apache.org/bugzilla/attachment.cgi?id=39854&action=edit)

notroj:
We normally bump MMN minor for ~~changes~~ additions to `proxy_worker_shared` right?

ylavic:
> We normally bump MMN minor for ~changes~ additions to `proxy_worker_shared` right?

Yes correct.

PR #477, opened 2024-08-30
mod_ssl: Add SSL_HANDSHAKE_RTT variable

csmutz:
OpenSSL 3.2 added the SSL_get_handshake_rtt function. This addition to mod_ssl exposes this data as a new variable, SSL_HANDSHAKE_RTT.

Notes about implementation:

SSL_HANDSHAKE_RTT is always set. It is set to 0 if handshake_rtt can not be retreived, either because a version of OpenSSL prior to 3.2 is used or because OpenSSL could not retrieve the handshake RTT (as in case of 0-RTT handshake). Given how the upstream function is called (after handshake is completed), there no meaningful feedback provided by upstream function so a second variable indicating errors does not seem warranted.

The handshake_rtt value is cast from uint64_t to long which should be adequate considering that it still supports handshakes up to 2147s in length on systems where long is 32-bit. I think proposed approach is best. If supporting handshakes that take over 2147s to complete is necessary, options include:
- divide value by 1000 and report value in units of ms (most people expect latency measurements in ms and precision below ms resolution is not necessary for most applications)
- use equivalent of apr_ltoa() that supports 64-bit int (please point me to example of this)
- value could be split into two variables (SSL_HANDSHAKE_RTT_S, SSL_HANDSHAKE_RTT_US). 

I've provided documentation addition in English only. I think it's preferable for others to provide French and Spanish translation (but I'll try if necessary).

I've tested this patch on Ubuntu 24 Apache/2.4.62,OpenSSL/3.3.1 and Ubuntu 22 Apache/2.4.52,OpenSSL/3.0.2 and both operated as expected.

### DIFF
--- a/docs/manual/mod/mod_ssl.xml
+++ b/docs/manual/mod/mod_ssl.xml
@@ -109,6 +109,7 @@ compatibility variables.</p>
 <tr><td><code>SSL_SRP_USER</code></td>                  <td>string</td>    <td>SRP username</td></tr>
 <tr><td><code>SSL_SRP_USERINFO</code></td>              <td>string</td>    <td>SRP user info</td></tr>
 <tr><td><code>SSL_TLS_SNI</code></td>                   <td>string</td>    <td>Contents of the SNI TLS extension (if supplied with ClientHello)</td></tr>
+<tr><td><code>SSL_HANDSHAKE_RTT</code></td>             <td>number</td>    <td>Round-trip time of TLS handshake in microseconds including endpoint processing (set to 0 if OpenSSL version prior to 3.2 or if round-trip time can not be determined)</td></tr>
 </table>
 
 <p><em>x509</em> specifies a component of an X.509 DN; one of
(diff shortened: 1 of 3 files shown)


notroj:
Thanks for the PR!

1) Using an apr_uint64_t and  `apr_psprintf(p, "%" APR_UINT64_T_FMT, rtt)` is the probably the only way to translate it without conversion to another type. No strong opinion on whether exporting ns or ms is better here but I'd say that coping with 35+ minute handshakes is not necessary! Better to export only a single variable anyway.

2) I think it would be better to return NULL for the error case and for the OpenSSL < 3.2 case, ssl_var_lookup() will then translate that to ""  - wrap the whole function body with the OpenSSL version test and have the `#else` branch as just `return NULL;`

3) for trunk the function should return `const char *` 

4) English-only documentation is definitely fine :)

csmutz:
1. Implemented apr_uint64_t as recommended--this is now consistent with upstream function which is best. Thanks for the help

2. Done. Implemented as recommended. Now set to null/empty string if handshake_rtt can't be retrieved for any reason. This code block is much cleaner now. Docs updated to reflect this change (0 vs. empty string).

3. Done.

4. OK

Thanks for quick response and help with improving code.

notroj:
Very nice, thanks a lot. Merging - if you want your full name for the credit in CHANGES please let me know.

PR #479, opened 2024-09-12
Check there's enough space to store macro result

marcstern:
https://bz.apache.org/bugzilla/show_bug.cgi?id=69258

### DIFF
--- a/modules/core/mod_macro.c
+++ b/modules/core/mod_macro.c
@@ -483,13 +483,13 @@ static const char *process_content(apr_pool_t * pool,
     for (i = 0; i < contents->nelts; i++) {
         const char *errmsg;
         /* copy the line and substitute macro parameters */
-	if (strlen(((char**)contents->elts)[i]) >= MAX_STRING_LEN) {
+	if (strlen(((char**)contents->elts)[i]) >= sizeof(line)) {
 	    return apr_psprintf(pool,
 		    "while processing line %d of macro '%s' (%s) %s",
 		    i + 1, macro->name, macro->location, "macro too long");
 	}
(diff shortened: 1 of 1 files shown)


modules/core/mod_macro.c
```diff
@@ -483,6 +483,11 @@ static const char *process_content(apr_pool_t * pool,
     for (i = 0; i < contents->nelts; i++) {
         const char *errmsg;
         /* copy the line and substitute macro parameters */
+	if (strlen(((char**)contents->elts)[i]) >= MAX_STRING_LEN) {
```
notroj: Shouldn't this be `(strlen(line) + strlen(...)) >= MAX_STRING_LEN` ?

notroj (approved):

modules/core/mod_macro.c
marcstern:
```
char line[MAX_STRING_LEN];
apr_cpystrn(line, ((char **) contents->elts)[i], MAX_STRING_LEN);
```
We copy `contents->elts[i]` onto `line`, so we need to be sure that `contents->elts[i]` is shorter than `sizeof(line)`, so `MAX_STRING_LEN`. Did I miss something?

Btw, we could use `sizeof(line)` instead of `MAX_STRING_LEN` everywhere to be cleaner ...

notroj: Sorry, yes I wasn't read that properly. Makes sense, thanks.

notroj: Also: yes using `sizeof line` would be better I agree.

notroj (approved):

modules/core/mod_macro.c
marcstern: done

PR #481, opened 2024-09-22
Add configuration snippet for post-quantum key exchange

nisbet-hubbard:


nisbet-hubbard:
Request review from @rbowen

bukka:
The post quantum provider is just for research purpose and definitely not for production use so it should be at least noted there.

nisbet-hubbard:
Thanks for your feedback!

I’m closing the PR because X25519MLKEM768 will become available in OpenSSL 3.5 and the method I described is getting obsolete.

PR #482, opened 2024-09-24
Document `htcacheclean`’s timestamp output

nisbet-hubbard:
This PR makes it clear that the datetime output of `htcacheclean -A` is a Unix timestamp in microseconds. This helps those trying to convert it to human-readable format using `date -d` or `gawk`’s `strftime()`, which take seconds.


nisbet-hubbard:
Request review from @rbowen

notroj:
Getting through some old PRs - looks good, thanks.

PR #483, opened 2024-09-25
mod_ssl: ClientHello variable collection

csmutz:
This patch implements collection of variables from the ClientHello which are later made available in the same manner as the other environment variables of mod_ssl. A new directive, SSLClientHelloVars, enables collection of the raw variables during the clienthello callback which are then formated and provided as environment variables for all the requests in that connection, subject to the standard StdEnvVars functionality. If SSLClientHelloVars is not enabled or if openssl prior to 1.1.1 is used, this option should have little impact--no clienthello information is collected. The environment variables are populated with null if openssl < 1.1.1 is used or SSLClientHelloVars is not set to on.

The ClientHello variables are provided as hex encoded data the same as the raw network protocol/what is returned from openssl.

This patch has been tested on ubuntu 24.10 apache 2.4.62/openssl 3.3.1 verifying correct operation of the config directive when not set (default to off), set to off, and set to on. The variables have been tested in the context of both environment variables for cgi scripts and CustomLog entries. The variables have been used to succesfully generate the correct ja3 and ja4 fingerprints for a small number of common clients (as validated by comparison to zeek implementations of ja3/ja4).

This patch complements my prior pull request, #477.

### DIFF
--- a/modules/ssl/ssl_engine_kernel.c
+++ b/modules/ssl/ssl_engine_kernel.c
@@ -1547,6 +1547,14 @@ static const char *const ssl_hook_Fixup_vars[] = {
     "SSL_SRP_USERINFO",
 #endif
     "SSL_HANDSHAKE_RTT",
+    "SSL_CLIENTHELLO_VERSION",
+    "SSL_CLIENTHELLO_CIPHERS",
+    "SSL_CLIENTHELLO_EXTENSIONS",
+    "SSL_CLIENTHELLO_GROUPS",
+    "SSL_CLIENTHELLO_EC_FORMATS",
+    "SSL_CLIENTHELLO_SIG_ALGOS",
+    "SSL_CLIENTHELLO_ALPN",
+    "SSL_CLIENTHELLO_VERSIONS",
     NULL
 };
 
@@ -2466,10 +2474,48 @@ int ssl_callback_ServerNameIndication(SSL *ssl, int *al, modssl_ctx_t *mctx)
 
 #if OPENSSL_VERSION_NUMBER >= 0x10101000L && !defined(LIBRESSL_VERSION_NUMBER)
 /*
+ * Copy data from clienthello for env vars use later
+ */
+static void copy_clienthello_vars(conn_rec *c, SSL *ssl)
+{
+    SSLConnRec *sslcon;
+    modssl_clienthello_vars *clienthello_vars;
+    const unsigned char *data;
+    int *ids;
+
+    sslcon = myConnConfig(c);
+
+    sslcon->clienthello_vars = apr_palloc(c->pool, sizeof(*clienthello_vars));
+    clienthello_vars = sslcon->clienthello_vars;
+
+    if (clienthello_vars) {
+        clienthello_vars->version = SSL_client_hello_get0_legacy_version(ssl);
+        clienthello_vars->ciphers_len = SSL_client_hello_get0_ciphers(ssl, &data);
+        clienthello_vars->ciphers_data = apr_pmemdup(c->pool, data, clienthello_vars->ciphers_len);
+        if (SSL_client_hello_get1_extensions_present(ssl, &ids, &clienthello_vars->extids_len) == 1) {
+            clienthello_vars->extids_data = apr_pmemdup(c->pool, ids, clienthello_vars->extids_len * sizeof(int));
+            OPENSSL_free(ids);
+        }
+        SSL_client_hello_get0_ext(ssl, 0x0a, &data, &clienthello_vars->ecgroups_len);
+        clienthello_vars->ecgroups_data = apr_pmemdup(c->pool, data, clienthello_vars->ecgroups_len);
+        SSL_client_hello_get0_ext(ssl, 0x0b, &data, &clienthello_vars->ecformats_len);
+        clienthello_vars->ecformats_data = apr_pmemdup(c->pool, data, clienthello_vars->ecformats_len);
+        SSL_client_hello_get0_ext(ssl, 0x0d, &data, &clienthello_vars->sigalgos_len);
+        clienthello_vars->sigalgos_data = apr_pmemdup(c->pool, data, clienthello_vars->sigalgos_len);
+        SSL_client_hello_get0_ext(ssl, 0x10, &data, &clienthello_vars->alpn_len);
+        clienthello_vars->alpn_data = apr_pmemdup(c->pool, data, clienthello_vars->alpn_len);
+        SSL_client_hello_get0_ext(ssl, 0x2b, &data, &clienthello_vars->versions_len);
+        clienthello_vars->versions_data = apr_pmemdup(c->pool, data, clienthello_vars->versions_len);
+    }
+}
+
+/*
  * This callback function is called when the ClientHello is received.
  */
 int ssl_callback_ClientHello(SSL *ssl, int *al, void *arg)
 {
+    server_rec *s;
+    SSLSrvConfigRec *sc;
     char *servername = NULL;
     conn_rec *c = (conn_rec *)SSL_get_app_data(ssl);
     const unsigned char *pos;
@@ -2520,6 +2566,12 @@ int ssl_callback_ClientHello(SSL *ssl, int *al, void *arg)
 
 give_up:
     init_vhost(c, ssl, servername);
+    
+    s = mySrvFromConn(c);
+    sc = mySrvConfig(s);
+    if (sc->clienthello_vars == TRUE)
+        copy_clienthello_vars(c, ssl);
+
     return SSL_CLIENT_HELLO_SUCCESS;
 }
 #endif /* OPENSSL_VERSION_NUMBER < 0x10101000L */
--- a/modules/ssl/ssl_engine_kernel.c
+++ b/modules/ssl/ssl_engine_kernel.c
@@ -2489,75 +2489,35 @@ static void copy_clienthello_vars(conn_rec *c, SSL *ssl)
     clienthello_vars = sslcon->clienthello_vars;
 
     clienthello_vars->version = SSL_client_hello_get0_legacy_version(ssl);
-
     clienthello_vars->ciphers_len = SSL_client_hello_get0_ciphers(ssl, &data);
     if (clienthello_vars->ciphers_len > 0) {
         clienthello_vars->ciphers_data = apr_pmemdup(c->pool, data, clienthello_vars->ciphers_len);
     }
-    else {
-        clienthello_vars->ciphers_data = NULL;
-    }
-
     if (SSL_client_hello_get1_extensions_present(ssl, &ids, &clienthello_vars->extids_len) == 1) {
         if (clienthello_vars->extids_len > 0)
             clienthello_vars->extids_data = apr_pmemdup(c->pool, ids, clienthello_vars->extids_len * sizeof(int));
         OPENSSL_free(ids);
     }
-    else {
-        clienthello_vars->extids_len = 0;
-    }
-    if (clienthello_vars->extids_len == 0)
-        clienthello_vars->extids_data = NULL;
-
     if (SSL_client_hello_get0_ext(ssl, TLSEXT_TYPE_supported_groups, &data, &clienthello_vars->ecgroups_len) == 1) {
         if (clienthello_vars->ecgroups_len > 0)
             clienthello_vars->ecgroups_data = apr_pmemdup(c->pool, data, clienthello_vars->ecgroups_len);
     }
-    else {
-        clienthello_vars->ecgroups_len = 0;
-    }
-    if (clienthello_vars->ecgroups_len == 0)
-        clienthello_vars->ecgroups_data = NULL;
-
     if (SSL_client_hello_get0_ext(ssl, TLSEXT_TYPE_ec_point_formats, &data, &clienthello_vars->ecformats_len) == 1) {
         if (clienthello_vars->ecformats_len > 0)
             clienthello_vars->ecformats_data = apr_pmemdup(c->pool, data, clienthello_vars->ecformats_len);
     }
-    else {
-        clienthello_vars->ecformats_len = 0;
-    }
-    if (clienthello_vars->ecformats_len == 0)
-        clienthello_vars->ecformats_data = NULL;
-
     if (SSL_client_hello_get0_ext(ssl, TLSEXT_TYPE_signature_algorithms, &data, &clienthello_vars->sigalgos_len) == 1) {
         if (clienthello_vars->sigalgos_len > 0)
             clienthello_vars->sigalgos_data = apr_pmemdup(c->pool, data, clienthello_vars->sigalgos_len);
     }
-    else {
-        clienthello_vars->sigalgos_len = 0;
-    }
-    if (clienthello_vars->sigalgos_len == 0)
-        clienthello_vars->sigalgos_data = NULL;
-
     if (SSL_client_hello_get0_ext(ssl, TLSEXT_TYPE_application_layer_protocol_negotiation, &data, &clienthello_vars->alpn_len) == 1) {
         if (clienthello_vars->alpn_len > 0)
             clienthello_vars->alpn_data = apr_pmemdup(c->pool, data, clienthello_vars->alpn_len);
     }
-    else {
-        clienthello_vars->alpn_len = 0;
-    }
-    if (clienthello_vars->alpn_len == 0)
-        clienthello_vars->alpn_data = NULL;
-
     if (SSL_client_hello_get0_ext(ssl, TLSEXT_TYPE_supported_versions, &data, &clienthello_vars->versions_len) == 1) {
         if (clienthello_vars->versions_len > 0)
             clienthello_vars->versions_data = apr_pmemdup(c->pool, data, clienthello_vars->versions_len);
     }
-    else {
-        clienthello_vars->versions_len = 0;
-    }
-    if (clienthello_vars->versions_len == 0)
-        clienthello_vars->versions_data = NULL;
 }
 
 /*
--- a/modules/ssl/ssl_engine_kernel.c
+++ b/modules/ssl/ssl_engine_kernel.c
@@ -2488,25 +2488,23 @@ static void copy_clienthello_vars(conn_rec *c, SSL *ssl)
     sslcon->clienthello_vars = apr_palloc(c->pool, sizeof(*clienthello_vars));
     clienthello_vars = sslcon->clienthello_vars;
 
-    if (clienthello_vars) {
-        clienthello_vars->version = SSL_client_hello_get0_legacy_version(ssl);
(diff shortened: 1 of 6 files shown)


notroj:
Very interesting indeed - thanks for another PR.

Higher-level comment: does this need to be a new directive? I think it would be better as another `SSLOptions` option, it's quite similar to `SSLOptions +ExportCertData` . Secondly, if the only point of exporting these is to create the JA3/JA4 fingerprints, would it be better to do that within `mod_ssl`? If the algorithm is simple enough and stable, it would not be inappropriate feature I guess.

modules/ssl/ssl_engine_vars.c
```diff
@@ -975,6 +979,56 @@ static const char *ssl_var_lookup_ssl_handshake_rtt(apr_pool_t *p, SSL *ssl)
...
+    int i;
+
+    clienthello_vars = sslconn->clienthello_vars;
+
+    if (!clienthello_vars)
+        return NULL; 
+
+    if (strEQ(var, "VERSION")) {
+        return apr_psprintf(p, "%04x", (uint16_t) clienthello_vars->version);
+    } else if (strEQ(var, "CIPHERS") && (clienthello_vars->ciphers_len > 0)) {
```
notroj: Code style nit: newline before `else if`

modules/ssl/ssl_engine_vars.c
```diff
@@ -975,6 +979,56 @@ static const char *ssl_var_lookup_ssl_handshake_rtt(apr_pool_t *p, SSL *ssl)
...
+        value = apr_palloc(p, clienthello_vars->ciphers_len * 2 + 1);
+        ap_bin2hex(clienthello_vars->ciphers_data, clienthello_vars->ciphers_len, value);
+        return value;
+    } else if (strEQ(var, "EXTENSIONS") && (clienthello_vars->extids_len > 0)) {
+        value = apr_palloc(p, clienthello_vars->extids_len * 4 + 1);
+        for (i = 0; i < clienthello_vars->extids_len; i++)
+        {
+            snprintf(value + i * 4, 5, "%04x", (uint16_t) clienthello_vars->extids_data[i]);
```
notroj: Should use `apr_snprintf` here

modules/ssl/ssl_engine_vars.c
```diff
@@ -975,6 +979,56 @@ static const char *ssl_var_lookup_ssl_handshake_rtt(apr_pool_t *p, SSL *ssl)
...
+    if (strEQ(var, "VERSION")) {
+        return apr_psprintf(p, "%04x", (uint16_t) clienthello_vars->version);
+    } else if (strEQ(var, "CIPHERS") && (clienthello_vars->ciphers_len > 0)) {
+        value = apr_palloc(p, clienthello_vars->ciphers_len * 2 + 1);
+        ap_bin2hex(clienthello_vars->ciphers_data, clienthello_vars->ciphers_len, value);
+        return value;
+    } else if (strEQ(var, "EXTENSIONS") && (clienthello_vars->extids_len > 0)) {
+        value = apr_palloc(p, clienthello_vars->extids_len * 4 + 1);
+        for (i = 0; i < clienthello_vars->extids_len; i++)
+        {
```
notroj: style nit: no newline before `{`

notroj (changes requested):
Also would be easier to review if you squashed it :)

notroj:
I pushed your fork to get a CI run and there are a couple of compiler warnings too https://github.com/notroj/httpd/actions/runs/11028425131

modules/ssl/ssl_engine_kernel.c
```diff
@@ -2465,11 +2473,49 @@ int ssl_callback_ServerNameIndication(SSL *ssl, int *al, modssl_ctx_t *mctx)
...
+    modssl_clienthello_vars *clienthello_vars;
+    const unsigned char *data;
+    int *ids;
+
+    sslcon = myConnConfig(c);
+
+    sslcon->clienthello_vars = apr_palloc(c->pool, sizeof(*clienthello_vars));
+    clienthello_vars = sslcon->clienthello_vars;
+
+    if (clienthello_vars) {
```
ylavic: apr_palloc() cannot return NULL in httpd, no need for this conditional.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -2465,11 +2473,49 @@ int ssl_callback_ServerNameIndication(SSL *ssl, int *al, modssl_ctx_t *mctx)
...
+        SSL_client_hello_get0_ext(ssl, 0x0a, &data, &clienthello_vars->ecgroups_len);
+        clienthello_vars->ecgroups_data = apr_pmemdup(c->pool, data, clienthello_vars->ecgroups_len);
+        SSL_client_hello_get0_ext(ssl, 0x0b, &data, &clienthello_vars->ecformats_len);
+        clienthello_vars->ecformats_data = apr_pmemdup(c->pool, data, clienthello_vars->ecformats_len);
+        SSL_client_hello_get0_ext(ssl, 0x0d, &data, &clienthello_vars->sigalgos_len);
+        clienthello_vars->sigalgos_data = apr_pmemdup(c->pool, data, clienthello_vars->sigalgos_len);
+        SSL_client_hello_get0_ext(ssl, 0x10, &data, &clienthello_vars->alpn_len);
+        clienthello_vars->alpn_data = apr_pmemdup(c->pool, data, clienthello_vars->alpn_len);
+        SSL_client_hello_get0_ext(ssl, 0x2b, &data, &clienthello_vars->versions_len);
+        clienthello_vars->versions_data = apr_pmemdup(c->pool, data, clienthello_vars->versions_len);
```
ylavic: Don't these magic `0x..` extensions have a `#define` somewhere?

modules/ssl/ssl_engine_kernel.c
```diff
@@ -2520,6 +2566,12 @@ int ssl_callback_ClientHello(SSL *ssl, int *al, void *arg)
 
 give_up:
     init_vhost(c, ssl, servername);
+    
+    s = mySrvFromConn(c);
+    sc = mySrvConfig(s);
```
ylavic: Using `mySrvConfigFromConn(c)->clienthello_vars` directly could do probably.

modules/ssl/ssl_private.h
```diff
@@ -574,6 +574,30 @@ typedef enum {
...
+ * Define the structure to hold clienthello variables
+ * (later exposed as environment vars)
+ */
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+typedef struct {
+    unsigned int version;
+    apr_size_t ciphers_len;
+    const unsigned char *ciphers_data;
+    apr_size_t extids_len;
+    int *extids_data;
```
ylavic: `const int *`?

csmutz:
These are the critical questions in implementation approach here--I'm willing to change.

> Higher-level comment: does this need to be a new directive? I think it would be better as another `SSLOptions` option, it's quite similar to `SSLOptions +ExportCertData` . 

I used a directive to help clarify that the SSLClientHelloVars directive only applies at the server/virtual host level, not at directory level, and because this directive has preformance impacts for every connection where it is enabled, regardless if StdEnvVars is on or not. This is because the clienthello variables have to be collected at the start of the connection (you can't retrieve them later if they aren't stored) and the default needs to be to not collect these (most users don't want the additional overhead, even if small). I viewed this as different from the other SSLOptions which are focused on ENV variable creation/clutter only (the certs are available, ExportCertData just switches additional parsing, env vars). 

If you want me to convert this to an  SSLOptions, I will do so. Then we would just document that this specific option only applies at the server, virtualhost level (is ignored elsewhere) and this option impacts more than env var clutter, it has a performance and memory impact on every connection for the life of the connection?

> Secondly, if the only point of exporting these is to create the JA3/JA4 fingerprints, would it be better to do that within `mod_ssl`? If the algorithm is simple enough and stable, it would not be inappropriate feature I guess.

The fact that we have both ja3 and ja4 speak to the longevity of these. I don't know how long it will be until there is yet another TLS fingerprint method. For my purposes, both ja3 and ja4 don't meet all my needs and I would like to be able to use the raw values in both logs and cgi scripts. I think there are others who will want to do so. Ex. to answer questions like what percentage of clients still advertise TLS 1.0 or TLS 1.1 in supported versions? I think exposing the raw values is best to cover all possible use cases. At the very least, this patch makes implementing ja3 and ja4 easier as the source values are readily available. Regarding implementation difficulty, ja3 is pretty simple. ja4 has a lot of moving parts and edge cases (I've seen not quite right ja4 implementations). If you want to make acceptance of PR contingent upon implementation of ja3 and ja4 in addition to exposing the raw data I would probably do that, but removing access to the raw variables would be a mistake IMO.

I'll work on the fixes requested. Can you give me a pointer to methodology to compile/test against trunk so I can replicate your test pipeline? I've been testing on 2.4 release and simply manually splicing diffs in trunk. This worked in #477 but this is much bigger patch.

notroj:
Thanks a lot for the detailed response @csmutz. Makes sense to me - keep it as a new directive, and keep it exporting everything.

Regarding dev methodology, I recommend starting with git or SVN trunk. If you develop against trunk (whether from git or SVN), you can run the test suite directly. Something like this is a good starting point:

```
$ svn co https://svn.apache.org/repos/asf/httpd/test/framework/trunk test/framework
$ ./configure --with-test-suite=./test/framework --enable-maintainer-mode=c11 --enable-mods-shared=reallyall ...
$ make && make check
```

... should be documented somewhere but maybe isn't. Unfortunately because of quirks of our hybrid SVN/git config the CI doesn't run for non-committer PRs currently. I think it should run when you push to your personal fork but maybe you need GitHub Actions enabled or something.

csmutz:
@notroj 

In addition to all the above fixes, I have two things I'd like you to double check for me. I really don't want to introduce any subtle errors. 

1. Mismatch in define statements for openssl version 1.1.1.

In the clienthello callback I piggyback on the existing define statement that includes a check for LIBRESSL_VERSION_NUMBER:

```
#if OPENSSL_VERSION_NUMBER >= 0x10101000L && !defined(LIBRESSL_VERSION_NUMBER)
```

everywhere else the define statements I use only check openssl version:

```
#if OPENSSL_VERSION_NUMBER >= 0x10101000L
```

Is there a potential to cause a problem (failure to compile) with this mismatch? I don't understand why both openssl and libressl versions would be defined at the same time but if that never occurs, why would this condition exist? I could easily duplicate the libressl check but I don't want to duplicate it if I don't understand it.

2. Reliance on calloc in initialization of sslconn to test for allocation and population of clienthello_vars struct.

At variable formatting time (per request) I check to see if the clienthello variables were populated previously by testing the pointer to the clienthello_vars [ssl_engine_vars.c line 991](https://github.com/apache/httpd/pull/483/files#diff-4704312f9f23dfb697dfa3f1c6f71ff54fc43c01ead7f73fdf2c952b54dd09ccR991) struct which should fail if clienthello_vars weren't populated because of the calloc used to create sslconn [mod_ssl.c line 482](https://github.com/csmutz/httpd/blob/trunk/modules/ssl/mod_ssl.c#L482). I've convinced myself the logic here is sound (and testing doesn't result in crashes) and I think this follows conventions I see in other places, but could you double check that you agree there's no way an uninitialized pointer could be used here?

Thanks very much for the help!

modules/ssl/ssl_engine_vars.c
csmutz: fixed

modules/ssl/ssl_engine_vars.c
csmutz: fixed, using apr_snprintf

modules/ssl/ssl_engine_vars.c
csmutz: fixed

modules/ssl/ssl_engine_kernel.c
csmutz: fixed, removed unnecessary conditional

modules/ssl/ssl_engine_kernel.c
csmutz: fixed, using extension definitions from openssl (consistent with definition used in clienthello callback to get SNI).

modules/ssl/ssl_engine_kernel.c
csmutz: fixed, using much more concise code as recommended

modules/ssl/ssl_private.h
csmutz: fixed, changed to const int *

csmutz:
I've fixed all the requested changes and have fixed the compiler warnings.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -2465,6 +2473,40 @@ int ssl_callback_ServerNameIndication(SSL *ssl, int *al, modssl_ctx_t *mctx)
...
+    sslcon = myConnConfig(c);
+
+    sslcon->clienthello_vars = apr_palloc(c->pool, sizeof(*clienthello_vars));
+    clienthello_vars = sslcon->clienthello_vars;
+
+    clienthello_vars->version = SSL_client_hello_get0_legacy_version(ssl);
+    clienthello_vars->ciphers_len = SSL_client_hello_get0_ciphers(ssl, &data);
+    clienthello_vars->ciphers_data = apr_pmemdup(c->pool, data, clienthello_vars->ciphers_len);
+    if (SSL_client_hello_get1_extensions_present(ssl, &ids, &clienthello_vars->extids_len) == 1) {
+        clienthello_vars->extids_data = apr_pmemdup(c->pool, ids, clienthello_vars->extids_len * sizeof(int));
```
ylavic: Does this work if ->extids_len == 0 can be returned succesfully?

Also, to avoid a possible dangling ->extids_data I would either:
```
else {
    clienthello_vars->extids_data = NULL;
    clienthello_vars->extids_len = 0;
}
```
or use apr_pcalloc() above, the additional/implicit memset() would be in the noise in this path anyway (I think), possibly more future extensions proof too.

modules/ssl/ssl_engine_vars.c
```diff
@@ -975,6 +979,62 @@ static const char *ssl_var_lookup_ssl_handshake_rtt(apr_pool_t *p, SSL *ssl)
     return NULL;
 }
 
+static const char *ssl_var_lookup_ssl_clienthello(apr_pool_t *p, const SSLConnRec *sslconn, const char *var)
+{
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+    char *value;
+    modssl_clienthello_vars *clienthello_vars;
+    int i;
```
ylavic: `apr_size_t i` to match the type of `->extids_len`

ylavic:
>     1. Mismatch in define statements for openssl version 1.1.1.

The problem here is that libressl always defines OPENSSL_VERSION_NUMBER to 0x20000000L (i.e. 2.0 has been appropriated while openssl was still in 1.x), that's why openssl went to 3.x directly after 1.x too.
Anyway, we have the situation where OPENSSL_VERSION_NUMBER >= 0x10101000L is always true for libressl, although all the new openssl-1.x APIs have not made it to libressl..
So using OPENSSL_VERSION_NUMBER >= 0x10101000L with or without checking LIBRESSL_VERSION_NUMBER depends on whether libressl actually has the enclosed feature.
If you don't (want to) know, require !defined(LIBRESSL_VERSION_NUMBER) and someone who care about libressl will either remove it or change to the exact LIBRESSL_VERSION_NUMBER where implemented.

>     2. Reliance on calloc in initialization of sslconn to test for allocation and population of clienthello_vars struct.

Looks fine here for the "sslconn->clienthello_vars" pointer itself.
Maybe zero the pointed to data [here](https://github.com/apache/httpd/pull/483/files#diff-68e64c00f254203d3cf1269f34795c960046adee39e7ecd57cbca3af0dbb77b8R2488) too (see new comment there).

modules/ssl/ssl_engine_kernel.c
csmutz: good catch. We have a big issue if extids_len is never set.

csmutz:
> The problem here is that libressl always defines OPENSSL_VERSION_NUMBER to 0x20000000L (i.e. 2.0 has been appropriated while openssl was still in 1.x), that's why openssl went to 3.x directly after 1.x too.

Oh, that's ugly.

Since the clienthello callback was already wrapped in the !libressl check, I'll replicate this same check across all the other places. As it is, this won't compile for libressl.

csmutz:
Thanks for your patience and help.

I've made the preprocessor checks for openssl version consistent (libressl won't have this functionality) and addressed the requested items above.

modules/ssl/ssl_engine_vars.c
csmutz: fixed

modules/ssl/ssl_engine_kernel.c
csmutz: There were a couple serious bugs and other issues here. Much of this code assumed that the openssl functions would set the associated lengths correctly even if the functions returned failure which is not right. 

To fix these issues, I've rewritten this whole code block doing the following:
 - using calloc for struct (strictly shouldn't be necessary but agreed this is not expensive)
 - correctly check return codes from various SSL_client_hello_* functions, explicitly set len vars to 0 on failure
 - conditions to check len > 0 before memdup, explicitly set pointers to null if len == 0
 
I also have tested with diversity of clients including some lacking some collected extensions that prior to fixes would cause crash and other undefined behavior. These cases were all handled correctly after fixes.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -2485,26 +2485,72 @@ static void copy_clienthello_vars(conn_rec *c, SSL *ssl)
...
     clienthello_vars = sslcon->clienthello_vars;
 
     clienthello_vars->version = SSL_client_hello_get0_legacy_version(ssl);
+
     clienthello_vars->ciphers_len = SSL_client_hello_get0_ciphers(ssl, &data);
-    clienthello_vars->ciphers_data = apr_pmemdup(c->pool, data, clienthello_vars->ciphers_len);
+    if (clienthello_vars->ciphers_len > 0) {
+        clienthello_vars->ciphers_data = apr_pmemdup(c->pool, data, clienthello_vars->ciphers_len);
+    } else {
+        clienthello_vars->ciphers_data = NULL;
```
notroj: Since you're now using `apr_pcalloc` the explicit setting of NULL / 0 through here is not necessary.

csmutz: Done. I've removed the explicit/redundant assignments of _len to 0 and _data to NULL which should be fine as long as failed calls to the openssl client hello functions never modify _len. I've left the checks for _len >0 before memdup.

I've tested on conditions that triggered bugs due to uninitialized or 0 _len before by testing connections where some of the collected client hello extensions is not present and these cases functioned correctly.

ylavic (approved):
+1, thanks for this PR!

notroj:
Seconded, thanks again.

PR #486, opened 2024-10-05
mod_proxy: Allow override Host header and SSL SNI in ProxyPass

leo9800:
This patch adds 2 optional parameters, `override_http_host` and `override_ssl_sni` for `ProxyPass` directive.

`override_http_host` allows setting custom HTTP `Host:` header. This may be done by utilizing `ProxyPreserveHost On` and `RequestHeader set Host`.

`override_ssl_sni` allows setting custom SSL server name (SNI). This seems could not be done without this patch.

This patch could be helpful in the scenario described below:

- A HTTPS server on `203.0.113.2:443` hosting a site `domain.tld`, **with strict HTTP Host header check and TLS SNI check**.
- Apache HTTPd, running on `203.0.113.1`, is being used to reverse proxy `https://domain.tld` hosted on `203.0.113.2:443`
- There is no (type A) DNS record pointed to `203.0.113.2`, which forced us to use IP address in `ProxyPass`
- Administrator adds `ProxyPass / https://203.0.113.2` to Apache's configuration
- Reverse proxy fails, because SNI and HTTP host was `203.0.113.2` instead of `domain.tld`
- Administrator adds `ProxyPreserveHost On` and `RequestHeader set Host domain.tld` to Apache's configuration
- Reverse proxy still fail, because only HTTP host header was set correctly, SNI is still incorrect.

With this patch ...

- Administrator adds a single line `ProxyPass / https://203.0.113.2 override_http_host=domain.tld override_ssl_sni=domain.tld`
- It works

Besides, this patch introduce no breaking change. (i.e. incompatibility against current configurations)

### DIFF
--- a/modules/proxy/mod_proxy.c
+++ b/modules/proxy/mod_proxy.c
@@ -368,6 +368,20 @@ static const char *set_worker_param(apr_pool_t *p,
         return "multipathtcp is not supported on your platform";
 #endif
     }
+    else if (!strcasecmp(key, "override_http_host")) {
+        if (strlen(val) >= sizeof(worker->s->override_http_host))
+            return apr_psprintf(p, "Override HTTP Host header length must be < %d characters",
+                                (int)sizeof(worker->s->override_http_host));
+        PROXY_STRNCPY(worker->s->override_http_host, val);
+        worker->s->override_http_host_set = 1;
+    }
+    else if (!strcasecmp(key, "override_ssl_sni")) {
+        if (strlen(val) >= sizeof(worker->s->override_ssl_sni))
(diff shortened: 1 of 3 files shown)


ylavic:
Can't the same be achieved by adding an `203.0.113.2:443 domain.tld` entry in `/etc/hosts` and simply using `ProxyPass / https://domain.tld` ?

leo9800:
@ylavic Yes, this could be done by appending `203.0.113.2 domain.tld` to `/etc/hosts`, theoretically.

But in case we could not modify `/etc/hosts`, says, doing so could break other software running on same host which httpd lives on, (_203.0.113.2_ in our sample scenario) then this patch would help to address such dilemma.

PR #492, opened 2024-10-23
Remove dead assignments

partimScurra:
Removed some minor dead assignments and initializations detected by static analyzer.


notroj:
Dead assignments are a programming error of some sort, but removing the assignment in a way which preserves behaviour may preserve whatever error was made at the same time as eliminating a warning. For example:

https://github.com/apache/httpd/pull/492/commits/733801cb2b37e60ebab591a1725323212895ee91#diff-0b6fc2b9b7368823d9b36f7307ebaab80552c906effb06812f796aa9c25cec73 

it is not obvious that removing the first assignment is better than removing the second one. IMO, these should be investigated rather than simply removed. Or, submit fixes for those which are obviously correct.

notroj:
No further feedback, closing.

PR #496, opened 2024-10-28
ap_log_error: Include ap_config_auto.h to define _GNU_SOURCE

bkuhls:
gettid() needs _GNU_SOURCE defined which is provided by ap_config_auto.h

Fixes buildroot error:
http://autobuild.buildroot.net/results/2f6/2f6b7bbb4c97e4c91b3abd6bb205e237e57045fa//build-end.log

log.c: In function 'log_tid':
log.c:637:21: error: implicit declaration of function 'gettid'; did you mean 'getgid'? [-Wimplicit-function-declaration]
  637 |         pid_t tid = gettid();

### DIFF
--- a/server/log.c
+++ b/server/log.c
@@ -21,6 +21,10 @@
  *
  */
 
+#if !defined(WIN32) && !defined(NETWARE)
+#include "ap_config_auto.h"
+#endif
+
 #include "apr.h"
 #include "apr_general.h"        /* for signal stuff */
 #include "apr_strings.h"


notroj:
What glibc version is used there?

bkuhls:
afaics
```
GNU C Library (Buildroot) stable release version 2.39.
```

notroj:
Thanks. I don't understand why this is necessary. 

https://github.com/apache/httpd/blob/trunk/server/log.c#L52 -> `#include "ap_config.h"`
https://github.com/apache/httpd/blob/trunk/include/ap_config.h#L183 -> `#include "ap_config_auto.h"`

The `ap_config_auto.h.in` generated by autoheader should have:

```
/* Enable GNU extensions on systems that have them.  */
#ifndef _GNU_SOURCE
# undef _GNU_SOURCE
#endif
```

Which happens on my system (Fedora with autoconf 2.71) , and in the tarballs released at `apache.org`. So I would assume that something is special about the way you're running autoheader?

notroj:
That can't be it, because your patch also assumes that `ap_config_auto.h` defines _GNU_SOURCE.
Moving the `#include "ap_config.h"` to the top of log.c looks correct/justifiable, but directly including `ap_config_auto.h` here is wrong/unnecessary.

I think there's something broken with this build environment due to the cross-build. You are not seeing "normal" CPPFLAGS from APR 1.x either, which on Linux should look like:

```
checking for APR... yes
  setting CPP to "gcc -E"
  setting CPPFLAGS to " -DLINUX -D_REENTRANT -D_GNU_SOURCE"
```

Instead you see only:

```
checking for APR... yes
  setting LDFLAGS to " "
```

notroj:
I'm still guessing this is a defect of a cross-build environment which should be investigated and fixed properly rather than trying to tweak the headers, closing since no feedback but feel free to update/reopen.

PR #499, opened 2024-12-15
log getpwuid error

divinity76:
was in a situation where getpwuid() failed, and I would've liked to see the errno/strerror at the time of the error.

quoting https://pubs.opengroup.org/onlinepubs/009604499/functions/getpwuid.html
> Applications wishing to check for error situations should set errno to 0 before calling getpwuid(). If getpwuid() returns a null pointer and errno is set to non-zero, an error occurred.

### DIFF
--- a/modules/arch/unix/mod_unixd.c
+++ b/modules/arch/unix/mod_unixd.c
@@ -97,12 +97,13 @@ static int set_group_privs(void)
         if (ap_unixd_config.user_name[0] == '#') {
             struct passwd *ent;
             uid_t uid = atol(&ap_unixd_config.user_name[1]);
-
+            errno = 0;
             if ((ent = getpwuid(uid)) == NULL) {
                 ap_log_error(APLOG_MARK, APLOG_ALERT, errno, NULL, APLOGNO(02155)
-                         "getpwuid: couldn't determine user name from uid %ld, "
+                         "getpwuid: couldn't determine user name from uid %ld (errno %d: %s), "
                          "you probably need to modify the User directive",
-                         (long)uid);
+                         (long)uid, errno, strerror(errno));
+
                 return -1;
             }
 


covener:
do you still have the original logging handy?  Usually it would have been incorporated into the prefix of the message by being already passed as the third arg.

divinity76:
@covener is the errno in here? 
```
[Sun Dec 15 14:00:09.049198 2024] [unixd:alert] [pid 366799:tid 140737350023040] AH02155: getpwuid: couldn't determine user name from uid 4294967295, you probably need to modify the User directive
[Sun Dec 15 14:00:09.170560 2024] [core:crit] [pid 366799:tid 140737339500224] [client 127.0.0.1:43404] AH00025: configuration error:  couldn't check user: /
[Sun Dec 15 14:08:13.207344 2024] [core:crit] [pid 477285:tid 140737339291328] [client 127.0.0.1:37614] AH00025: configuration error:  couldn't check user: /
[Sun Dec 15 14:11:00.561799 2024] [core:crit] [pid 477416:tid 140737339291328] [client 127.0.0.1:55350] AH00025: configuration error:  couldn't check user: /
```
(edit: added the errors that followed too, just in case)

covener:
No it's not incorporated there, an example is the `(13)Permission denied:` in:

```
[Sun Dec 15 08:34:47.756892 2024] [core:error] [pid 20158:tid 13125787648] (13)Permission denied: [client 127.0.0.1:56009] AH00132: file permissions deny server access: /Users/covener/SRC/2.4.x/built/htdocs/index.html
```

which comes from
```
            ap_log_rerror(APLOG_MARK, APLOG_ERR, status, r, APLOGNO(00132)
                          "file permissions deny server access: %s", r->filename);
```

Does your patch end up recording a non-zero errno/strerrror in the same situation?

divinity76:
hmm no, and now i'm unable to re-create the conditions needed to trigger the error in the first place 🤔 when trying, I now get the error
```
[Sun Dec 15 14:43:49.302411 2024] [core:crit] [pid 540727:tid 140737339291328] [client 127.0.0.1:45078] AH00025: configuration error:  couldn't check user: /
```
instead

divinity76:
i'll close this pr, and if i encounter the same issue again, and find a way to reproduce it, perhaps i'll re-open it.

PR #503, opened 2025-01-01
mod_session_dbd: set_cookie_name: ensure correct format

thomasmey:
If args is an empty string, apr_strtok will return NULL and *last will never get set which results in a SIGSEGV in apr_isspace check

### DIFF
--- a/modules/session/mod_session_dbd.c
+++ b/modules/session/mod_session_dbd.c
@@ -571,6 +571,11 @@ static const char *set_cookie_name(cmd_parms * cmd, void *config, const char *ar
     char *line = apr_pstrdup(cmd->pool, args);
     session_dbd_dir_conf *conf = (session_dbd_dir_conf *) config;
     char *cookie = apr_strtok(line, " \t", &last);
+    if (!cookie) {
+        return apr_pstrcat(cmd->pool, cmd->directive->directive,
+                           " requires at least one argument!",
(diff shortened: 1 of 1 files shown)


modules/session/mod_session_dbd.c
```diff
@@ -571,6 +571,9 @@ static const char *set_cookie_name(cmd_parms * cmd, void *config, const char *ar
     char *line = apr_pstrdup(cmd->pool, args);
     session_dbd_dir_conf *conf = (session_dbd_dir_conf *) config;
     char *cookie = apr_strtok(line, " \t", &last);
+    if(!cookie) {
+        return "Invalid directive, skipping";
```
covener: The string has "skipping" but won't this result in a startup error?  Maybe "SessionDBDCookieName requires at least one argument"?

modules/session/mod_session_dbd.c
```diff
@@ -571,6 +571,9 @@ static const char *set_cookie_name(cmd_parms * cmd, void *config, const char *ar
     char *line = apr_pstrdup(cmd->pool, args);
     session_dbd_dir_conf *conf = (session_dbd_dir_conf *) config;
     char *cookie = apr_strtok(line, " \t", &last);
+    if(!cookie) {
```
covener: nit. httpd style has a space between if and (

PR #504, opened 2025-01-01
mod_log_config: better handle invalid log file name

thomasmey:
if log file name is an invalid string, ap_server_root_relative will return NULL, which will result in SIGSEGV in ap_make_dirstr_parent


covener:
What's an invalid string? Can it be reached via configuration?

PR #505, opened 2025-01-01
core: error_log: better handle invalid error log file name

thomasmey:
if error log file name is an invalid string, ap_server_root_relative will return NULL, which will result in SIGSEGV in ap_make_dirstr_parent


covener:
What's an invalid string? Can it be reached via configuration?

thomasmey:
I have the fuzzer corpus on my harddrive, i can send it to you directly or maybe upload somewhere

server/core.c
```diff
@@ -5316,15 +5316,24 @@ AP_DECLARE(int) ap_sys_privileges_handlers(int inc)
 
 static int check_errorlog_dir(apr_pool_t *p, server_rec *s)
 {
+    char *dir;
+    apr_finfo_t finfo;
+    apr_status_t rv;
+
```
covener: why broader scope for these if still used only within the else?

server/core.c
```diff
@@ -5316,15 +5316,24 @@ AP_DECLARE(int) ap_sys_privileges_handlers(int inc)
...
         return APR_SUCCESS;
     }
     else {
         char *abs = ap_server_root_relative(p, s->error_fname);
-        char *dir = ap_make_dirstr_parent(p, abs);
-        apr_finfo_t finfo;
-        apr_status_t rv = apr_stat(&finfo, dir, APR_FINFO_TYPE, p);
+        if (!abs) {
+            ap_log_error(APLOG_MARK, APLOG_STARTUP|APLOG_EMERG, 0,
+                          ap_server_conf, APLOGNO(02291)
```
covener: should be `APLOGNO()`initially and then can be filled in after committed.

server/core.c
```diff
@@ -5316,15 +5316,24 @@ AP_DECLARE(int) ap_sys_privileges_handlers(int inc)
...
+                         "Cannot construct error log file path '%s'", s->error_fname);
```
covener, line 5331: can you include the input that triggers this? Hopefully in a way as simple as httpd.conf content rather than calling these functions directly.

covener (changes requested):

PR #506, opened 2025-01-02
fix value inconsistency in LimitXMLRequestBody directive example

Chenjp:
# Limit of 1 MiB
LimitXMLRequestBody 1073741824 -> LimitXMLRequestBody ***1048576***


covener:
Thanks!

PR #512, opened 2025-01-23
fix BZ 69546, enforce file size limitation for Partial PUT resource

Chenjp:
Enforce LimitRequestBody to Partial PUT destination resource.


notroj:
Thanks for the PR, but I think the current behaviour is correct per https://bz.apache.org/bugzilla/show_bug.cgi?id=69546#c3

Chenjp:
> Thanks for the PR, but I think the current behaviour is correct per https://bz.apache.org/bugzilla/show_bug.cgi?id=69546#c3

Hi, is there an official way to limit the single file size via webdav PUT (upload fully or partially)?

PR #516, opened 2025-02-13
Fix variable declaration after statement warning in scoreboard.c

jajik:
The latest commit broke compilation with maintainer mode (the warning is treated as an error) incl. some jobs in CI.

https://github.com/apache/httpd/actions/runs/13282453047/job/37083598064

This PR fixes it.

https://github.com/jajik/httpd/actions/runs/13311390743

### DIFF
--- a/server/scoreboard.c
+++ b/server/scoreboard.c
@@ -657,11 +657,11 @@ AP_DECLARE(void) ap_time_process_request(ap_sb_handle_t *sbh, int status)
 AP_DECLARE(void) ap_set_time_process_request(ap_sb_handle_t* const sbh,
 		const apr_time_t timebeg,const apr_time_t timeend)
 {
+    worker_score *ws;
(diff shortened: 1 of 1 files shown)


gstrauss:
The C99 specification was released *over 25 years ago!*
The C11 specification was released *over 13 years ago!*

Why is Apache still enforcing C89/C90 semantics?

An embedded system with an ancient compiler is probably not going to go using Apache 2.4.  It theoretically could use Apache 1.4, but more likely would use a different modern web server with lower resource consumption than Apache 2.4.

notroj:
@gstrauss httpd has a very high inertia :) We discussed on dev@ recently that for 2.5.x we could move to a more modern C standard, I'm definitely in favour of it.

notroj:
Thanks for the PR @jajik

icing:
Thanks for merging, @notroj !

PR #517, opened 2025-02-21
FIX: mod_lua: OutOfMemory in LuaOutputFilter (with LARGE data) bugzilla#69590

ggrandes:
Proposed patch to resolve OutOfMemory, in local testing it has worked correctly. 
https://bz.apache.org/bugzilla/show_bug.cgi?id=69590

### DIFF
--- a/modules/lua/mod_lua.c
+++ b/modules/lua/mod_lua.c
@@ -520,7 +520,7 @@ static apr_status_t lua_output_filter_handle(ap_filter_t *f, apr_bucket_brigade 
         }
         /* If we've safely reached the end, do a final call to Lua to allow for any 
         finishing moves by the script, such as appending a tail. */
-        if (APR_BUCKET_IS_EOS(APR_BRIGADE_LAST(pbbIn))) {
+        if (!APR_BRIGADE_EMPTY(pbbIn) && APR_BUCKET_IS_EOS(APR_BRIGADE_LAST(pbbIn))) {
             apr_bucket *pbktEOS;
             lua_pushnil(L);
             lua_setglobal(L, "bucket");
--- a/modules/lua/mod_lua.c
+++ b/modules/lua/mod_lua.c
@@ -498,6 +498,7 @@ static apr_status_t lua_output_filter_handle(ap_filter_t *f, apr_bucket_brigade 
                     APR_BRIGADE_INSERT_TAIL(ctx->tmpBucket, pbktOut);
                     rv = ap_pass_brigade(f->next, ctx->tmpBucket);
                     apr_brigade_cleanup(ctx->tmpBucket);
+                    apr_bucket_delete(pbktIn);
                     if (rv != APR_SUCCESS) {
                         return rv;
                     }
--- a/modules/lua/mod_lua.c
+++ b/modules/lua/mod_lua.c
@@ -473,10 +473,12 @@ static apr_status_t lua_output_filter_handle(ap_filter_t *f, apr_bucket_brigade 
     L = ctx->L;
     /* While the Lua function is still yielding, pass in buckets to the coroutine */
     if (!ctx->broken) {
-        for (pbktIn = APR_BRIGADE_FIRST(pbbIn);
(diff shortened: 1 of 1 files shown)


modules/lua/mod_lua.c
```diff
@@ -498,6 +498,7 @@ static apr_status_t lua_output_filter_handle(ap_filter_t *f, apr_bucket_brigade
                     APR_BRIGADE_INSERT_TAIL(ctx->tmpBucket, pbktOut);
                     rv = ap_pass_brigade(f->next, ctx->tmpBucket);
                     apr_brigade_cleanup(ctx->tmpBucket);
+                    apr_bucket_delete(pbktIn);
```
notroj: This is along the right lines but the code needs to be rewritten so the loop still works.

After `apr_bucket_delete(pbktIn)` the `APR_BUCKET_NEXT(pbktIn)` call will not work (reliably).

See also https://httpd.apache.org/docs/2.4/developer/output-filters.html#filtering

The loop should likely be of the form
```
while ((pbktIn = APR_BRIGADE_FIRST(bb)) != APR_BRIGADE_SENTINEL(bb)) {
```

ggrandes: I don't understand. I've tested this patch (concurrency and processed data integrity), and it fixes the bug without breaking anything.

ggrandes: Do you mean that instead of making the minimum necessary change, I should be better rewrite the original code? 🤕 
I don't know if I have necessary expertise, but If you think, I can try... 😅

ylavic: The added:
```
   apr_bucket_delete(pbktIn);
```
here is creating a use-after-free in the upper loop:
```
        for (pbktIn = APR_BRIGADE_FIRST(pbbIn);
            pbktIn != APR_BRIGADE_SENTINEL(pbbIn);
            pbktIn = APR_BUCKET_NEXT(pbktIn)) // <= HERE
```
or even the below:
```
            if (APR_BUCKET_IS_EOS(APR_BRIGADE_LAST(pbbIn))) // <= HERE
```

The loop should be structured like this instead:
```
        while (!APR_BRIGADE_EMPTY(pbbIn)) {
             pbktIn = APR_BRIGADE_FIRST(pbbIn);

             ...

             apr_bucket_delete(pbktIn);
        }
```

ggrandes: Thanks for the guidance @notroj, @ylavic 👍

notroj (changes requested):
Looks better, thanks! Only thing obviously missing is that you need a check for the empty brigade in the later test as well, i.e.

```
        if (!APR_BRIGADE_EMPTY(pbbIn) && APR_BUCKET_IS_EOS(APR_BRIGADE_LAST(pbbIn))) {
```

ylavic:
It seems that the code has changed in `trunk`, this PR should target `trunk` first before it can be backported to `2.4.x`.

ylavic:
> It seems that the code has changed in `trunk`, this PR should target `trunk` first before it can be backported to `2.4.x`.

Nevermind, this PR should apply cleanly to `trunk`, I was reading mod_ext_filter's ef_unified_filter() in `trunk` which is indeed quite different :)
Same kind of change should go there to address [BZ 69589](https://bz.apache.org/bugzilla/show_bug.cgi?id=69589) though.

ggrandes:
> Looks better, thanks! Only thing obviously missing is that you need a check for the empty brigade in the later test as well, i.e.
> 
> ```
>         if (!APR_BRIGADE_EMPTY(pbbIn) && APR_BUCKET_IS_EOS(APR_BRIGADE_LAST(pbbIn))) {
> ```

Done!

notroj (approved):
Super, thanks.

ggrandes:
@notroj Is there any timeline or estimate for when the hotfix will be applied to the 2.4 branch? Maybe it will be seen in 2.4.64? Thanks!

notroj:
@ggrandes I've nominated it for 2.4.x backport, depends on votes how fast it goes :) Thanks again for for the contribution.

PR #522, opened 2025-03-11
 mod_proxy_fcgi: Support Dynamic Buffering for Large FastCGI Headers

arturobernalg:
This pull request implements a dynamic header buffer in mod_proxy_fcgi to resolve Bug [64919](https://bz.apache.org/bugzilla/show_bug.cgi?id=64919). Previously, mod_proxy_fcgi used a fixed 8192-byte buffer to parse FastCGI headers, which caused failures when PHP-FPM (or similar backends) generated headers exceeding that size. With this patch, a dynamically resizable buffer is used so that arbitrarily large headers are handled gracefully.

### DIFF
--- a/modules/proxy/mod_proxy_fcgi.c
+++ b/modules/proxy/mod_proxy_fcgi.c
@@ -18,6 +18,7 @@
 #include "util_fcgi.h"
 #include "util_script.h"
 #include "ap_expr.h"
+#include "util_varbuf.h"
 
 module AP_MODULE_DECLARE_DATA proxy_fcgi_module;
 
@@ -593,6 +594,7 @@ static int handle_headers(request_rec *r, int *state,
     return 0;
 }
 
+
 static apr_status_t dispatch(proxy_conn_rec *conn, proxy_dir_conf *conf,
                              request_rec *r, apr_pool_t *setaside_pool,
                              apr_uint16_t request_id, const char **err,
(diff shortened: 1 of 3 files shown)


modules/proxy/mod_proxy_fcgi.c
```diff
@@ -593,6 +594,72 @@ static int handle_headers(request_rec *r, int *state,
     return 0;
 }
 
+/**
```
notroj: This part should be added to the varbuf API in server/util.c or else the existing `ap_varbuf_*` should be used, IMO.

modules/proxy/mod_proxy_fcgi.c
```diff
@@ -640,6 +707,10 @@ static apr_status_t dispatch(proxy_conn_rec *conn, proxy_dir_conf *conf,
     ib = apr_brigade_create(r->pool, c->bucket_alloc);
     ob = apr_brigade_create(r->pool, c->bucket_alloc);
 
+    /* Create our dynamic header buffer for large headers */
+    ap_varbuf *header_vb = NULL;
```
notroj: Declaration should be at the top of the function, this is causing the CI failures.

modules/proxy/mod_proxy_fcgi.c
arturobernalg: sound reasonable move it to  server/util.c

modules/proxy/mod_proxy_fcgi.c
arturobernalg: Fix it

modules/proxy/mod_proxy_fcgi.c
arturobernalg: Done. Please @notroj  do another pass

ethan-gallant:
Hey @notroj & @arturobernalg! Wondering if there is more work needed here or if there's additional steps to take.

Happy to help out if so! :)

modules/proxy/mod_proxy_fcgi.c
```diff
@@ -912,6 +934,7 @@ static apr_status_t dispatch(proxy_conn_rec *conn, proxy_dir_conf *conf,
                              * headers, so this part of the data will need
                              * to persist. */
                             apr_bucket_setaside(b, setaside_pool);
+                            }
```
bukka, line 937: this needs probably change indent for the above code...

PR #523, opened 2025-03-18
mod_ssl: Check the SSLProtocol directive when loading the configuration

mkauf:
Previously, the SSLProtocol directive was checked at runtime. Apache quit if the directive contained an invalid combination of protocols, and logged the message "AH02231: No SSL protocols available [hint: SSLProtocol]".

With this change, most invalid SSLProtocol directives are detected when checking the configuration, e.g. with "httpd -t -f httpd.conf".

Examples of invalid protocol combinations that are catched:
* SSLProtocol "-TLSv1"
* SSLProtocol "-all"
* SSLProtocol "TLSv1.2 -TLSv1.2"

### DIFF
--- a/modules/ssl/ssl_engine_config.c
+++ b/modules/ssl/ssl_engine_config.c
@@ -1483,6 +1483,10 @@ const char *ssl_cmd_SSLOptions(cmd_parms *cmd,
         }
     }
 
+    if (*options == SSL_PROTOCOL_NONE) {
+        return "SSLProtocol: No SSL protocols available";
+    }
+
     return NULL;
 }
 


notroj:
Looks like this was merged from 2.4 and added the new check in the wrong function but it's a trivial fix.
https://github.com/apache/httpd/pull/523/commits/6df6b1a33cf0be181cfa4ef9cf7c076aab7600b3#diff-14a3ac31e799b2734152995888041f2da912afcc74e66e774d926397db65711dR1486

Thanks for the PR!

mkauf:
@notroj You're right... thank you for fixing the commit and merging!

PR #524, opened 2025-03-18
mod_ssl: directly abort connection when receiving HTTP requests

daum3ns:
Shutdown the ssl filter and abort the connection instead of disabling is and pass a fake request to the core handler. The current implementation allows to exhaust workers by sendin HTTP request to HTTPS port. Additionally the Openssl lib doesn't detect the http methods PATCH, DELETE, OPTIONS and TRACE. So the current implementation only works partially. See openssl PR: https://github.com/openssl/openssl/pull/26968

### DIFF
--- a/modules/ssl/ssl_engine_io.c
+++ b/modules/ssl/ssl_engine_io.c
@@ -935,25 +935,6 @@ static apr_status_t ssl_filter_write(ap_filter_t *f,
     return outctx->rc;
 }
 
-/* Just use a simple request.  Any request will work for this, because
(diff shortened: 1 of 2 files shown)


daum3ns:
@ylavic maybe you could have a look here?

notroj:
"current implementation allows to exhaust workers by sendin HTTP request to HTTPS port"

Can you explain exactly what you mean by that? Why is sending an HTTP response on the HTTPS port going to exhaust workers?

PR #527, opened 2025-04-08
mod_asis: Fix the log level of the message AH01236

mkauf:
Change the log level from ERROR to DEBUG.

The message "mod_asis: ap_pass_brigade failed for file ..." was logged with the level ERROR. This log level is inappropriate here, because a client can trigger this log message by aborting the request.

Most other modules don't log at all or use the log level DEBUG when ap_pass_brigade() fails.


notroj:
Thanks, merging.

PR #531, opened 2025-05-21
mod_proxy: restore reuse of ProxyRemote connections when possible.

ylavic:
Fixes a regression from 2.4.59 (r1913907).

For a reverse proxy setup with a worker (enablereuse=on) and a forward/CONNECT ProxyRemote to reach it, an open connection/tunnel to/through the remote proxy for the same origin server (and using the same proxy auth) should be reusable. Avoid closing them like r1913534 did.

* modules/proxy/proxy_util.c:
  Rename the struct to remote_connect_info since it's only used for connecting through remote CONNECT proxies. Axe the use_http_connect field, always true.

* modules/proxy/proxy_util.c(ap_proxy_connection_reusable):
  Remote CONNECT (forward) proxy connections can be reused if the auth and origin server infos are the same, so conn->forward != NULL is not a condition to prevent reusability.

* modules/proxy/proxy_util.c(ap_proxy_determine_connection):
  Fix the checks around conn->forward reuse and connection cleanup if that's not possible.

### DIFF
--- a/modules/proxy/proxy_util.c
+++ b/modules/proxy/proxy_util.c
@@ -45,16 +45,16 @@
 APLOG_USE_MODULE(proxy);
 
 /*
- * Opaque structure containing target server info when
- * using a forward proxy.
- * Up to now only used in combination with HTTP CONNECT to ProxyRemote
+ * Opaque structure containing infos for CONNECT-ing an origin server through a
+ * remote (forward) proxy. Saved in the (opaque) proxy_conn_rec::forward pointer
+ * field for backend connections kept alive, allowing for reuse when subsequent
+ * requests should be routed through the same remote proxy.
  */
 typedef struct {
-    int          use_http_connect; /* Use SSL Tunneling via HTTP CONNECT */
-    const char   *target_host;     /* Target hostname */
-    apr_port_t   target_port;      /* Target port */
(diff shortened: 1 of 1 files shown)


modules/proxy/proxy_util.c
```diff
@@ -1608,9 +1608,7 @@ PROXY_DECLARE(int) ap_proxy_connection_reusable(proxy_conn_rec *conn)
 {
     proxy_worker *worker = conn->worker;
 
-    return !(conn->close
-             || conn->forward
-             || worker->s->disablereuse);
+    return !(conn->close || worker->s->disablereuse);
```
jfclere, line 1611: That is the dirty patch I proposed, correct?

jfclere (approved):

modules/proxy/proxy_util.c
ylavic: Yes, this allows to reuse remote CONNECT proxy connections eventually, and it's probably enough to recover the performances.
But the code (of mine) in ap_proxy_determine_connection() made no sense too, starting by setting `conn->forward = NULL` and then checking if it didn't change (while the connection was closed anyway in `connection_cleanup()`..). Since both the connection and `conn->forward` are reusable now we'd better change that part too ;)

PR #532, opened 2025-05-22
Backport: mod_proxy: Reuse ProxyRemote connections when possible, like prior to 2.4.59

ylavic:


ylavic:
Merged in 7996ffc ([r1926317](https://svn.apache.org/repos/asf/httpd/httpd/branches/2.4.x@1926317))

PR #534, opened 2025-06-05
mod_md: change types of fields of ocsp_summary_ctx_t

naya451:
The number of members in ostat_by_id may be up to UINT_MAX and there are no guarantees that all types of members (good, revoked or unknown) are present. An integer overflow may also occur in md_ocsp_get_summary() when they are summed as ints.

Change types of good, revoked and unknown to unsigned.

Found by Linux Verification Center (linuxtesting.org) with SVACE.


notroj:
Thanks for the PR.

PR #544, opened 2025-08-02
util_ldap: prevent double-free by nullifying pointer after free()

sanjay20m:
In util_ald_free(), memory referenced by `ptr` is freed without resetting the pointer. If this function (or cleanup routines) are invoked multiple times on the same pointer, a double-free can occur, leading to memory corruption.

Fix by wrapping free() in braces and setting the pointer to NULL after freeing in both shared-memory and non-shared-memory branches. This improves memory safety and prevents accidental reuse of freed pointers.


covener:
This function doesn't have access to change the callers pointer, this PR just changes the local copy in a short function where it will never be accessed again.

PR #545, opened 2025-08-02
mod_systemd: validate LISTEN_FDS environment variable before atoi()

sanjay20m:
In ap_find_systemd_socket(), getenv("LISTEN_FDS") was called and passed directly to atoi() without verifying if getenv() returned NULL. If the LISTEN_FDS environment variable is missing, this results in undefined behavior and can cause a segmentation fault.

Fix:
- Store the result of getenv() in a local variable.
- Check for NULL before calling atoi().
- Log an error and return -1 if LISTEN_FDS is missing.

This prevents a potential crash (Denial of Service) when running under systemd without the LISTEN_FDS environment variable set.

### DIFF
--- a/modules/arch/unix/mod_systemd.c
+++ b/modules/arch/unix/mod_systemd.c
@@ -139,8 +139,16 @@ static int ap_find_systemd_socket(process_rec * process, apr_port_t port) {
         return -1;
     }
 
-    fdcount = atoi(getenv("LISTEN_FDS"));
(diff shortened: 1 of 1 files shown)


notroj:
If `sd_listen_fds` returns > 0 it is guaranteed that $LISTEN_FDS is set.

You have wasted our time with what look like several AI-generated commits. I'm sure we can block your user from submitting PRs if you continue.

PR #548, opened 2025-08-14
Fix cross build

bastien-roucaries:
./configure fails finding .pc files, because it uses the build architecture pkg-config. It should be using AC_PATH_TOOL (or better PKG_PROG_PKG_CONFIG) rather than AC_PATH_PROG.

@rpluem


bastien-roucaries:
should also go to 2.4 branch

PR #549, opened 2025-08-14
Do not test euid=0 before going chroot

bastien-roucaries:
Nowaday chroot need CAP_SYS_CHROOT capability in its user namespace, and could work without root.

Will allow to use chroot with lesser permission.


bastien-roucaries:
@notroj could you propose for backport for 2.4

PR #550, opened 2025-08-17
Freebind

bastien-roucaries:
Hi,

Can you find a backport for 2.4x of freebind patches

On debian side we will like it

@notroj

### DIFF
--- a/docs/manual/mod/mpm_common.xml
+++ b/docs/manual/mod/mpm_common.xml
@@ -889,4 +889,35 @@ client connections</description>
 </usage>
 </directivesynopsis>
 
+<directivesynopsis>
+<name>AcceptErrorsNonFatal</name>
+<description>Treat some errors accepting a new connection as non-fatal
+to the httpd process.</description>
+<syntax>AcceptErrorsNonFatal ON</syntax>
+<default>OFF (ECONNREFUSED, ECONNABORTED, ECONNRESET cause the process to 
+         exit)</default>
+<contextlist><context>server config</context></contextlist>
+<modulelist><module>event</module><module>worker</module>
+<module>prefork</module>
+</modulelist>
+<compatibility>2.4.66 and later</compatibility>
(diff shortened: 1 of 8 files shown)


bastien-roucaries:
I also believe that ECONNREFUSED, ECONNABORTED, or ECONNRESET should be extended to:
- EPERM due to firewall and systemd filtering
- ENOSYS (syscall filtering)

bastien-roucaries:
Any news ?

bastien-roucaries:
@notroj @rpluem  @icing ?

icing:
I have no spare cycles to work on this.

rpluem:
Sorry, but no free cycles either currently.

notroj:
For [387c78b](https://github.com/apache/httpd/pull/550/commits/387c78bd0078c01f720e5b57938f55958991b51e)  I doubt we can extend `ap_listen_rec` safely in a 2.4.x release.

notroj:
I haven't tried it but I think the listener options like V6ONLY and FREEBIND could and should be handled entirely within the `alloc_listener` loop and doesn't really need to be stored in ->flags, i.e. set the option straight after creating the socket, rather than delaying that until `make_sock()` is called much later. Maybe I'm missing something.

notroj:
Moving the setting of FREEBIND and REUSEPORT options into `alloc_listener()` looks feasible but we'd then have to duplicate the `setsockopt` calls in `ap_duplicate_listener` for these options to manually copy them to the duplicate bucket listeners. So at least those options are doable for 2.4.x but it'll be slightly less clean code and more effort to backport.

[^ manually copying the options using getsockopt(orig)/setsockopt(new) each time, kind of ugly but it will work]

notroj:
This needs to be redone safely for 2.4.x without touching `ap_listen_rec`, somehow, feel free to reopen if you want to attempt that.

PR #551, opened 2025-08-18
OpenSSL ECH integration

sftcd:
This PR adds Encrypted Client Hello ([ECH](https://datatracker.ietf.org/doc/draft-ietf-tls-esni/)) functionality to apache2, when using OpenSSL for TLS.

Notes:

- ECH is not yet part of an OpenSSL release. We'd hope ECH will be part of OpenSSL 4.0 in April 2026. However, we have been working with OpenSSL maintainers on the so-called ["ECH feature branch"](https://github.com/openssl/openssl/tree/feature/ech) and that branch (subject to the same OpenSSL maintainer approval process as the OpenSSL master branch) now includes sufficient ECH code for web servers like apache2. So there's plenty of time for this PR to be discussed, but starting now may be timely.
- This PR includes [documentation](https://github.com/sftcd/httpd/blob/ECH-shared/docs/ECH-build.md) in a markdown document in the repo's top directory, which is certainly the wrong place, but may be useful short-term. That describes how to do the build, configuration and logging changes, and the code changes for ECH. (So should be a good place for reviewers to start.)
- While OpenSSL releases do not yet include ECH support, some other TLS libraries do, in particular boringssl. If useful, we could extend this PR to also support boringssl, or that could be a follow-up. (It'd be good if the server configuration were the same regardless of the TLS library.)
- ECH support using these OpenSSL ECH APIs was [included in the ligthttpd](https://redmine.lighttpd.net/projects/lighttpd/wiki/TLS_ECH) web server (in January 2025) so some code and patterns are common with that. We also plan to submit similar PRs to apache2 and haproxy, and ideally all would share some commonality.
- We have proposed a very similar PR for the NGINX project (https://github.com/nginx/nginx/pull/840).
- All that said, we're not fixated at all on things being done this way, and would be happy to make whatever changes are desired for apache2 and there are some notes on potential changes in [the documentation](https://github.com/sftcd/httpd/blob/ECH-shared/docs/ECH-build.md).

Lastly, for open-ness, our work on this has been funded by the Open Technology Fund (OTF) in the [DEfO project](https://defo.ie/).


notroj:
Thanks for the PR, have not reviewed but wanted to try hooking it up in CI: https://github.com/notroj/httpd/actions/runs/17064056660/job/48376952186

rpluem:
First short style nitpick: Please take care to put spaces around operands like `=`, `==` and `!=` when doing tests or assignments.

sftcd:
thanks for the CI setup! and that of course found a thing that's hopefully fixed in the above push - if there're more will similarly fix as they come

sftcd:
> First short style nitpick: Please take care to put spaces around operands like `=`, `==` and `!=` when doing tests or assignments.

will go over those in a bit, thanks

modules/ssl/ssl_engine_init.c
```diff
@@ -189,6 +197,122 @@ static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
...
+     *
+     * This code is derived from what I added to openssl s_server, which you can find
+     * in apps/s_server.c in my openssl fork, https://github.com/sftcd/openssl
+     */
+    if (echdir==NULL) {
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10520)
+                "load_echkeys: no directory name - exiting");
+        return -1;
+    }
+    size_t elen=strlen(echdir);
```
rpluem: We still try to keep some C89 compatibility for older compilers on more exotic platforms. Hence please move the declaration of `elen` at the top of the function.

modules/ssl/ssl_engine_config.c
```diff
@@ -222,6 +222,9 @@ static SSLSrvConfigRec *ssl_config_server_new(apr_pool_t *p)
 #endif
     sc->clienthello_vars       = UNSET;
     sc->session_tickets        = UNSET;
+#ifdef HAVE_OPENSSL_ECH
+    sc->echkeydir             = NULL;
```
rpluem: Whitespace Nit:
```suggestion
    sc->echkeydir              = NULL;
```

rpluem:
Can you please also add some documentation for the new directive `SSLECHKeyDir` to `docs/manual/mod/mod_ssl.xml`?
Can you please also add a new `txt` file to the `changes-entries directory that describes the change?

modules/ssl/ssl_engine_config.c
```diff
@@ -840,6 +846,20 @@ const char *ssl_cmd_SSLEngine(cmd_parms *cmd, void *dcfg, const char *arg)
     return "Argument must be On or Off";
 }
 
+#ifdef HAVE_OPENSSL_ECH
+const char *ssl_cmd_SSLECHKeyDir(cmd_parms *cmd, void *dcfg, const char *arg)
+{
+    SSLSrvConfigRec *sc = mySrvConfig(cmd->server);
+
+    sc->echkeydir=arg;
+    ap_log_error(APLOG_MARK, APLOG_TRACE4, 0, cmd->server, APLOGNO(10519)
```
rpluem: For `APLOG_TRACE?` logging we don't use `APLOGNO`s.

```suggestion
    ap_log_error(APLOG_MARK, APLOG_TRACE4, 0, cmd->server,
```

modules/ssl/ssl_engine_init.c
```diff
@@ -189,6 +197,122 @@ static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
...
+        struct stat thestat;
+        apr_finfo_t theinfo;
+        if ( (apr_stat (&theinfo, fname, APR_FINFO_MIN, ptemp)==APR_SUCCESS) ) {
+            keystried++;
+
+            BIO *in = BIO_new_file(fname, "r");
+            const int is_retry_config = OSSL_ECH_FOR_RETRY;
+            if (in != NULL
+                && 1 == OSSL_ECHSTORE_read_pem(es, in, is_retry_config)) {
+                ap_log_error(APLOG_MARK, APLOG_TRACE4, 0, s, APLOGNO(10524)
```
rpluem: For `APLOG_TRACE?` logging we don't use `APLOGNO`s.

```suggestion
                ap_log_error(APLOG_MARK, APLOG_TRACE4, 0, s,
```

modules/ssl/ssl_engine_init.c
```diff
@@ -574,8 +698,20 @@ static apr_status_t ssl_init_ctx_tls_extensions(server_rec *s,
      * protocol version(s) according to the selected (name-based-)vhost, which
      * is not possible at the SNI callback stage (due to OpenSSL internals).
      */
+#ifdef HAVE_OPENSSL_ECH
+    SSLSrvConfigRec *sc = mySrvConfig(s);
+    if (sc!=NULL && sc->echkeydir!=NULL) {
+        SSL_CTX_ech_set_callback(mctx->ssl_ctx, ssl_callback_ECH);
+    } else {
+        ap_log_error(APLOG_MARK, APLOG_TRACE4, 0, s, APLOGNO(10530)
```
rpluem: For `APLOG_TRACE?` logging we don't use `APLOGNO`s.

```suggestion
        ap_log_error(APLOG_MARK, APLOG_TRACE4, 0, s,
```

modules/ssl/ssl_engine_init.c
```diff
@@ -189,6 +197,122 @@ static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
...
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10521)
+                "load_echkeys: directory name too long: %s - exiting",echdir);
+        return -1;
+    }
+    int keystried=0;
+    int keysworked=0;
+
+    apr_dir_t *dir;
+    apr_finfo_t direntry;
+    apr_int32_t finfo_flags = APR_FINFO_TYPE|APR_FINFO_NAME;
```
rpluem: Same as above regarding C89.

modules/ssl/ssl_engine_init.c
```diff
@@ -189,6 +197,122 @@ static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
...
+                ap_log_error(APLOG_MARK, APLOG_INFO, 0, s, APLOGNO(10525)
+                    "load_echkeys: failed for %s (could be non-fatal)",fname);
+            }
+            BIO_free_all(in);
+        }
+
+    }
+    apr_dir_close(dir);
+
+    int keysloaded=0;
```
rpluem: Same as above regarding C89.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -1517,6 +1521,37 @@ int ssl_hook_Fixup(request_rec *r)
...
 
+#ifdef HAVE_OPENSSL_ECH
+    /*
+     * Add the ECH information to the environment 
+     */
+    char *inner_sni=NULL;
+    char *outer_sni=NULL;
+    char buf[PATH_MAX];
+    memset(buf,0,PATH_MAX);
+    int echrv=SSL_ech_get1_status((SSL*)ssl,&inner_sni,&outer_sni);
```
rpluem: Same as above regarding C89.

rpluem:
Please also document the `SSL_ECH_` environment variables in `docs/manual/mod/mod_ssl.xml`.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -1517,6 +1521,37 @@ int ssl_hook_Fixup(request_rec *r)
...
+        snprintf(buf,PATH_MAX,"success");
+        break;
+    default:
+        snprintf(buf,PATH_MAX, "error getting ECH status");
+    }
+    apr_table_set(env, "SSL_ECH_INNER_SNI", (inner_sni?inner_sni:"NONE"));
+    apr_table_set(env, "SSL_ECH_OUTER_SNI", (outer_sni?outer_sni:"NONE"));
+    apr_table_set(env, "SSL_ECH_STATUS", buf);
+
+#endif
```
rpluem: This whole stuff belongs to `ssl_engine_vars.c`. You should only amend `ssl_hook_Fixup_vars` here.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -2377,6 +2412,54 @@ static apr_status_t init_vhost(conn_rec *c, SSL *ssl, const char *servername)
...
+        ap_log_cerror(APLOG_MARK, APLOG_INFO, 0, c, APLOGNO(10536)
+            "ECH worked but bad name");
+        break;
+    case SSL_ECH_STATUS_SUCCESS:
+        ap_log_cerror(APLOG_MARK, APLOG_INFO, 0, c, APLOGNO(10537)
+            "ECH success outer_sni: %s inner_sni: %s",
+            (outer_sni?outer_sni:"NONE"),(inner_sni?inner_sni:"NONE"));
+        break;
+    default:
+        ap_log_cerror(APLOG_MARK, APLOG_INFO, 0, c, APLOGNO(10538)
```
rpluem: I think these should be DEBUG and not INFO messages.

modules/ssl/ssl_engine_init.c
sftcd: sorry, doing that kind of tidying up now generally

modules/ssl/ssl_engine_config.c
sftcd: ack

modules/ssl/ssl_engine_config.c
sftcd: ack, doing that generally

sftcd:
push above should tidy the code/address nits; documentation requested still TBD but coming soon:-)

modules/ssl/ssl_engine_config.c
sftcd: ack

rpluem:
> push above should tidy the code/address nits; documentation requested still TBD but coming soon:-)

Thanks. This just leaves the last two comments to modules/ssl/ssl_engine_kernel.c open (move code to ssl_engine_vars.c and loglevel discussion).

sftcd:
In the push above, I took a stab at adding documentation in the right place (`docs/manual/mod/mod_ssl.xml`) and building the resulting HTML which sort-of seems ok, but likely needs change. I've left the `ECH-build.md` file in the PR for now, as that still contains information that's not elsewhere at the moment. (We can ditch that once there's nothing left of value in the markdown file.)

sftcd:
> > push above should tidy the code/address nits; documentation requested still TBD but coming soon:-)
> 
> Thanks. This just leaves the last two comments to modules/ssl/ssl_engine_kernel.c open (move code to ssl_engine_vars.c and loglevel discussion).

Apologies for being slow (I often am:-) but I'm not seeing where those "last two comments" are - can you point me at them when you have a chance? Thanks.

rpluem:
> In the push above, I took a stab at adding documentation in the right place (`docs/manual/mod/mod_ssl.xml`) and building the resulting HTML which sort-of seems ok, but likely needs change. I've left the `ECH-build.md` file in the PR for now, as that still contains information that's not elsewhere at the moment. (We can ditch that once there's nothing left of value in the markdown file.)

Thanks. You can remove the HTML changes from PR as these will be created in a step separately from the commit. Just the changes to the xml file are required.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -1517,6 +1526,31 @@ int ssl_hook_Fixup(request_rec *r)
...
+        snprintf(buf, PATH_MAX, "success");
+        break;
+    default:
+        snprintf(buf, PATH_MAX, "error getting ECH status");
+    }
+    apr_table_set(env, "SSL_ECH_INNER_SNI", (inner_sni ? inner_sni : "NONE"));
+    apr_table_set(env, "SSL_ECH_OUTER_SNI", (outer_sni ? outer_sni : "NONE"));
+    apr_table_set(env, "SSL_ECH_STATUS", buf);
+#endif
+
```
rpluem: This whole stuff belongs to `ssl_engine_vars.c`. You should only amend `ssl_hook_Fixup_vars` here.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -2377,6 +2411,56 @@ static apr_status_t init_vhost(conn_rec *c, SSL *ssl, const char *servername)
...
+        break;
+    case SSL_ECH_STATUS_SUCCESS:
+        ap_log_cerror(APLOG_MARK, APLOG_INFO, 0, c, APLOGNO(10537)
+            "ECH success outer_sni: %s inner_sni: %s",
+            (outer_sni ? outer_sni : "NONE"),
+            (inner_sni ? inner_sni : "NONE"));
+        break;
+    default:
+        ap_log_cerror(APLOG_MARK, APLOG_INFO, 0, c, APLOGNO(10538)
+            "Error getting ECH status");
```
rpluem: I think these should be DEBUG and not INFO messages.

rpluem:
> > > push above should tidy the code/address nits; documentation requested still TBD but coming soon:-)
> > 
> > 
> > Thanks. This just leaves the last two comments to modules/ssl/ssl_engine_kernel.c open (move code to ssl_engine_vars.c and loglevel discussion).
> 
> Apologies for being slow (I often am:-) but I'm not seeing where those "last two comments" are - can you point me at them when you have a chance? Thanks.

I just readded the comments to the latest version of the code as I was not sure how I should point you to them.

changes-entries/ech.txt
```diff
@@ -0,0 +1,3 @@
+
+  *) mod_ssl: Add support for Encrypted Client Hello (ECH)
+     Github #551 [Stephen Farrell]
```
rpluem: People who are not committers to this project are typically credited with name and email address (unless you don't want to give yours), hence a pattern like [John Doe <john.doe@example.com>].

docs/manual/mod/mod_ssl.xml
```diff
@@ -3016,4 +3019,122 @@ httpd -t -D DUMP_SSL_POLICIES
 </usage>
 </directivesynopsis>
 
+<directivesynopsis>
+<name>SSLECHKeyDir</name>
+<description>Load the set of Encrypted Client Hello (ECH) PEM files in the named directory</description>
+<syntax>SSLECHKeyDir <em>dirname</em></syntax>
+<contextlist><context>server config</context></contextlist>
+<compatibility>Available in httpd 2.X.0 and later</compatibility>
```
rpluem:
```Suggestion
<compatibility>Available in Apache HTTP Server 2.5.1 and later</compatibility>
```

sftcd:
If I've done it right the (force) push above just removes the documentation files that I shouldn't have included yesterday and make no other change

sftcd:
> I just readded the comments to the latest version of the code as I was not sure how I should point you to them.

I've no idea why, but I'm not seeing any comments at all on `ssl_engine_kernel.c` sorry

rpluem:
> > I just readded the comments to the latest version of the code as I was not sure how I should point you to them.
> 
> I've no idea why, but I'm not seeing any comments at all on `ssl_engine_kernel.c` sorry

Very strange. I thought that my comments might get removed after another commit or on a force push, but I still can see them on the files tab. There are currently 4 open.

rpluem:
I think I know what I did wrong: I forgot to submit my review and hence it was only visible to me.

notroj:
Updated CI run https://github.com/notroj/httpd/actions/runs/17120295279/job/48559395252#step:11:2461

modules/ssl/ssl_engine_init.c
```diff
@@ -189,6 +189,112 @@ static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
...
+    if (es == NULL) {
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10523)
+                "load_echkeys: can't alloc store");
+        return -1;
+    }
+
+    while ((apr_dir_read(&direntry, finfo_flags, dir)) == APR_SUCCESS) {
+        const char *fname;
+        size_t pnlen = 0;
+        struct stat thestat;
```
rpluem:
```suggestion
```

rpluem (changes requested):

sftcd:
> I think I know what I did wrong: I forgot to submit my review and hence it was only visible to me.

Thanks for those, I'll process 'em shortly. (And am relieved I wasn't doing something wrong:-)

modules/ssl/ssl_engine_init.c
sftcd: yep fixed that

modules/ssl/ssl_engine_init.c
sftcd: also fixed that

modules/ssl/ssl_engine_init.c
sftcd: should be fixed now too

modules/ssl/ssl_engine_init.c
sftcd: also fixed

modules/ssl/ssl_engine_kernel.c
sftcd: also fixed

modules/ssl/ssl_engine_kernel.c
sftcd: sure, happy to take guidance on what level's best, so changed them to `APLOG_DEBUG`

modules/ssl/ssl_engine_kernel.c
sftcd: so I've tried to do that, but not sure I'm understanding the reference to `ssl_hook_Fixup_vars` so might've done it wrong, but it should be closer at least. I also need to run this version in a test setup with PHP, which I've not done yet, (will later today, when on a different n/w where that's easier for me), so I might have more to change when I do that.

modules/ssl/ssl_engine_kernel.c
sftcd: yep, changed to that

changes-entries/ech.txt
sftcd: no problem doing that so I added an email

docs/manual/mod/mod_ssl.xml
sftcd: sure, also tweaked the example using `dig`

modules/ssl/ssl_engine_init.c
sftcd: indeed!

sftcd:
The push above tries to address today's @rpluem comments.

notroj:
Can you select the box to allow edits by maintainers? I can then push my CI change to your fork

sftcd:
> Can you select the box to allow edits by maintainers? I can then push my CI change to your fork

That was turned on already. (Assuming it's "Allow edits and access to secrets by maintainers" on the PR) But I turned if off and on again, so maybe that's helped?

Below though it says "No applicable reviews submitted by reviewers with write access" so maybe something else needed?

sftcd:
Push above is a change so that the CGI/PHP vars stuff now seems to work. 
Accessing the CGI/PHP vars didn't work before this change.
But while it works, the change to make it work looks a bit wrong. 
Advice welcome.

changes-entries/ech.txt
```diff
@@ -0,0 +1,3 @@
+
+  *) mod_ssl: Add support for Encrypted Client Hello (ECH)
+     Github #551 [Stephen Farrell stephen.farrell@cs.tcd.ie]
```
rpluem:
```suggestion
     Github #551 [Stephen Farrell <stephen.farrell@cs.tcd.ie>]
```

modules/ssl/ssl_engine_kernel.c
rpluem: Currently it is not the way I  would like to see it, but the way I told you it should look like is probably a bit wrong as well :-). The core thing I would like to get to is that the 3 new variables can be queried through the central `ssl_var_lookup` function from `ssl_engine_vars.c` to enable other possible consumers of these variables to get its values. Once this is possible you can add them to the environment in `ssl_engine_kernel.c::ssl_hook_Fixup` via

```c
#ifdef HAVE_OPENSSL_ECH
extract_to_env(r, env, "SSL_ECH_INNER_SNI");
extract_to_env(r, env, "SSL_ECH_OUTER_SNI");
extract_to_env(r, env, "SSL_ECH_STATUS_SNI");
#endif
```

You would implement your extraction in `ssl_engine_vars.c::ssl_var_lookup_ssl` and would add cases for `ECH_INNER_SNI`, `ECH_OUTER_SNI` and `ECH_STATUS`. The `SSL_` prefix gets already striped by  `ssl_engine_vars.c::ssl_var_lookup`  when it calls  `ssl_engine_vars.c::ssl_var_lookup_ssl`. Keep in mind that you have to `apr_pstrdup()` the strings in this case here before returning them. For example have a look how `SSL_SESSION_RESUMED` or better `SESSION_RESUMED` is  currently implemented (https://github.com/apache/httpd/blob/f09577aa96822cf2fee8b3e9f8829bd6eabfbe1d/modules/ssl/ssl_engine_vars.c#L468).

rpluem (changes requested):

changes-entries/ech.txt
sftcd: that's easily done:-)

sftcd:
> Currently it is not the way I would like to see it, but the way I told you it should look like is probably a bit wrong as well :-).

Heh:-) I think I see what you mean now, so will take a stab at that...

rpluem:
> > Currently it is not the way I would like to see it, but the way I told you it should look like is probably a bit wrong as well :-).
> 
> Heh:-) I think I see what you mean now, so will take a stab at that...

Great :-). Like with `SESSION_RESUMED` you don't need to do any kind of copying of the literal strings you return for `SSL_ECH_STATUS`. This is only needed for the `_SNI` variables via `apr_pstrdup()` if their contents is provided via OpenSSL functions.

sftcd:
Push above tries to do the `ssl_engine_vars.c` thing better, and seems to work ok locally. How's that?

Ah - didn't spot your most recent comment, so guess I'm copying too much. Gotta go afk now for the evening though so will improve tomorrow. (Sat)

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+static const char *ssl_var_lookup_ech_status(apr_pool_t *p, const char *var, SSL *ssl)
+{
+    char *inner_sni = NULL, *outer_sni = NULL;
+    int echrv;
+    const char *result = NULL;
+
+    if (ssl == NULL)
+        return "";
+    echrv = SSL_ech_get1_status(ssl, &inner_sni, &outer_sni);
+    if(strcEQ(var, "ECH_STATUS")) {
```
rpluem:
```suggestion
    if (strcEQ(var, "STATUS")) {
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+            break;
+        case SSL_ECH_STATUS_SUCCESS:
+            result = apr_pstrdup(p, "success");
+            break;
+        default:
+            result = apr_pstrdup(p, "error getting ECH status");
+        }
+    }
+    else if (echrv == SSL_ECH_STATUS_SUCCESS) {
+        if(strcEQ(var, "ECH_INNER_SNI")) {
```
rpluem:
```suggestion
        if (strcEQ(var, "INNER_SNI")) {
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+            result = apr_pstrdup(p, inner_sni);
+        }
+        if(strcEQ(var, "ECH_OUTER_SNI")) {
```
rpluem:
```suggestion
        if (strcEQ(var, "OUTER_SNI")) {
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+            result = apr_pstrdup(p, outer_sni);
+        }
+    }
+    OPENSSL_free(inner_sni);
+    OPENSSL_free(outer_sni);
+    return result ? result : "";
```
rpluem: If nothing was found leave it at `NULL`.
```suggestion
    return result;
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+    int echrv;
+    const char *result = NULL;
+
+    if (ssl == NULL)
+        return "";
+    echrv = SSL_ech_get1_status(ssl, &inner_sni, &outer_sni);
+    if(strcEQ(var, "ECH_STATUS")) {
+        switch (echrv) {
+        case SSL_ECH_STATUS_NOT_TRIED:
+            result = apr_pstrdup(p, "not attempted");
```
rpluem:
```suggestion
            result ="not attempted";
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+            break;
+        case SSL_ECH_STATUS_FAILED:
+            result = apr_pstrdup(p, "tried but failed");
```
rpluem:
```suggestion
            result = "tried but failed";
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+            break;
+        case SSL_ECH_STATUS_BAD_NAME:
+            result = apr_pstrdup(p, "ECH worked but bad name");
```
rpluem:
```suggestion
            result = "ECH worked but bad name";
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+            break;
+        case SSL_ECH_STATUS_SUCCESS:
+            result = apr_pstrdup(p, "success");
```
rpluem:
```suggestion
            result = "success";
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+            break;
+        default:
+            result = apr_pstrdup(p, "error getting ECH status");
```
rpluem:
```suggestion
            result = "error getting ECH status";
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -430,10 +430,54 @@ const char *ssl_var_lookup(apr_pool_t *p, server_rec *s,
...
+#ifdef HAVE_OPENSSL_ECH
+/* Extract ECH status variable from SSL object 'ssl' */
+static const char *ssl_var_lookup_ech_status(apr_pool_t *p, const char *var, SSL *ssl)
+{
+    char *inner_sni = NULL, *outer_sni = NULL;
+    int echrv;
+    const char *result = NULL;
+
+    if (ssl == NULL)
+        return "";
```
rpluem: If nothing was found leave it at `NULL`.
```suggestion
        return result;
```

modules/ssl/ssl_engine_vars.c
```diff
@@ -544,6 +588,17 @@ static const char *ssl_var_lookup_ssl(apr_pool_t *p, const SSLConnRec *sslconn,
...
+#ifdef HAVE_OPENSSL_ECH
+    else if(ssl != NULL && strcEQ(var, "ECH_INNER_SNI")) {
+        result = ssl_var_lookup_ech_status(p, var, ssl);
+    }
+    else if(ssl != NULL && strcEQ(var, "ECH_OUTER_SNI")) {
+        result = ssl_var_lookup_ech_status(p, var, ssl);
+    }
+    else if(ssl != NULL && strcEQ(var, "ECH_STATUS")) {
+        result = ssl_var_lookup_ech_status(p, var, ssl);
+    }
```
rpluem:
```suggestion
    else if (ssl != NULL && strcEQn(var, "ECH_", 4)) {
        result = ssl_var_lookup_ech_status(p, var+4, ssl);
    }
```

rpluem (changes requested):

modules/ssl/ssl_engine_vars.c
sftcd: sure, will do (and same for all in this set of suggestions below)

modules/ssl/ssl_engine_vars.c
sftcd: sure

modules/ssl/ssl_engine_vars.c
sftcd: sure

modules/ssl/ssl_engine_vars.c
sftcd: sure

modules/ssl/ssl_engine_vars.c
sftcd: sure

modules/ssl/ssl_engine_vars.c
sftcd: sure

modules/ssl/ssl_engine_vars.c
sftcd: sure

modules/ssl/ssl_engine_vars.c
sftcd: sure

modules/ssl/ssl_engine_vars.c
sftcd: sure

modules/ssl/ssl_engine_vars.c
sftcd: sure

modules/ssl/ssl_engine_vars.c
sftcd: sure

sftcd:
push above makes yesterday's suggested changes to `ssl_engine_vars.c`

rpluem:
@notroj : Code looks good to me now. Any chance that you can get it run in CI again? Apart from this any comments from your side?
@icing : Any comments on the code from your side?

icing:
> @notroj : Code looks good to me now. Any chance that you can get it run in CI again? Apart from this any comments from your side? @icing : Any comments on the code from your side?

Not an expert on `mod_ssl` code to have a serious opinion here.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -2377,6 +2386,58 @@ static apr_status_t init_vhost(conn_rec *c, SSL *ssl, const char *servername)
...
+    int echrv;
+    conn_rec *c = NULL;
+    const char *ech_servername;
+    apr_status_t ivstatus;
+
+    c = (conn_rec *)SSL_get_app_data(ssl);
+    ech_servername = SSL_get_servername(ssl, TLSEXT_NAMETYPE_host_name);
+    if (ech_servername == NULL) {
+        return SSL_TLSEXT_ERR_NOACK;
+    }
```
ylavic, line 2401: Does `SSL_get_servername()` return the inner_sni when ECH is active? Don't we need to check that both match here?

Also, we have some code in httpd that depends on the SNI being readable at the earliest of TLS processing (in `ssl_callback_ClientHello()`, see 076e283) to be able to replace the SSL_CTX attached to an SSL (from the one of the first VirtualHost initially), before the negotiation is really starting (this is needed for multiple vhosts on the same IP:port to be able to use different TLS protocol versions for instance).
So I wonder if `ssl_callback_ClientHello()` is still called in OpenSSL-4?
If yes, is `ssl_callback_ClientHello()` called before `ssl_callback_ECH()` here still?
If no, does `ssl_callback_ECH()` run early enough to be able to `SSL_set_SSL_CTX(c->SSL, s->SSL_CTX)` below in `init_vhost()`?

notroj:
I still got permission denied trying to push to sftcd:ECH-shared via ssh so maybe there is some other restriction here, oh well.

Latest run: https://github.com/notroj/httpd/actions/runs/17236024872

modules/ssl/ssl_engine_init.c
```diff
@@ -189,6 +189,111 @@ static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
...
+                        apr_pool_t *ptemp)
+{
+    size_t elen = 0;
+    int keystried = 0, keysworked = 0, keysloaded=0;
+    OSSL_ECHSTORE *es = NULL;
+    apr_dir_t *dir = NULL;
+    apr_finfo_t direntry;
+    apr_int32_t finfo_flags = APR_FINFO_TYPE|APR_FINFO_NAME;
+
+    if (echdir == NULL) {
```
notroj: Checking for `echdir!=NULL` is redundant since the calling function already checks that, so you can document it as a constraint for the interface.

modules/ssl/ssl_engine_init.c
```diff
@@ -189,6 +189,111 @@ static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
...
+                "load_echkeys: no directory name - exiting");
+        return -1;
+    }
+    elen = strlen(echdir);
+    if ((elen + 7) >= PATH_MAX) {
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10521)
+                "load_echkeys: directory name too long: %s - exiting", echdir);
+        return -1;
+    }
+    if (!echdir || (apr_dir_open(&dir, echdir, ptemp) != APR_SUCCESS)) {
```
notroj: If `apr_dir_open` fails catch the returned `apr_status_t` and pass it to the `ap_log_error` so the actual error is logged

modules/ssl/ssl_engine_init.c
```diff
@@ -903,7 +1022,39 @@ static apr_status_t ssl_init_ctx_protocol(server_rec *s,
...
+            SSL_CTX_free(ctx);
+            mctx->ssl_ctx = NULL;
+            return ssl_die(s);
+        } 
+    }
+
+#else
+    if (sc->echkeydir) {
+        ap_log_error(APLOG_MARK, APLOG_WARN, 0, s, APLOGNO(10533)
+                 "ECHKeyDir configured but no TLSv1.3 so ECH will be ignored.");
```
notroj: IMO if TLSv1.3 is not supported by the TLS toolkit, using `SSLECHKeyDir` should be a config-time error, so merely warning for this is not sufficient. So in the definition of `HAVE_OPENSSL_ECH` in ssl_private.h check that SSL_HAVE_PROTOCOL_TLSV1_3 definition too, and drop this case?

notroj: Alternative: in `ssl_cmd_SSLECHKeyDir` check that SSL_HAVE_PROTOCOL_TLSV1_3 is defined and if not always return an error string - this would give better feedback than simply getting an error that the directive isn't supported.

modules/ssl/ssl_engine_init.c
```diff
@@ -189,6 +189,111 @@ static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
...
+    }
+    if (!echdir || (apr_dir_open(&dir, echdir, ptemp) != APR_SUCCESS)) {
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10522)
+                "load_echkeys: can't open directory %s - exiting", echdir);
+        return -1;
+    }
+    es = OSSL_ECHSTORE_new(NULL, NULL);
+    if (es == NULL) {
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10523)
+                "load_echkeys: can't alloc store");
```
ylavic, line 226: `apr_dir_close(dir);`?

modules/ssl/ssl_engine_init.c
```diff
@@ -189,6 +189,111 @@ static void ssl_add_version_components(apr_pool_t *ptemp, apr_pool_t *pconf,
...
+            }
+            BIO_free_all(in);
+        }
+
+    }
+    apr_dir_close(dir);
+
+    if (!OSSL_ECHSTORE_num_keys(es, &keysloaded)) {
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10526)
+            "OSSL_ECHSTORE_num_keys failed - exiting");
```
ylavic, line 275: `OSSL_ECHSTORE_free(es);`?

modules/ssl/ssl_engine_init.c
```diff
@@ -574,8 +682,19 @@ static apr_status_t ssl_init_ctx_tls_extensions(server_rec *s,
...
      * is not possible at the SNI callback stage (due to OpenSSL internals).
      */
+
+#ifdef HAVE_OPENSSL_ECH
+    if (sc != NULL && sc->echkeydir != NULL) {
+        SSL_CTX_ech_set_callback(mctx->ssl_ctx, ssl_callback_ECH);
+    } else {
+        ap_log_error(APLOG_MARK, APLOG_TRACE4, 0, s,
+            "ECHKeyDir not set - using ClientHello callback for SNI");
+        SSL_CTX_set_client_hello_cb(mctx->ssl_ctx, ssl_callback_ClientHello, NULL);
```
ylavic: This `} else {` looks unnecessary if you leave the `SSL_CTX_set_client_hello_cb()` below out of `#else /* HAVE_OPENSSL_ECH */`?

rpluem: But shouldn't we set only one, `SSL_CTX_ech_set_callback` or `SSL_CTX_set_client_hello_cb`but not both at the same time?

rpluem: Answering myself: Scratch the above. You answered it to me in a later comment and why we do `SSL_CTX_set_client_hello_cb` since 076e28399c7336f2b287b102a6e4e40934f2057d if possible. Hence we should always call `SSL_CTX_set_client_hello_cb` independent of `HAVE_OPENSSL_ECH`.

modules/ssl/ssl_engine_kernel.c
sftcd: Good questions! It might help to say as a preamble that ECH decryption happens very very early (as soon as the ClientHello is rx'd) and if ECH decryption succeeded, then the outer ClientHello content is pretty much discarded (other than remembering the `public_name` used) - that's basically to reduce the already large scale of code changes inside the TLS library. (IIRC the same is true of boringssl, not sure about other libraries)

given the above: `SSL_get_servername()` will return the inner SNI so that applications aren't affected by turning on ECH

I believe (need to re-confirm as it's been ages since I looked at that code) that ECH decryption will have worked or failed before any callbacks are called, and so `ssl_callback_ClientHello()` should continue to work as before. In another server that still had it's own code directly decoding the ClientHello I did need to make more changes but anything depending on OpenSSL APIs should be good without change. 

I'm not sure in which order the callbacks will be hit if both are set, but will check.

sftcd:
> I still got permission denied trying to push to sftcd:ECH-shared via ssh so maybe there is some other restriction here, oh well.

That's odd. Maybe if one of you close then re-open the PR that'd do something? Or perhaps you don't care:-) Either's fine by me.

modules/ssl/ssl_engine_init.c
sftcd: happy to do that if that's the local preference

modules/ssl/ssl_engine_init.c
sftcd: good idea

modules/ssl/ssl_engine_init.c
sftcd: Hmm. I agree having some cases handled as warnings (as above) and others as fatal is a bit off and all should be handled at the same error level. (For the moment I'll do s/WARN/EMERG/ just to be consistent and we can fix better based on the conclusion of this point.)

I guess one could take either of two positions though:

1) ECH is an optional privacy improvement, already vulnerable to some active attacks (on the DNS), and (some) clients/browsers might in any case fallback to non-ECH (in a 2nd connection) so logging but continuing in the face of misconfiguration is better, or,
2) if ECH is configured on and the server can't actually do ECH, then that should be treated as fatal so the server admin gets to know about and fix things (which could be the server config, or the server s/w etc.)

I could be convinced either way;-)

Is there any guidance from how other misconfigurations are handled?

modules/ssl/ssl_engine_init.c
sftcd: ta

modules/ssl/ssl_engine_init.c
sftcd: also ta

modules/ssl/ssl_engine_init.c
sftcd: So I guess neither `ssl_callback_ECH()` nor `ssl_callback_ClientHello()` are that well named - something like `ssl_callback_set_vhost_from_[ClientHello|ECH]()` might be better. But that said, I'm not sure if the `#else` is better left in - will get back when I've re-checked which callback fires when in which cases.

sftcd:
push above has changes from various comments today, with two remaining things to for me to check/to-discuss - how best to handle misconfiguration and the ordering of callbacks (I'll do the checking Wed sometime and get back)

modules/ssl/ssl_engine_init.c
sftcd: Having thought about that, I figure the better thing to do is number 2 above - treat all these errors as fatal and don't start the server (so use `APLOG_EMERG`) - what convinced me was that if you start an old server that doesn't have any ECH support, and that sees an `SSLECHKeyDir` directive then that old server will fail to start, so doing likewise in these cases seems correct. So I think the code now in the PR does the right thing.

@notroj - is there a reason to do this in `ssl_cmd_SSLECHKeyDir` or is it ok as-is?

modules/ssl/ssl_engine_init.c
```diff
@@ -903,7 +1023,39 @@ static apr_status_t ssl_init_ctx_protocol(server_rec *s,
...
+                 "ECHKeyDir configured but TLSv1.3 turned off - exiting.");
+            SSL_CTX_free(ctx);
+            mctx->ssl_ctx = NULL;
+            return ssl_die(s);
+        } 
+    }
+
+#else
+    if (sc->echkeydir) {
+        ap_log_error(APLOG_MARK, APLOG_EMREG, 0, s, APLOGNO(10533)
```
rpluem:
```suggestion
        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10533)
```

modules/ssl/ssl_engine_init.c
```diff
@@ -903,7 +1023,39 @@ static apr_status_t ssl_init_ctx_protocol(server_rec *s,
...
+                 "ECHKeyDir configured but TLSv1.3 not supported.");
```
rpluem:
```suggestion
                 "ECHKeyDir configured but TLSv1.3 not supported - exiting..");
            SSL_CTX_free(ctx);
            mctx->ssl_ctx = NULL;
            return ssl_die(s);
```

rpluem (changes requested):

modules/ssl/ssl_engine_init.c
```diff
@@ -903,7 +1023,42 @@ static apr_status_t ssl_init_ctx_protocol(server_rec *s,
...
+        } 
+    }
+
+#else
+    if (sc->echkeydir) {
+        ap_log_error(APLOG_MARK, APLOG_EMERG, 0, s, APLOGNO(10533)
+                 "ECHKeyDir configured but TLSv1.3 not supported - exiting..");
+            SSL_CTX_free(ctx);
+            mctx->ssl_ctx = NULL;
+            return ssl_die(s);
```
rpluem: Sorry my bad in my original suggestion, but space police :-(
```suggestion
        SSL_CTX_free(ctx);
        mctx->ssl_ctx = NULL;
        return ssl_die(s);
```

rpluem (changes requested):

modules/ssl/ssl_engine_init.c
sftcd: So I added a [new ECH test](https://github.com/openssl/openssl/pull/28270/commits/9702c985c6e7aa4d7f844180e7e8868828c12477) to my current PR for the OpenSSL feature branch that basically mimics what `ssl_callback_ClientHello()` does. 

The result is: the client_hello callback fires first, and has the inner CH, after ECH decryption has worked, so all should be well regardless of which callback is used. The ECH callback doesn't fire until the ServerHello message is being prepared (IIRC it has to be that way in case of HRR, as the ECH callback shouldn't fire until HRR and ECH have both succeeded.)

So I guess we could get rid of `ssl_callback_ECH()` entirely, but I wonder what's the correct thing for the server to do when HRR happens?

modules/ssl/ssl_engine_init.c
notroj: Not sure it's written down as guidance but yes your (2) is generally preferred, if the configuration cannot be applied, fail at config time. It's better to do it in `ssl_cmd_SSLECHKeyDir` so that `httpd -t` will fail for the configuration.

sftcd: ah right, will make that change so

sftcd: Just pushed a change for that, let me know if more needed - the commit is [here](https://github.com/apache/httpd/pull/551/commits/5a4db6c1b0412260ea28435ccf34664a1ccb5ef3)

modules/ssl/ssl_engine_init.c
sftcd: I did a test build getting rid of `ssl_callback_ECH()` and just using `ssl_callback_ClientHello()` and it seems to work fine, even though the callback is hit twice when we hit HRR, but I didn't seen any bad side-effects of that, and it's what happens already so is presumably ok. I guess that's less change so unless there's a reason to keep `ssl_callback_ECH()` I'll make that change in a while.

sftcd: Oops - we still want `ssl_callback_ECH()` to log the ECH outcome, but the vhost stuff can now be left to the CH callback.

sftcd:
push above simplifies the use of callbacks as per discussion above

modules/ssl/ssl_engine_init.c
sftcd: push below (https://github.com/apache/httpd/pull/551/commits/50e1f81c8bc391551af1562fb1c31ea4f2402e6d) makes these changes

modules/ssl/ssl_engine_kernel.c
sftcd: this has been addressed in discussion below and  https://github.com/apache/httpd/pull/551/commits/50e1f81c8bc391551af1562fb1c31ea4f2402e6d

notroj:
https://github.com/notroj/httpd/actions/runs/17288258105/job/49069552136#step:11:2463
```
ssl_engine_kernel.c: In function 'ssl_callback_ECH':
ssl_engine_kernel.c:2396:18: error: unused variable 'ivstatus' [-Werror=unused-variable]
 2396 |     apr_status_t ivstatus;
      |                  ^~~~~~~~
cc1: all warnings being treated as errors
```

sftcd:
push above removes ivstatus (thanks for re-doing the CI run)

modules/ssl/ssl_engine_init.c
ylavic: > it seems to work fine, even though the callback is hit twice when we hit HRR, but I didn't seen any bad side-effects of that, and it's what happens already so is presumably ok.

If ssl_callback_ClientHello() can be called multiple times with HRR, maybe we should not let its init_vhost() be a noop for the second time (should the SNI be chang{able,ed} in between..)? Maybe for another PR/commit though since it's not really related to ECH.

> Oops - we still want `ssl_callback_ECH()` to log the ECH outcome, but the vhost stuff can now be left to the CH callback.

Agreed.

modules/ssl/ssl_engine_kernel.c
```diff
@@ -2447,7 +2499,17 @@ int ssl_callback_ClientHello(SSL *ssl, int *al, void *arg)
...
+
+#ifdef HAVE_OPENSSL_ECH
+    if (SSL_client_hello_get0_ext(ssl, TLSEXT_TYPE_ech, &pos, &remaining)) {
+        ap_log_cerror(APLOG_MARK, APLOG_INFO, 0, c, APLOGNO(10540)
+                      "there is an ECH extension");
+    } else {
+        ap_log_cerror(APLOG_MARK, APLOG_INFO, 0, c, APLOGNO(10541)
+                      "there is NO ECH extension");
+    }
+#endif
```
ylavic: Not sure this is needed since ssl_callback_ECH() will log the status anyway?
And I don't think we want this to be logged at APLOG_INFO for every connection, APLOG_DEBUG/TRACEn would be enough IMHO.

modules/ssl/ssl_engine_init.c
sftcd: > If ssl_callback_ClientHello() can be called multiple times with HRR, maybe we should not let its init_vhost() be a noop for the second time (should the SNI be chang{able,ed} in between..)? Maybe for another PR/commit though since it's not really related to ECH.

Agree that that'd  be for a separate PR. I didn't notice any leak in the HRR case via valgrind fwiw. The TLSv1.3 spec says how the 2nd CH can differ from the first (at https://datatracker.ietf.org/doc/html/rfc8446#section-4.1.2) and there's some text too about that in the ECH spec (as the 2nd CH uses the same HPKE context). Those rules do not allow for changing the SNI between the 1st and 2nd CH's, but of course a bad client might cheat. I'm not sure how or if different TLS libraries make those checks, nor whether there'd be information on that available to applications. When the OpenSSL client hello callback fires, I don't believe you'd know if HRR will happen or not though as the CH hasn't even been parsed, so the library won't have noticed that the key shares are such as to cause HRR. If there's some setup where the `init_vhost()` can cause side-effects then I guess this would be worth trying to better handle, but I don't know if that's the case.

Sounds like creating an issue for httpd for this is probably right? I can do that but it might be better done by one of you who understand the server code better. (let me know)

modules/ssl/ssl_engine_kernel.c
sftcd: yeah, that's fair - I guess I added that while figuring stuff out;-)

it's removed in the push below

rpluem:
@notroj : Can you start another probably final CI build? I think all questions have been addressed.

notroj:
Sorry I did already forgot to mention - all green! https://github.com/notroj/httpd/actions/runs/17299086256

notroj:
Possibly the most reviewed PR ever... I merged this with minor syntax/style changes only. I didn't include the new doc file but included the relevant parts of that text in the commit message and I think most of the rest is covered in the docs.

Many thanks @sftcd !

sftcd:
That's great thanks! It was a pleasure getting/handling those comments. And do ping if/as ECH issues arise, happy to help out as I can.

PR #552, opened 2025-08-26
Merge r1927885, r1927916 from trunk

rpluem:
Merge r1927885, r1927916 from trunk


rpluem:
I miss a documentation and changes entry.

PR #556, opened 2025-09-04
mod_dav: optionally set mtime against x-oc-mtime header

leo9800:
WebDAV RFCs did not specify a standard approach to set modification time for files or directory on the server, this feature enables better presevation of file metadata when uploading.

as a compensation, OwnCloud (now known as NextCloud) proposed a `X-OC-Mtime` header which contains UNIX timestamp of mtime with file-modifying methods (e.g. `PUT`) which implements the feature above as a extension to the standard.

this patch allows apache to optionally honor the `X-OC-Mtime` header and set mtime accordingly if `DavHonorMtimeHeader On` is set.


modules/dav/main/mod_dav.c
```diff
@@ -1255,7 +1266,13 @@ static int dav_method_put(request_rec *r)
 
         err2 = (*resource->hooks->close_stream)(stream,
                                                 err == NULL /* commit */);
+
+        if (err == NULL && mtime_ret == 1) {
```
notroj, line 1277: Need to check the ->set_mtime hook is non-NULL here too?

modules/dav/main/mod_dav.c
```diff
@@ -1162,6 +1165,14 @@ static int dav_method_put(request_rec *r)
         mode = DAV_MODE_WRITE_TRUNC;
     }
 
+    /* try parsing x-oc-mtime header */
+    if ((mtime_ret = dav_parse_mtime(r, &mtime)) == -1) {
```
notroj: The header should be completely ignored if the config directive is not enabled.

modules/dav/fs/repos.c
```diff
@@ -1164,6 +1164,20 @@ static dav_error * dav_fs_deliver(const dav_resource *resource,
...
+static dav_error * dav_fs_set_mtime(dav_resource *resource, apr_time_t mtime)
+{
+    apr_pool_t *pool;
+    apr_status_t status;
+
+    pool = resource->pool;
+    status = apr_file_mtime_set(resource->info->pathname, mtime, pool);
+
+    if (status != APR_SUCCESS) {
+        return dav_new_error(pool, HTTP_BAD_REQUEST, 0, status, "Could not set mtime.");
```
notroj: If we get this far this is likely not a client error so this should be mapped to something appropriate rather than a 400. Or default to 500?

notroj (commented):
This is an interesting feature. Are you aware of any standardisation effort for that header?

leo9800:
> This is an interesting feature. Are you aware of any standardisation effort for that header?

seems there is not much such efforts (i.e. RFC drafts) from what i know.

it is an approach OwnCloud (predecessor of NextCloud) adopts to enable `mtime` presevation when uploading file via WebDAV as RFCs of WebDAV does not define a standard way to do so.

this feature is also implemented by rclone's `rclone serve webdav` WebDAV server command: https://github.com/rclone/rclone/blob/3f0e9f5fca8080f87094341f8ac5027091cc7601/cmd/serve/webdav/webdav.go#L339

given that this is a non-standard, i would implement it but keep it disabled by default to avoid any non-standard breaking change. users who prefer to use this feature should have it opted in manually with `DAVHonorMtimeHeader On`.

modules/dav/main/mod_dav.c
leo9800: yes you are correct, it was my sloppy work. gonna fix it soon.

modules/dav/main/mod_dav.c
leo9800: yes again you are correct, some providers, maybe in future and/or in private 3rd party implementations, dav may be backed by some kind of storages that could not storage modification time metadata, where `set_mtime` callback should be NULL.

I would prefer log a debug level message, do nothing (related to setting mtime) and succeed the request in such cases. or should the request fail and 5xx response code being returned?

modules/dav/fs/repos.c
leo9800: actually i was a bit unsure whether 4xx or 5xx should be returned in such case. despite most invalid patterns has already being sanitized in `dav_parse_mtime`, including:

- non-digits, e.g. `nonsense   bytes`
- negatives, e.g. `-123123123`
- non-decimal, e.g. `0xa0b1`
- empty strings
- overflowing

while `apr_file_mtime_set` may still fail, in most cases it is clear that a 500 should be returned, e.g. kernel VFS failed to respond `utime` syscall. but i was wondering if there is any possibility that an invalid user input fails `apr_file_mtime_set` where 400 should be returned.

as of now I would return a 500 in such cases instead of 400.

modules/dav/fs/repos.c
```diff
@@ -1164,6 +1164,25 @@ static dav_error * dav_fs_deliver(const dav_resource *resource,
...
+    apr_pool_t *pool;
+    apr_status_t status;
+    char errbuf[1024];
+
+    pool = resource->pool;
+    status = apr_file_mtime_set(resource->info->pathname, mtime, pool);
+
+    if (status != APR_SUCCESS) {
+        apr_strerror(status, errbuf, sizeof(errbuf));
+        ap_log_perror(APLOG_MARK, APLOG_ERR, status, pool, APLOGNO(10521)
```
notroj: Passing the status means the strerror handling is unnecessary, the `apr_log_perror` call will include the error text for that apr_status_t value

modules/dav/main/mod_dav.c
```diff
@@ -1207,7 +1272,25 @@ static int dav_method_put(request_rec *r)
 
         err2 = (*resource->hooks->close_stream)(stream,
                                                 err == NULL /* commit */);
+
+
+        mtime_aware = *resource->hooks->set_mtime != NULL;
+        if (err == NULL && conf->honor_mtime_header == DAV_ENABLED_ON && mtime_ret == 1) {
+            if (mtime_aware) {
+                err3 = (*resource->hooks->set_mtime)(resource, mtime);
+                ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(10519)
```
notroj: You should APLOGNO() for DEBUG-level logging

modules/dav/main/mod_dav.c
```diff
@@ -1207,7 +1272,25 @@ static int dav_method_put(request_rec *r)
...
+                      "Setting mtime for file.");
+            } else {
+                ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(10520)
+                      "Unable to set mtime: provider does not support \"set_mtime\".");
```
notroj: "does not support setting modification times" is sufficient

modules/dav/main/mod_dav.c
```diff
@@ -2839,6 +2938,27 @@ static int dav_method_mkcol(request_rec *r)
         }
     }
 
+    mtime_aware = resource->hooks->set_mtime != NULL;
+    if (conf->honor_mtime_header == DAV_ENABLED_ON && mtime_ret == 1) {
+        if (mtime_aware) {
+            err = (resource->hooks->set_mtime)(resource, mtime);
+            ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO()
+                          "Setting mtime for file.");
+        } else {
```
notroj: as above don't use APLOGNO() for DEBUG-level logging

modules/dav/main/mod_dav.c
```diff
@@ -947,6 +964,37 @@ static int dav_parse_range(request_rec *r,
     return 1;
 }
 
+/**
+ * @return  1 if valid x-oc-mtime,
```
notroj, line 968: I would prefer this with an `apr_status_t` return value, though it's a minor question of style

modules/dav/main/mod_dav.c
```diff
@@ -947,6 +964,37 @@ static int dav_parse_range(request_rec *r,
...
+    }
+
+    for (apr_size_t i = 0; i < strlen(hdr); i++) {
+        if (!apr_isdigit(hdr[i])) {
+            return -1;
+        }
+    }
+
+    errno = 0;
+    n = apr_strtoi64(hdr, &endp, 10);
```
notroj, line 989: Manually setting errno to 0 is not necessary with `apr_strtoi64`

notroj (changes requested):
Thanks for the update, looking good. One minor style nit:
```
} else {
```

should be written as

```
}
else {
```

modules/dav/main/mod_dav.c
leo9800: i am not sure if `apr_status_t` is proper for this case personally.

since `0` is defined as `APR_SUCCESS`, while for the returning value of this function both `1` and `0` are considered success, and `-1` stands for a failure.

actually i was imitating `static int dav_parse_range(request_rec *r, apr_off_t *range_start, apr_off_t *range_end)` in `mod_dav.c`

leo9800:
@notroj 

i am still a bit confused on how should `APLOGNO()` being handled with `APLOG_DEBUG` leveled logs.

there are 3 possible handlings:

1. `ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO() "PLACEHOLDER MSG");`: do not assign tags, let apache maintainers to invoke `docs/log-message-tags/update-log-msg-tags`
2. `ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(01234) "PLACEHOLDER MSG");`: assign tags, by invoking `docs/log-message-tags/update-log-msg-tags`
3. `ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, "PLACEHOLDER MSG");`: do not include `APLOGNO()` at all

the comment 

> as above don't use APLOGNO() for DEBUG-level logging

suggests approach 3, but by invoking `git grep -F APLOG_DEBUG` on the repo huge amount of modules included `APLOGNO()`, which suggests approach 1 or 2.

```
modules/aaa/mod_auth_digest.c:        ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, NULL, APLOGNO(01757)
modules/aaa/mod_auth_digest.c:    ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, cmd->server, APLOGNO(01763)
modules/aaa/mod_auth_digest.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(01764)
modules/aaa/mod_auth_digest.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(01765)
modules/aaa/mod_auth_digest.c:    ap_log_error(APLOG_MARK, APLOG_DEBUG, 0, s, APLOGNO(01768)
modules/aaa/mod_auth_form.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(02982)
modules/aaa/mod_auth_form.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(02983)
modules/aaa/mod_auth_form.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(01812)
modules/aaa/mod_authn_socache.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, rv, r, APLOGNO(01679)
modules/aaa/mod_authn_socache.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(01681)
modules/aaa/mod_authn_socache.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(01684)
modules/aaa/mod_authn_socache.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(01685)
modules/aaa/mod_authn_socache.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(01687)
modules/aaa/mod_authn_socache.c:        ap_log_rerror(APLOG_MARK, APLOG_DEBUG, 0, r, APLOGNO(01688)
# truncated git grep output ...
```

notroj:
(3). For DEBUG-level or TRACE-level , omit the `APLOGNO()` entirely. This was the consensus on the dev@ list, possibly isn't documented anywhere.

leo9800:
@notroj gotcha

leo9800:
should i also rebase this patch to current `trunk`?

i noticed a merge conflict related to log-number, (`APLOGNO`) probably caused by another patch submitted later than this introducing some new log messages.

notroj:
Don't worry about APLOGNO() conflicts, leave that to a committer to fix up. I did a CI run - one issue here: https://github.com/notroj/httpd/actions/runs/18156145629/job/51676367710
move the `apr_size_t` to the top of the function.

leo9800:
@notroj gotcha

seems we need to ensure conformity with C89/C90

notroj:
Some more failures now https://github.com/notroj/httpd/actions/runs/18197214275/job/51806547259#logs

and probably best to rebase onto the trunk again because the trunk tests are failing without 60ad836ff81839e22e508530522f0d4dbbe802cf

notroj:
https://github.com/notroj/httpd/actions/runs/18216313264

:tada: I'll try to get it merged this week, thanks for your persistence!

leo9800:
@notroj thanks for reviewing, cheers🎉

also, kindly asking, could you please review another patch initiated by me at #558? (probably try with CI builds, etc)

notroj:
Sorry took me a bit longer as usual. @leo9800 if you want a full name in CHANGES for this please suggest one :wink: -otherwise your credit will be as seen here:  https://github.com/apache/httpd/commit/2d92bae1277bb3fd84237c7465b1d1cb83346834#diff-732cc45bdcc31912401891f55bf0fb993b939d00bc92ad62f953f6c633c2fffaR1

leo9800:
@notroj gotcha

i would prefer *Leo LI* as my fullname if it is possible to change it anyway 😉

PR #560, opened 2025-10-18
mod_ssl: Keep existing flags when calling SSL_set_shutdown()

mkauf:
Preserve existing flags (SSL_RECEIVED_SHUTDOWN or SSL_SENT_SHUTDOWN) when calling SSL_set_shutdown().

For abortive or unclean shutdowns, additionally call SSL_set_quiet_shutdown().


notroj:
Thanks @mkauf LGTM, is there a user-visible impact for this?

mkauf:
@notroj I think that there is a user-visible impact if "ssl-accurate-shutdown" is used, and the client shuts down the connection first by sending a close notify alert. Previously, mod_ssl would wait for a second close notify alert from the client, ignoring that the client has already sent one.

PR #561, opened 2025-10-20
Draft: Add SSLVHostSNIPolicy directive

notroj:
Draft: Add SSLVHostSNIPolicy directive, which sets a global policy for SSL configuration compatibility between VirtualHosts.

Usage: `SSLVHostSNIPolicy <policy>` where `policy` must be one of:
-  `strict`   => fail for any vhost mismatch
-  `authonly` => fail only for client verification/auth differences
-  `secure`   => fail as authonly plus ciphersuite, protocol, keypair differences
-  `insecure` => allow everything

"secure" is the default.

PR: 69743

### DIFF
--- a/modules/ssl/ssl_engine_kernel.c
+++ b/modules/ssl/ssl_engine_kernel.c
@@ -105,112 +105,26 @@ static int fill_reneg_buffer(request_rec *r, SSLDirConfigRec *dc)
 }
 
 #ifdef HAVE_TLSEXT
-static int ap_array_same_str_set(apr_array_header_t *s1, apr_array_header_t *s2)
+/* Check whether a transition from vhost sc1 to sc2 via SNI is
+ * permitted according to the SSLVHostSNIPolicy setting.  Returns 1 if
+ * the policy treats the vhosts as compatible, else 0. */
+static int ssl_check_vhost_sni_policy(SSLSrvConfigRec *sc1,
+                                      SSLSrvConfigRec *sc2)
 {
-    int i;
-    const char *c;
-    
-    if (s1 == s2) {
+    modssl_snivhpolicy_t policy = sc1->mc->snivh_policy;
+
+    /* Policy: insecure => allow everything. */
+    if (policy == MODSSL_SNIVH_INSECURE)
         return 1;
-    }
-    else if (!s1 || !s2 || (s1->nelts != s2->nelts)) {
-        return 0;
-    }
     
-    for (i = 0; i < s1->nelts; i++) {
-        c = APR_ARRAY_IDX(s1, i, const char *);
-        if (!c || !ap_array_str_contains(s2, c)) {
-            return 0;
-        }
-    }
-    return 1;
-}
+    /* Policy: strict => fail for any vhost transition. */
+    if (policy == MODSSL_SNIVH_STRICT && sc1 != sc2)
+        return 0;
 
-static int ssl_pk_server_compatible(modssl_pk_server_t *pks1, 
-                                    modssl_pk_server_t *pks2) 
-{
-    if (!pks1 || !pks2) {
-        return 0;
-    }
-    /* both have the same certificates? */
-    if ((pks1->ca_name_path != pks2->ca_name_path)
-        && (!pks1->ca_name_path || !pks2->ca_name_path 
-            || strcmp(pks1->ca_name_path, pks2->ca_name_path))) {
-        return 0;
-    }
-    if ((pks1->ca_name_file != pks2->ca_name_file)
-        && (!pks1->ca_name_file || !pks2->ca_name_file 
-            || strcmp(pks1->ca_name_file, pks2->ca_name_file))) {
-        return 0;
-    }
-    if (!ap_array_same_str_set(pks1->cert_files, pks2->cert_files)
-        || !ap_array_same_str_set(pks1->key_files, pks2->key_files)) {
-        return 0;
-    }
-    return 1;
-}
+    AP_DEBUG_ASSERT(sc1->sni_policy_hash);
+    AP_DEBUG_ASSERT(sc2->sni_policy_hash);
 
-static int ssl_auth_compatible(modssl_auth_ctx_t *a1, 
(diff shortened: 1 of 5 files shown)


modules/ssl/ssl_engine_kernel.c
```diff
@@ -126,21 +126,63 @@ static int ap_array_same_str_set(apr_array_header_t *s1, apr_array_header_t *s2)
...
+        return 1;
+    if (a1->nelts != a2->nelts)
         return 0;
+
+    for (n = 0; n < a1->nelts; n++) {
+        ssl_ctx_param_t *p = &APR_ARRAY_IDX(a1, n, ssl_ctx_param_t);
+
+        if (!array_contains_ssl_param(a2, p->name, p->value))
+            return 0;
     }
```
rpluem: Can we have duplicate name/value pairs in a1? If yes, a2 can contain additional name/value pairs not in a1.

notroj: I don't think anything prevents duplicates.

Good point. So the function should be run both ways round, that's probably simplest?

```
    if (!array_cmp_ssl_ctx_params(ctx1->ssl_ctx_param, ctx2->ssl_ctx_param))
        || !array_cmp_ssl_ctx_params(ctx2->ssl_ctx_param, ctx1->ssl_ctx_param))

Comment (rpluem) [reply modules/ssl/ssl_engine_kernel.c]: This would work. I guess there are ways to do this in a better asymptotic runtime, but as the number of elements in the arrays are quite low this should not matter at all in practice and the way you propose makes it better to understand. Nevertheless a comment could be useful for others to understand why we do this both ways round :-)

Comment (notroj) [reply modules/ssl/ssl_engine_kernel.c]: Yeah, definitely. The code might even be simpler to just hash the two arrays? This is already two complex functions for an edge case.

Comment (notroj) [reply modules/ssl/ssl_engine_kernel.c]: It was simpler using hashes. :tada:

**Comment by notroj** [comment]
We could actually eliminate all the runtime complexity by computing and storing a hash for each `SSLSrvConfigRec` based on the policy mode in e.g. `ssl_init_CheckServers`, then just compare that at runtime. Am I overthinking?

File: modules/ssl/ssl_engine_kernel.c [hunk 2]
```diff
@@ -126,21 +126,62 @@ static int ap_array_same_str_set(apr_array_header_t *s1, apr_array_header_t *s2)
...
+
+    if (a1 == a2)
+        return 1;
+    if (a1->nelts != a2->nelts)
+        return 0;
+
+    hash_ssl_params(a1, dig1);
+    hash_ssl_params(a2, dig2);
+
+    return strcmp((char *)dig1, (char *)dig2) == 0;
```
rpluem: Are `dig1` and `dig2` null terminated strings that do not contain any nulls inside or do we need to do a `memcmp(dig1, dig2, APR_MD5_DIGESTSIZE)` instead of the `strcmp`?

modules/ssl/ssl_engine_kernel.c
```diff
@@ -126,21 +126,62 @@ static int ap_array_same_str_set(apr_array_header_t *s1, apr_array_header_t *s2)
...
+    apr_md5_ctx_t ctx;
+    int n;
+
+    apr_md5_init(&ctx);
+    for (n = 0; n < a->nelts; n++) {
+        ssl_ctx_param_t *p = &APR_ARRAY_IDX(a, n, ssl_ctx_param_t);
+
+        apr_md5_update(&ctx, p->name, strlen(p->name));
+        apr_md5_update(&ctx, "##", 2);
+        apr_md5_update(&ctx, p->value, strlen(p->value));
```
rpluem: Does a different ordering of the same key value pairs in the array mean a different configuration?
Won't the hashes be different if the arrays are the same but have a different ordering of the key value pairs?
Hence wouldn't we need to sort the arrays first prior to hashing? e.g.

1. Transform the original array into an array that contains `name + '##' + value` as strings
2. Sort the new array
3. Hash the new array as above

rpluem:
> We could actually eliminate all the runtime complexity by computing and storing a hash for each `SSLSrvConfigRec` based on the policy mode in e.g. `ssl_init_CheckServers`, then just compare that at runtime. Am I overthinking?

This sounds like a very good idea, but see my comments on the updated code.

modules/ssl/ssl_engine_kernel.c
notroj: Good point!

notroj:
I'll try rewriting it to use hashes.

For `SSLOpenSSLConfCmd` - I think semantically it's better to treat this as a list of commands, because that's exactly how it's executed. Sorting would assume the interpretation is indifferent to the order, but e.g. `Options` it is simply flipping bits in the bitmask for each `SSL_CONF_cmd` call so the order is significant. 

```
  SSLOpenSSLConfCmd Options -EmptyFragments
  SSLOpenSSLConfCmd Options EmptyFragments
```

would have different semantics to the inverse order.

rpluem:
> I'll try rewriting it to use hashes.
> 
> For `SSLOpenSSLConfCmd` - I think semantically it's better to treat this as a list of commands, because that's exactly how it's executed. Sorting would assume the interpretation is indifferent to the order, but e.g. `Options` it is simply flipping bits in the bitmask for each `SSL_CONF_cmd` call so the order is significant.
> 
> ```
>   SSLOpenSSLConfCmd Options -EmptyFragments
>   SSLOpenSSLConfCmd Options EmptyFragments
> ```
> 
> would have different semantics to the inverse order.

If order matters than all should be good with the current approach. What about the `memcmp` vs `strcmp`?

notroj:
Definitely will switch to using memcmp to compare the digests

notroj:
Switched to converting the digest to ASCII hex and using strcmp because it was easier for debugging for now. Still PoC but it passes my tests and the runtime code is a lot simpler now. :wink:

rpluem:
Looks good

ylavic (approved):
+1, looks good to me too.

PR #563, opened 2025-10-30
feat: add vnd.sqlite3 mime type and extensions

alexei:
See https://www.iana.org/assignments/media-types/media-types.xhtml for mime type.

See https://en.wikipedia.org/wiki/SQLite for extensions.


notroj:
Thanks for the PR

PR #578, opened 2025-11-20
md_crypt/md_ocsp: free BIGNUMs correctly

botovq:
OpenSSL BIGNUMs should be freed with BN_free() since they (usually) have an allocated bn->d array of BN_ULONG, which the call to OPENSSL_free() leaks.


bigio (approved):
@icing this should be committed to mod_md repository as well.

icing:
Thanks!

Merged in apache trunk as r1929883 and back ported to 2.4.x branch for release.
Released in https://github.com/icing/mod_md/releases/tag/v2.6.6

notroj:
Merged in ab9dd8e2cfe7d62efe5ff8925fbef1de756a2fc2

PR #582, opened 2025-12-13
fix file getting unmapped erroneously when server is under load in multi-threaded multi-core configuration

barr-israel:
As I described in my bugzilla report: [link](https://bz.apache.org/bugzilla/show_bug.cgi?id=69901).
the mod_file_cache module is not thread safe and causes files to get unmapped erroneously when the server is under load in a multi-threaded multi-core configuration.

The bug is caused by concurrent access to the apr_mmap_t ring of the file during creation/destruction of the mmap buckets.
This PR fixes this bug by using immortal buckets, which do not require adding a new apr_mmap_t to the ring.  
Since the files are mapped for the entirety of the server runtime, they will always outlive the immortal buckets, which live until the end of the request.  

I have stress tested this version with 0 failed transactions for over 30 minutes(compared to <0.1s before siege bailed out due to too many failures with the original code)

### DIFF
--- a/modules/cache/mod_file_cache.c
+++ b/modules/cache/mod_file_cache.c
@@ -275,9 +275,8 @@ static int mmap_handler(request_rec *r, a_file *file)
     apr_mmap_t *mm;
     apr_bucket_brigade *bb = apr_brigade_create(r->pool, c->bucket_alloc);
 
-    apr_mmap_dup(&mm, file->mm, r->pool);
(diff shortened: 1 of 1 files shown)


notroj:
Thanks for the PR :) How do you want credit in CHANGES, currently it has you as:

> barr.israel <barr.israel campus.technion.ac.il>

barr-israel:
You're welcome. The current credit is fine.

PR #589, opened 2025-12-19
Adds regression test for CGI env var override 

ichristod:
This change adds a regression test to the pyhttpd test suite to ensure that configuration-defined environment variables with invalid names cannot override server-calculated CGI environment variables, as described [here](https://www.cve.org/CVERecord?id=CVE-2025-65082).


notroj (approved):

notroj:
Thanks for the PR!

PR #593, opened 2026-01-07
Added p7e

alexsyvolap:


notroj:
@alexsyvolap Please provide some document which says this is a standard. 

https://www.iana.org/assignments/media-types/application/pkcs7-mime leads me to
https://datatracker.ietf.org/doc/html/rfc8551#section-3.2.1

there is no mention of .p7e.

alexsyvolap:
@notroj hi! I totally agree regarding the RFC 8551 standard (.p7m).

However, we are required to support the official Ukrainian National Public Key Infrastructure (NPKI) standards. The software used by government services (Diia) and banks forces specific file extensions to distinguish between signed and encrypted data.

According to the official IIT User Manual (hosted on the government portal diia.gov.ua), section File Encryption (page 26): "Всі зашифровані файли мають розширення «.p7e»", "All encrypted files have the extension .p7e" (translated from Ukrainian).

Source: [Official IIT Manual (ca.diia.gov.ua)](https://ca.diia.gov.ua/download/docs/EU13OManual_NAIS.pdf)

Since we receive these files directly from these certified systems, we must accept .p7e as valid application/pkcs7-mime content to ensure interoperability

notroj:
Sorry, we have a policy of only sticking to things published by IANA to keep this file manageable. https://httpd.apache.org/docs/current/mod/mod_mime.html#typesconfig

PR #596, opened 2026-01-19
SSL _V_REMAIN variable fix/improvement

notroj:


notroj:
Broken on feature/ech build which is missing openssl/openssl/pull/29337

PR #597, opened 2026-01-20
core: Add millisecond support to ErrorLogFormat time specifiers

uhliarik:
Add support for millisecond-resolution timestamps in ErrorLogFormat.

%{m}t prints the timestamp in millisecond-resolution.

* include/util_time.h(): Define new AP_CTIME_OPTION_MSEC option for printing time in milliseconds format.

* server/util_time.c(ap_recent_ctime_ex): Handle AP_CTIME_OPTION_MSEC to print time in a millisecond format.

* server/log.c(log_ctime): Recognize the m time option in both fast-path and composite %{...}t formats.


notroj:
Nice PR, thank you!

PR #604, opened 2026-02-17
Adds a custom pytest option for archiving logs and conf files

ichristod:
The purpose of this PR is to enrich python test suite with the capability of log archiving.

With the current situation, httpd/test/gen/apache folder keeps logs and configurations of the last pytest execution either it was about a module or single test.
In addition, if you wanted to compare test results between two httpd versions, you would have to manually preserve the contents of the test/gen/apache/, before cleaning, building and installing a new httpd version.

With the --archive/archive-all option you only need to provide the parameter of the /path/to/archive and it handles the creation and the updates of the archiving folder structure.

I think it helps when you want to compare results across httpd versions.

### DIFF
--- a/test/README.pytest
+++ b/test/README.pytest
@@ -35,6 +35,58 @@ also raises the error log level of the tested modules.
 > pytest -vvv -k test_h2_004_01
 run the specific test with mod_http2 at log level TRACE2.
 
+There is an option to archive the results across different
+modules and httpd versions. The archiving will preserve
+error_log and access_log (cumulative for all tests in the module),
+and save a separate config file for each test case.
+> pytest -k test_core --archive=/path/to/archive/
+
+If you don't provide any specific module, pytest will execute all of them
+and the already archived folders and files will be replaced by the new ones.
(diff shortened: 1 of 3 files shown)


ichristod:
@icing I would appreciate a feedback on this one.

icing:
Since `pytest` can be invoked for specific tests and test names are designed for this, I do not see the need to add that into the test env itself. My opinion.

One can as well run `pytest -k test_h2` and archive the `test/gen` directory contents where one likes.

I would therefore probably never use that option. But if people think it's useful for them, I'd not stop them.

ichristod:
I found it useful in my case mainly because if you choose to execute an entire module, `test/gen` keeps the configuration from the last test execution. So if you want to keep the configuration from each test, that isn’t possible. 
A similar issue occurs with logs too, when you execute more than one module you end up with the logs from the last module only.

ichristod:
The --archive options has been added in trunk with https://github.com/apache/httpd/commit/2414dd2d446c6be6e683307d19ec28970378ea66

PR #607, opened 2026-02-26
Prevent Y2038 bug by using SSL_SESSION_get_time_ex

DanielRuf:
The previous function is deprecated, see:
* https://github.com/openssl/openssl/commit/00a6d0743a38e179f5f9b5de4b73be9fcec0bb4c
* https://github.com/openssl/openssl/issues/23648
* https://github.com/openssl/openssl/pull/21206

### DIFF
--- a/modules/ssl/ssl_engine_kernel.c
+++ b/modules/ssl/ssl_engine_kernel.c
@@ -1965,10 +1965,17 @@ int ssl_callback_NewSessionCacheEntry(SSL *ssl, SSL_SESSION *session)
     idlen = session->session_id_length;
 #endif
 
+#if OPENSSL_VERSION_NUMBER >= 0x30300000
     rc = ssl_scache_store(s, id, idlen,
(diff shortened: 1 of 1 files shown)


covener:
Do we need OPENSSL_VERSION_NUMBER-like checks here?

DanielRuf:
Good question, the function was added in OpenSSL 3.3.0 according to https://github.com/openssl/openssl/blob/master/CHANGES.md#changes-between-32-and-330-9-apr-2024

So I guess yes?
What is the recommended style for the version check?

notroj:
I would inline a `#if OPENSSL_VERSION_NUMBER etc` in that file since there is only one place using that function.

DanielRuf:
I have added some OpenSSL version check in the latest commit.

notroj:
Thanks a lot!

DanielRuf:
Also thank you for the very smooth contribution process.

PR #609, opened 2026-03-03
mod_ssl: Update to OpenSSL 4.0

notroj:


markt-asf:
If building against OpenSSL master you'll also need to switch to using `ASN1_STRING_get0_data()` and friends.

PR #611, opened 2026-03-16
fix message length validation in ajp_msg_check_header

rootvector2:
ajp_msg_check_header() validates msglen against msg->max_size,
but ajp_ilink_receive() reads the body starting at buf + AJP_HEADER_LEN.

This means msglen values up to msg->max_size can pass validation while
writing past the end of the allocated buffer.

Restrict msglen to msg->max_size - AJP_HEADER_LEN to ensure the body
fits within msg->buf.


notroj:
Thanks for the PR, this was fixed in 76dac5d445a

PR #612, opened 2026-03-17
set a liberal default handshake timeout

covener:


modules/filters/mod_reqtimeout.c
```diff
@@ -391,9 +390,7 @@ static int reqtimeout_init(conn_rec *c)
         ap_add_input_filter(reqtimeout_filter_name, ccfg, NULL, c);
 
         ccfg->type = "handshake";
-        if (cfg->handshake.timeout > 0) {
-            INIT_STAGE(cfg, ccfg, handshake);
-        }
+        INIT_STAGE(cfg, ccfg, handshake);
```
rpluem, line 393: This seems wrong in the case that someone explicitly disabled a handshake timeout via configuration by setting it to 0, but has at least another  timeout set.

rpluem: Scratch this. I don't see now why the if was needed at all.

covener:
Closing, probably better to only update the default configuration and not apply to other protocols.

PR #613, opened 2026-03-19
Replace manual string truncation with apr_pstrndup to fix memory leak

metsw24-max:
Replaced manual malloc and memcpy with apr_pstrndup to safely trim the string.
This removes a memory leak avoids edge cases like length underflow and keeps memory handling consistent with APR pools.
No functional change just safer and cleaner code.


notroj:
I expect the number of people who build httpd for BS2000 is similar to the number of humans who reviewed this submission.

PR #615, opened 2026-03-22
Add integer overflow checks to URL escape allocation functions

kodareef5:
`ap_escape_path_segment`, `ap_os_escape_path`, and `ap_escape_urlencoded` in `server/util.c` allocate output buffers using `3 * strlen(input) + constant` without checking for integer overflow. On platforms where `size_t` is 32-bit, inputs exceeding ~1.33GB cause the multiplication to wrap, resulting in undersized allocation.

The HTML escape function `ap_escape_html2` in the same file already has overflow protection at line 2148:

    if (i + j > APR_SIZE_MAX - 6) {
        abort();
    }

This applies the same pattern to the three URL escape functions for consistency. Each now checks `len > (APR_SIZE_MAX - constant) / 3` before the multiplication, calling `abort()` on overflow.

### DIFF
--- a/server/util.c
+++ b/server/util.c
@@ -1816,8 +1816,10 @@ AP_DECLARE(char *) ap_escape_shell_cmd(apr_pool_t *p, const char *str)
     char *cmd;
     unsigned char *d;
     const unsigned char *s;
+    apr_size_t len = strlen(str);
 
-    cmd = apr_palloc(p, 2 * strlen(str) + 1);        /* Be safe */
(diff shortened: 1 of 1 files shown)


notroj:
IMO using `ap_assert()` is better for these cases, it ends in `abort()` after logging something relevant.

kodareef5:
Good call, updated to use `ap_assert()` -- switched all four checks from `if/abort()` to `ap_assert(len <= ...)`.

PR #616, opened 2026-03-22
Add integer overflow checks to apreq parameter and cookie allocation

kodareef5:
`apreq_param_make()`, `apreq_param_decode()`, `apreq_param_encode()`, and `apreq_cookie_make()` compute allocation sizes as `nlen + vlen + constant` without checking for integer overflow. On platforms where `size_t` is 32-bit, large parameter or cookie values can cause the sum to wrap, resulting in undersized allocation.

This adds overflow checks before each `apr_palloc` call:

- `apreq_param_make` (`server/apreq_param.c:38`): checks `nlen + vlen + 1 + sizeof *param`
- `apreq_param_decode` (`server/apreq_param.c:84`): same pattern
- `apreq_param_encode` (`server/apreq_param.c:144`): checks `nlen + dlen` addition and `3 * sum` multiplication
- `apreq_cookie_make` (`server/apreq_cookie.c:143`): checks `nlen + vlen + 1 + sizeof *c`

On overflow, each function returns `NULL` (or `APR_ENOMEM` for `apreq_param_decode`), consistent with existing allocation failure handling.

### DIFF
--- a/server/apreq_cookie.c
+++ b/server/apreq_cookie.c
@@ -140,6 +140,11 @@ APREQ_DECLARE(apreq_cookie_t *) apreq_cookie_make(apr_pool_t *p,
     apreq_cookie_t *c;
     apreq_value_t *v;
 
+    if (nlen > APR_SIZE_MAX - sizeof *c - 1
+        || nlen + sizeof *c + 1 > APR_SIZE_MAX - vlen) {
+        return NULL;
+    }
+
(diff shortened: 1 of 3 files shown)


notroj:
Thanks for the PR. After the vote at https://lists.apache.org/thread/yknw39f3ds9l4x6rsx0y4do096hklob6 we have removed apreq from trunk and this project will longer maintain that code.

PR #617, opened 2026-03-23
reject invalid NameWidth values to prevent unsafe memcpy size

metsw24-max:
This change fixes a potential memory safety issue in mod_autoindex caused by insufficient validation of the NameWidth directive.

Previously values less than 5 were not properly rejected due to a conditional check that allowed 0 to bypass validation


metsw24-max:
@notroj any updates?

notroj:
@metsw24-max thanks for the PRs!

PR #618, opened 2026-03-24
add length validation for Lock-Token to prevent underflow

metsw24-max:
This change improves the handling of the Lock-Token request header in mod_dav by adding proper length validation before performing string indexing and trimming operations

The existing implementation processes the `Lock-Token` header without verifying that the string length is sufficient before

-Accessing `lock_token_hdr[len - 1]`
-Computing `len - 2` for `apr_pstrndup`

### DIFF
--- a/modules/dav/main/util.c
+++ b/modules/dav/main/util.c
@@ -599,12 +599,8 @@ DAV_DECLARE(dav_error *) dav_parse_locktoken(apr_pool_t *p,
     int has_open;
     int has_close;
 
-    if (output == NULL) {
-        return dav_new_error(p, HTTP_INTERNAL_SERVER_ERROR, 0, 0,
-                             "DAV lock token parser called with NULL output pointer.");
-    }
     *output = NULL;
-
+    
     if (input == NULL) {
         return dav_new_error(p, HTTP_BAD_REQUEST, 0, 0,
                              "No Lock-Token specified in header.");
@@ -617,14 +613,15 @@ DAV_DECLARE(dav_error *) dav_parse_locktoken(apr_pool_t *p,
 
     if (len < 2 || !has_open || !has_close) {
         return dav_new_error(p, HTTP_BAD_REQUEST, 0, 0,
-                             "Malformed Lock-Token header.");
+                             "Malformed Lock-Token");
     }
 
-    token = apr_pstrndup(p, input + 1, len - 2);
+    token = apr_pmemdup(p, input + 1, len - 2);
+    token[len - 2] = '\0';
 
     if (*token == '\0') {
         return dav_new_error(p, HTTP_BAD_REQUEST, 0, 0,
-                             "Malformed Lock-Token header.");
+                             "Malformed Lock-Token");
     }
 
     *output = token;
--- a/modules/dav/main/util.c
+++ b/modules/dav/main/util.c
@@ -590,6 +590,47 @@ DAV_DECLARE(time_t) dav_get_timeout_string(request_rec *r,
     return DAV_TIMEOUT_INFINITE;
 }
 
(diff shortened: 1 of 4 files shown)


metsw24-max:
Ensure safe handling of Lock-Token headers by validating string length
before indexing and trimming operations. Prevents potential out-of-bounds
access and size underflow when processing malformed or empty inputs.

Guard access to `[0]` and `[len - 1]` with `len > 0` checks
Ensure `len >= 2` before trimming angle brackets
Reject malformed inputs ->example: <abc, abc>,etc
Apply consistent validation pattern across mod_dav and ms_wdv

This change improves robustness and establishes a safer parsing pattern
for similar header handling paths.

notroj:
This mod_dav.c change doesn't compile, so it is not obvious that you are "improving robustness" here.

metsw24-max:
@notroj I have fixed the compilation issue and pushed an updated patch

notroj:
Seems like we now have a complex test duplicated across two places. How about making a single new function in dav/main/util.c which safely extracts the Lock-Token header, returns it stripped of <> via a char ** output pointer, and returns dav_error * for the error case

metsw24-max:
@notroj 
Refactored the duplicated Lock-Token parsing logic into a single helper function `dav_parse_locktoken` in util.c.

This centralizes validation, ensures safe length checks, and avoids out-of-bounds access. The function returns a `dav_error *` on failure and provides the stripped token via an output parameter.

Both mod_dav.c and ms_wdv.c have been updated to use this helper removing duplication and aligning behavior across modules.

modules/dav/main/util.c
```diff
@@ -590,6 +590,47 @@ DAV_DECLARE(time_t) dav_get_timeout_string(request_rec *r,
...
+    char *token;
+    apr_size_t len;
+    int has_open;
+    int has_close;
+
+    if (output == NULL) {
+        return dav_new_error(p, HTTP_INTERNAL_SERVER_ERROR, 0, 0,
+                             "DAV lock token parser called with NULL output pointer.");
+    }
+    *output = NULL;
```
notroj, line 602: This is unnecessary, we don't generally do argument validation in httpd

metsw24-max:
I have implemented the helper-based refactor for Lock-Token parsing as discussed

modules/dav/main/util.c
```diff
@@ -590,6 +590,47 @@ DAV_DECLARE(time_t) dav_get_timeout_string(request_rec *r,
...
+
+    has_open = (len > 0 && input[0] == '<');
+    has_close = (len > 0 && input[len - 1] == '>');
+
+    if (len < 2 || !has_open || !has_close) {
+        return dav_new_error(p, HTTP_BAD_REQUEST, 0, 0,
+                             "Malformed Lock-Token header.");
+    }
+
+    token = apr_pstrndup(p, input + 1, len - 2);
```
notroj: Use apr_pstrememdup() here

modules/dav/main/util.c
```diff
@@ -590,6 +590,47 @@ DAV_DECLARE(time_t) dav_get_timeout_string(request_rec *r,
...
+
+    if (*token == '\0') {
+        return dav_new_error(p, HTTP_BAD_REQUEST, 0, 0,
+                             "Malformed Lock-Token header.");
```
notroj: For both the error paths I'd say just use "Malformed Lock-Token" since it's used to parse <If> headers as well.

notroj (changes requested):
Thanks a lot for updating the PR, couple more bits of feedback here.

metsw24-max:
-removed redundant validation
-switched to apr_pmemdup as suggested
-unified error handling to "Malformed Lock-Token"

metsw24-max:
@notroj any updates

PR #619, opened 2026-03-24
Fix unchecked i2d_PrivateKey() return value causing size_t overflow and abort

metsw24-max:
The return value of i2d_PrivateKey() in ssl_asn1_table_set() is not validated before being assigned to an apr_size_t.

i2d_PrivateKey() returns an int and may return <= 0 on failure. This value is directly cast to apr_size_t, which can result in a large unsigned value when negative. This leads to excessive memory allocation attempts via ap_malloc()/ap_realloc(), causing the server to abort.

This patch introduces proper validation of the return value before casting, preventing invalid allocations.

Additionally, the caller in ssl_engine_pphrase.c is updated to handle NULL return values from ssl_asn1_table_set() to avoid NULL dereference.

Reproduction:
- Force i2d_PrivateKey() to return -1
- Apache attempts to allocate a huge buffer and aborts

After fix:
- Failure is handled gracefully
- No crash occurs

### DIFF
--- a/modules/ssl/ssl_engine_pphrase.c
+++ b/modules/ssl/ssl_engine_pphrase.c
@@ -179,13 +179,13 @@ apr_status_t ssl_load_encrypted_pkey(server_rec *s, apr_pool_t *p, int idx,
      * are used to give a better idea as to what failed.
      */
     if (pkey_mtime) {
-        ssl_asn1_t *asn1 = ssl_asn1_table_get(mc->retained->privkeys, key_id);
-        if (asn1 && (asn1->source_mtime == pkey_mtime)) {
-            ap_log_error(APLOG_MARK, APLOG_INFO, 0, s, APLOGNO(02575)
-                         "Reusing existing private key from %s on restart",
-                         ppcb_arg.pkey_file);
-            return APR_SUCCESS;
-        }
+    ssl_asn1_t *asn1 = ssl_asn1_table_get(mc->retained->privkeys, key_id);
+    if (asn1 && (asn1->source_mtime == pkey_mtime)) {
+        ap_log_error(APLOG_MARK, APLOG_INFO, 0, s, APLOGNO(02575)
+                     "Reusing existing private key from %s on restart",
+                     ppcb_arg.pkey_file);
+        return APR_SUCCESS;
+    }
     }
 
     ap_log_error(APLOG_MARK, APLOG_INFO, 0, s, APLOGNO(02576)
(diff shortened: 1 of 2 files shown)


metsw24-max:
@notroj any updates?

notroj:
How is it it possible to trigger the `i2d_PrivateKey` failure case here? I would assume this is some only possible in some pathological case, not a practical thing anybody will really it. So add an ap_assert() on the first invocation at most here.

metsw24-max:
Done
replaced the first-call error handling with `ap_assert(derlen > 0)` and removed the check on the second invocation as suggested.

modules/ssl/ssl_engine_pphrase.c
```diff
@@ -181,9 +181,9 @@ apr_status_t ssl_load_encrypted_pkey(server_rec *s, apr_pool_t *p, int idx,
     if (pkey_mtime) {
         ssl_asn1_t *asn1 = ssl_asn1_table_get(mc->retained->privkeys, key_id);
         if (asn1 && (asn1->source_mtime == pkey_mtime)) {
-            ap_log_error(APLOG_MARK, APLOG_INFO, 0, s, APLOGNO(02575)
-                         "Reusing existing private key from %s on restart",
-                         ppcb_arg.pkey_file);
+            ap_log_error(APLOG_MARK, APLOG_INFO, 0, s,
```
notroj: this appears to be a noop apart from removing the APLOGNO()

modules/ssl/ssl_util.c
```diff
@@ -201,7 +201,12 @@ ssl_asn1_t *ssl_asn1_table_set(apr_hash_t *table, const char *key,
 {
     apr_ssize_t klen = strlen(key);
     ssl_asn1_t *asn1 = apr_hash_get(table, key, klen);
-    apr_size_t length = i2d_PrivateKey(pkey, NULL);
+    int derlen = i2d_PrivateKey(pkey, NULL);
+    /* Encoding the key length can only fail in pathological cases which
+     * cannot occur for a key which has already been loaded and used. */
+    ap_assert(derlen > 0);
+
+    apr_size_t length = (apr_size_t)derlen;
```
notroj: C90 violation. Build with `--enable-maintainer-mode`

modules/ssl/ssl_engine_pphrase.c
metsw24-max: Reverted, that hunk slipped in by accident. Also dropped the leftover NULL check on ssl_asn1_table_set() further down, since with the assert it can't return NULL any more, so ssl_engine_pphrase.c is now unchanged from trunk.

modules/ssl/ssl_util.c
metsw24-max: Fixed, moved the declarations back above the statements. Compiles cleanly with -Wdeclaration-after-statement now.

PR #623, opened 2026-03-31
Disable decompression in mod_mime_magic by default

notroj:


modules/metadata/mod_mime_magic.c
```diff
@@ -484,6 +488,7 @@ static void *merge_magic_server_config(apr_pool_t *p, void *basev, void *addv)
                             apr_palloc(p, sizeof(magic_server_config_rec));
 
     new->magicfile = add->magicfile ? add->magicfile : base->magicfile;
+    new->decompression_enabled = add->decompression_enabled;
```
notroj, line 491: This means the setting doesn't inherit properly across vhosts. The simplest way to do this properly is to have ->decomp_enabled default to an "UNSET" macro defined as -1, and then use a ternary operator to inherit in the merge ` = (add->decompress_enabled  != UNSET) ? add->decompress_enabled : base->decompress_enabled`. And also adjust the code that uses the field to special-case the `UNSET` value

modules/metadata/mod_mime_magic.c
```diff
@@ -502,6 +507,19 @@ static const char *set_magicfile(cmd_parms *cmd, void *dummy, const char *arg)
     return NULL;
 }
 
+static const char *set_decompression(cmd_parms *cmd, void *dummy, int arg)
+{
+    magic_server_config_rec *conf = (magic_server_config_rec *)
+    ap_get_module_config(cmd->server->module_config,
+                      &mime_magic_module);
+
+    if (!conf) {
```
notroj, line 516: This is unnecessary, that can never return NULL (it's an array lookup).

notroj (commented):
Thanks for filing the PR and especially for the docs update.

notroj:
Now tracked at #625

PR #624, opened 2026-04-02
improve bounds checking for line length validation

metsw24-max:
Improves bounds checking logic in `mod_substitute` within Apache HTTP Server

Replaces combined arithmetic validation with step-wise boundary checks

Ensures `vb.strlen`, `len`, and `replen` remain within `max_line_length` limits

Updates conditions to avoid reliance on aggregated expressions for validation

Applies consistent bounds checking pattern across related code paths

Adjusts checks involving `space_left` to use safe comparisons

Refines validation of `regm[0].rm_so` against remaining buffer space

Prevents operations from exceeding configured buffer limits


metsw24-max:
@notroj any updates?

PR #626, opened 2026-04-03
harden size calculations, bounds checks, and SHM layout validation

metsw24-max:
adds overflow-safe helpers for size arithmetic and applies them to all allocation and layout calculations in both slotmem backends

rejects creation when `item_size * item_num` or derived sizes overflow preventing under-allocation that could lead to out-of-bounds access during slot indexing

replaces direct size expressions with validated `basesize`, `inuse_size`, and `alloc_size` to ensure all memory allocations are correctly bounded

adds checks in `slotmem_get` and `slotmem_put` to ensure `dest_len` and `src_len` do not exceed the fixed slot size, preventing out-of-bounds reads and writes via memcpy

reorders validation so that slot index bounds are checked before deriving pointers like `inuse + id`, avoiding invalid pointer computation

adds guards against overflow in pointer arithmetic `size * id` before computing slot addresses

validates shared memory layout during `slotmem_attach` by recomputing expected size from descriptor fields and ensuring it does not exceed the actual segment size

prevents use of corrupted or inconsistent shared memory segments that could otherwise lead to invalid memory access

### DIFF
--- a/modules/slotmem/mod_slotmem_plain.c
+++ b/modules/slotmem/mod_slotmem_plain.c
@@ -38,6 +38,26 @@ struct ap_slotmem_instance_t {
 static struct ap_slotmem_instance_t *globallistmem = NULL;
 static apr_pool_t *gpool = NULL;
 
+static int slotmem_size_mul(apr_size_t a, apr_size_t b, apr_size_t *res)
+{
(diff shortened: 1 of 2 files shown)


metsw24-max:
@notroj look into this

notroj:
Thanks for the PR

PR #627, opened 2026-04-05
harden bounds checks and layout size validation in providers

jmestwa-coder:
### slotmem: consolidate layout size checks and fix remaining unchecked paths

Redone against trunk now that r1936268 (#626) covers the create/attach/get/put checks. Remaining changes:

- Replace the slotmem_size_mul/slotmem_size_add call chains with a single slotmem_layout_sizes() helper so create/attach/store share one overflow-checked layout computation.
- store_slotmem() still computed the persist size unchecked; use the helper and drop the persist file on failure.
- get/put: reject NULL buffers with nonzero length and skip the memcpy for zero length.
- slotmem_release() in the shm provider read inuse[id] in the log message before the id bounds check; only read it when id is valid.

### DIFF
--- a/modules/slotmem/mod_slotmem_plain.c
+++ b/modules/slotmem/mod_slotmem_plain.c
@@ -38,24 +38,31 @@ struct ap_slotmem_instance_t {
 static struct ap_slotmem_instance_t *globallistmem = NULL;
 static apr_pool_t *gpool = NULL;
 
-static int slotmem_size_mul(apr_size_t a, apr_size_t b, apr_size_t *res)
+static apr_status_t slotmem_layout_sizes(apr_size_t item_size,
+                                         unsigned int item_num,
+                                         apr_size_t *basesize,
+                                         apr_size_t *totalsize)
 {
-    if (a != 0 && b > ((apr_size_t)-1) / a) {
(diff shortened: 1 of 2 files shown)


jmestwa-coder:
any update?

notroj:
This was pretty similar to #626 but I didn't read it first, sorry. Maybe I slightly prefer the code style in this one, would you be able to redo against trunk?

jmestwa-coder:
Done, redone against trunk. Since r1936268 already covers most of the original patch, this is now just the delta:

- consolidated the size_mul/size_add call chains into a single slotmem_layout_sizes() helper shared by create/attach/store
- store_slotmem() was still computing the persist size unchecked; it now uses the helper and drops the persist file on failure
- get/put reject NULL buffers with nonzero length and skip the memcpy when the length is zero
- the release log message read inuse[id] before the id bounds check in the shm provider; it now only reads it when id is valid

Both files build warning-free with clang, and I checked the helper against the old formulas: identical sizes for all valid inputs, EINVAL for the shapes that used to wrap.

notroj:
Just getting back to some PRs. Thanks @jmestwa-coder for redoing it.

PR #628, opened 2026-04-06
input validation and removal of unsafe numeric parsing

metsw24-max:
This change hardens balancer-manager parameter handling by replacing permissive numeric parsing `atoi atof` with strict range-checked parsing
It ensures malformed or partially valid inputs are rejected instead of being silently coerced into valid values.

### DIFF
--- a/modules/proxy/mod_proxy_balancer.c
+++ b/modules/proxy/mod_proxy_balancer.c
@@ -26,6 +26,9 @@
 #include "apr_escape.h"
 #include "mod_watchdog.h"
 
+#include <errno.h>
+#include <stdlib.h>
+
 static const char *balancer_mutex_type = "proxy-balancer-shm";
 ap_slotmem_provider_t *storage = NULL;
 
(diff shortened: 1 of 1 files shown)


metsw24-max:
@notroj look into this

notroj:
I'll be brutally honest here @metsw24-max - the effect of these comments putting pressure on me to merge your patches is that I will trust you less, and take more time to review your work, not to merge them faster. You're a new contributor here, and we definitely appreciate your PRs. We're under high workload at the moment, give us time.

metsw24-max:
apologies for that i didn’t mean to put any pressure on the review process. I understand the workload and appreciate you taking the time to review. I will wait for your feedback.

modules/proxy/mod_proxy_balancer.c
```diff
@@ -1139,29 +1188,50 @@ static int balancer_process_balancer_worker(request_rec *r, proxy_server_conf *c
          * on that # character, since the character == the flag
          */
         if ((val = apr_table_get(params, "w_status_I"))) {
-            ap_proxy_set_wstatus(PROXY_WORKER_IGNORE_ERRORS_FLAG, atoi(val), wsel);
+            int ival;
```
notroj: coding style nit, declaring `int ival` in every single branch here is clearly making the code worse / harder to read, so move it up a context and just declare it once.

modules/proxy/mod_proxy_balancer.c
```diff
@@ -26,6 +26,9 @@
 #include "apr_escape.h"
 #include "mod_watchdog.h"
 
+#include <errno.h>
+#include <stdlib.h>
```
notroj: `<errno.h>` should be inherited via APR headers already, are these really needed?

notroj (changes requested):
Minor style nits otherwise LGTM

modules/proxy/mod_proxy_balancer.c
metsw24-max: Done, moved it up to the enclosing block so it's declared once, and dropped the per-branch declarations.

modules/proxy/mod_proxy_balancer.c
metsw24-max: You're right, both come in via the APR headers already. Removed them and the module still builds cleanly with -Wall.

PR #629, opened 2026-04-06
harden mod_lbmethod_heartbeat parsing and overflow handling

metsw24-max:
This hardens heartbeat-based load balancing input handling and arithmetic in `mod_lbmethod_heartbeat.c`

Replaces permissive numeric parsing in heartbeat data handling with strict bounded parsing

Added strict integer parser using `apr_strtoi64`

Added safe overflow guards in load-slot accumulation and pick-range computation by saturating at `APR_UINT32_MAX`

### DIFF
--- a/modules/proxy/balancers/mod_lbmethod_heartbeat.c
+++ b/modules/proxy/balancers/mod_lbmethod_heartbeat.c
@@ -22,6 +22,9 @@
 #include "ap_slotmem.h"
 #include "heartbeat.h"
 
+#include <errno.h>
+#include <limits.h>
+
 #ifndef LBM_HEARTBEAT_MAX_LASTSEEN
 /* If we haven't seen a heartbeat in the last N seconds, don't count this IP
  * as allive.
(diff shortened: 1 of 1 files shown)


modules/proxy/balancers/mod_lbmethod_heartbeat.c
```diff
@@ -22,6 +22,9 @@
 #include "ap_slotmem.h"
 #include "heartbeat.h"
 
+#include <errno.h>
+#include <limits.h>
+
```
notroj: These should not be included directly (and should not be needed), they are included by <apr.h> already.

notroj (commented):
One nit otherwise LGTM, thanks for the PR

modules/proxy/balancers/mod_lbmethod_heartbeat.c
metsw24-max: Good catch, dropped both. apr.h already pulls in limits.h and errno via apr_errno.h, so they were redundant.

metsw24-max:
gentle ping

PR #630, opened 2026-04-07
Fix incorrect pool reference in h2_request_rcreate

joesuf4:
Two serious h2 bugs:

First one involves the way h2 defers r->pool cleanups until the connection closes.  This is simply insane, with a one-line repair to tie r->pool to the rcreate pool instead of the long-lived stream pool. This is trivially DOSable by someone with a clue.

Second one avoids a segfault during certain kinds of subrequest lookup processing.


joesuf4:
Not a bug

PR #631, opened 2026-04-08
New tests

kataokaa:


notroj:
This is really impressive and I appreciate the submission... I think it's too complex to merge as-is.
I am not sure exactly what the Docker containers are doing in here.  Is the `infrastructure` directory missing? I can't see where that's coming from.

PR #635, opened 2026-04-16
mpm/event: fix 'occured' -> 'occurred' in event.c comment

SAY-5:
Comment in `server/mpm/event/event.c` line 919 reads `unrecoverable error occured`. Fixed to `occurred`. Comment-only change.


notroj:
Thanks

PR #640, opened 2026-04-28
CMake: update minimum version required

jmckenna:
- to avoid the following fatal error with CMake 4.3.2:
```
CMake Error at CMakeLists.txt:20 (CMAKE_MINIMUM_REQUIRED):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
  ```


notroj:
Thanks for the PR!

notroj:
Not sure if you're using this to build on Linux or Windows but still seeing build failures in CI on Windows after this https://github.com/apache/httpd/actions/runs/25502886791/job/74840430128#step:5:1097

jmckenna:
@notroj I build on Windows with Visual Studio - odd, I'll have a look (I compile in CMake with `CMAKE_BUILD_TYPE=RelWithDebInfo`, but I see that your action uses `CMAKE_BUILD_TYPE=Debug`, I'll try a build with that locally too).

PR #644, opened 2026-05-07
mod_ssl: Expose serialNumber (2.5.4.5) as standard env var from certi…

michael-o:
…ficate DN

serialNumber (2.5.4.5) is a standard LDAP attribute embedded in the subject's and/or issuer's DN, extract it by standard means from the DN and expose via StdEnvVars.

This fixes BZ #35154.

### DIFF
--- a/modules/ssl/ssl_engine_vars.c
+++ b/modules/ssl/ssl_engine_vars.c
@@ -720,6 +720,7 @@ static const struct {
     { "G",     NID_givenName,              1 },
     { "S",     NID_surname,                1 },
     { "D",     NID_description,            1 },
+    { "SerialNumber", NID_serialNumber,    1 },
 #ifdef NID_userId
     { "UID",   NID_userId,                 1 },
 #endif


michael-o:
Verified and works perfectly for me via PHP and `SSLUserName SSL_CLIENT_S_DN_SerialNumber`.

notroj:
Thanks @michael-o any chance of a test case update too? https://github.com/apache/httpd-tests/blob/trunk/t/ssl/varlookup.t

michael-o:
> Thanks @michael-o any chance of a test case update too? https://github.com/apache/httpd-tests/blob/trunk/t/ssl/varlookup.t

Thanks! Checking...

Any chance to nominate for 2.4.x? The change is straight forward.

michael-o:
@notroj Trying to wrap around my head how the stuff works. As far as I understand I first need to modify https://metacpan.org/release/SHAY/Apache-Test-1.43/source/lib/Apache/TestSSLCA.pm#PApache::TestSSLCA and add some "serialNumber" before I can even modify the test. If so, who will publish a new version to CPAN?
The canonical repo is obviously https://svn.apache.org/repos/asf/perl/Apache-Test/trunk/

notroj:
@michael-o oohh.. I was thinking this is just the serial number for the cert which is already there... but it's in the DN. Don't worry about it, too painful to adjust the TestSSLCA.pm for it I expect.

michael-o:
> @michael-o oohh.. I was thinking this is just the serial number for the cert which is already there... but it's in the DN. Don't worry about it, too painful to adjust the TestSSLCA.pm for it I expect.

Alright! Can you backport this to 2.4.x? The change is trivial, but very helpful.

PR #647, opened 2026-05-15
Stab in the dark attempt at fixing the Windows CI

notroj:


notroj:
Made it worse, I give up.

PR #648, opened 2026-05-16
Use macros for status and status line instead of literals

michael-o:
No functional change, just catch up.


notroj:
LGTM

michael-o:
Merged manually.

PR #649, opened 2026-05-18
mod_ssl: fix leak of X509 reference in ssl_hook_Access_classic quick renegotiation path

rootvector2:
While reading `ssl_hook_Access_classic()`, I noticed that the `renegotiate_quick` branch obtains a peer certificate with `SSL_get_peer_certificate(ssl)` and never releases it on the common path.

## The leak

```c
cert_stack = (STACK_OF(X509) *)SSL_get_peer_cert_chain(ssl);
cert = SSL_get_peer_certificate(ssl);   /* refcount +1, owned by caller */

if (!cert_stack || (sk_X509_num(cert_stack) == 0)) {
    ...
    cert_stack = sk_X509_new_null();
    sk_X509_push(cert_stack, cert);     /* (A) ownership now in stack */
}
...
if (cert_stack != SSL_get_peer_cert_chain(ssl)) {
    sk_X509_pop_free(cert_stack, X509_free);  /* frees (A) */
}
```

`SSL_get_peer_certificate()` increments the X509 refcount and returns a reference the caller owns. Every other call site in this file releases it with `X509_free()` (lines 319, 615, 1531).

The trailing `sk_X509_pop_free()` only runs when we built the stack ourselves — i.e., only when the original `SSL_get_peer_cert_chain()` was empty/NULL. In the common case where the peer chain already contains certificates, the reference is never freed, leaking one X509 refcount per quick renegotiation. The two early-return paths ("Cannot find certificate storage" and the `X509_STORE_CTX_init` failure return added in bz 65902) leak unconditionally.

This matches the documented OpenSSL contract for `SSL_get_peer_certificate(3)`:

> The reference count of the X509 object [...] is incremented by one [...]. The X509 object must be explicitly freed using X509_free().

## The fix

Track the owned reference as `peer_cert`. When ownership is transferred into the freshly-created stack via `sk_X509_push()`, clear `peer_cert` so `sk_X509_pop_free()` remains the sole owner and there is no double free. Release `peer_cert` on every exit path.

Only affects TLS ≤ 1.2 renegotiation (`ssl_hook_Access_classic`); TLS 1.3 has no renegotiation and uses `ssl_hook_Access_modern`.

## Notes for the reviewer

- Found by static reading. I have not run a leak-sanitizer reproducer; the analysis follows the OpenSSL ownership contract and the pattern used by the other `SSL_get_peer_certificate()` sites in this file.
- No `CHANGES` entry — happy to add one in the format the committer prefers.

### DIFF
--- a/modules/ssl/ssl_engine_kernel.c
+++ b/modules/ssl/ssl_engine_kernel.c
@@ -352,7 +352,101 @@ static int ssl_check_post_client_verify(request_rec *r, SSLSrvConfigRec *sc,
 }
 
 /*
- *  Access Handler, classic flavour, for SSL/TLS up to v1.2 
+ * Manually re-verify the peer certificate for the "quick" renegotiation
+ * path.  Split out of ssl_hook_Access_classic() so the X509 reference from
+ * SSL_get_peer_certificate() and the stack we may build are released on a
+ * single cleanup path instead of being duplicated across each error return.
+ * Returns OK on success, or an HTTP_* error.
+ */
+static int ssl_verify_peer_quick(request_rec *r, SSL *ssl, SSL_CTX *ctx)
+{
+    STACK_OF(X509) *cert_stack;
+    X509 *cert;
+    X509 *peer_cert;
+    X509_STORE *cert_store = NULL;
+    X509_STORE_CTX *cert_store_ctx;
+    int depth;
+    int rv = OK;
+
+    cert_stack = (STACK_OF(X509) *)SSL_get_peer_cert_chain(ssl);
+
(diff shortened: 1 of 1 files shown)


notroj:
Thanks for the PR, this looks correct to me. It does make an already-long function even longer, I wonder if we could split out the guts of that code path into a separate function and avoid duplicating the free()s on the error paths?

rootvector2:
good point. pulled the whole quick-reneg path out into `ssl_verify_peer_quick()` with a single `cleanup:` label, so the stack and `peer_cert` free()s live in one place and the error returns just `goto cleanup`. the caller is now a few lines. behavior is unchanged.

rootvector2:
any update?

PR #650, opened 2026-05-18
Mod random

PilouGit:
An Apache module that generates cryptographically secure random base64-encoded strings and injects them as environment variables for use by applications and other modules.


dune73:
Thank you.

notroj:
If it depends on OpenSSL and says it uses a CSPRNG I think it would be better to use `RAND_bytes` directly, which is actually a CSPRNG. No opinion really on adding it to httpd, you will have more flexibility maintaining this as a third-party module.

PilouGit:
I will have more flexibility maintaining this as a third-party module, that’s true. But my goal is to share this code with as many people as possible (nobody goes to my repo). In an industrial context, no one installs a third-party module if it’s not part of the httpd trunk.

PR #652, opened 2026-05-22
OpenSSL 4 fixes in md module

pierre-labastie:
This is needed in addition to the ssl module


notroj:
Thanks for the PR, I synced the change from https://github.com/icing/mod_md/commit/9fd90380db0722f0ccc14ff1c4da0ca21d4da75c
in d534e42b07969755b27eff2892ea56cfd909f063 which is our current/preferred way to keep this module in sync

PR #657, opened 2026-06-01
CMake: backport minimum version required change, and apply workaround for error Unexpected enablement value "i"

jmckenna:
into branch `2.4.x`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -606,6 +606,9 @@ FOREACH (modinfo ${MODULE_LIST})
 
 ENDFOREACH()
 
+#workaround for error: Unexpected enablement value "i" for mod_authnz_ldap
+UNSET(i)
+
 SET(install_targets)
 SET(install_bin_pdb)
 SET(install_modules) # special handling vs. other installed targets
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,9 +15,11 @@
 #
 # Read README.cmake before using this.
 
-PROJECT(HTTPD C)
+# CMAKE_MINIMUM_REQUIRED should be the first directive in the file:
+# https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html
+CMAKE_MINIMUM_REQUIRED(VERSION 3.16...3.29)
 
-CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
+PROJECT(HTTPD C)
 
 INCLUDE(CheckSymbolExists)
 INCLUDE(CheckCSourceCompiles)


jmckenna:
@notroj this should solve your branch 2.4.x issues.

jmckenna:
Can someone click "approve and run" for the workflow here?

jajik:
Note that this is not needed for the Windows workflow. The issue was the new VS (and its path), not the CMake.

However, increasing the minimum version is required for a newer CMake (such as the one in Fedora 44).

jmckenna:
> Note that this is not needed for the Windows workflow. The issue was the new VS (and its path), not the CMake.
> 
> However, increasing the minimum version is required for a newer CMake (such as the one in Fedora 44).

Incorrect, I compile locally using MSVC 2022 and both changes are needed to compile successfully (locally, yes I do this compiling locally) with branch `2.4.x`.

jmckenna:
> Note that this is not needed for the Windows workflow. The issue was the new VS (and its path), not the CMake.
> 
> However, increasing the minimum version is required for a newer CMake (such as the one in Fedora 44).

regarding increasing the minimum CMake version, I submitted that to `trunk` a month ago, yes I agree, it is also important (yet not backported to branch `2.4.x`).

For background: all this started when Joe was pointing to a failed branch `2.4.x` run, that was missing the minimum required version in CMakeLists.txt, and had the error: Unexpected enablement value "i" (the unset trick works around that).

jajik:
> > Note that this is not needed for the Windows workflow. The issue was the new VS (and its path), not the CMake.
> > However, increasing the minimum version is required for a newer CMake (such as the one in Fedora 44).
> 
> Incorrect, I compile locally using MSVC 2022 and both changes are needed to compile successfully (locally, yes I do this compiling locally) with branch `2.4.x`.

[This](https://github.com/jajik/httpd/actions/runs/26806644161) is the result of a Windows CI build of 2.4.x with only [one commit](dace8264b02f7e2110c731238fc46adde09cbfd0) on top of it. It throws warnings, but it works.

jmckenna:
@jajik I am not speaking of CI builds, or Docker, etc.  I am trying to tell you that to compile locally with MSVC 2022 and branch `2.4.x`, the CMake minimum version change, and the unset(i) are required for a successful compile.

(oddly enough, Joe showed a GitHub run in branch `2.4.x` that showed the same error, so I jumped in to try and provide a fix/workaround)

If any Windows MSVC 2022 users try to compile branch 2.4.x now they will need both of these workarounds.  

I am happy that you showed a MSVC 2026 successful run.  That doesn't help MSVC 2022 users trying to compile branch `2.4.x` (they will need both workarounds that I provided in this pull request).

Anyway, let's close all this and move on with our lives :)

I have closed the ticket.

Sorry for the noise.

jajik:
@jmckenna The run I shared is using VS 2022. Unfortunately, GitHub's continuous container image rollout is not helping here.

I'll try the build locally too, but I need to set up a Windows VM first. :)

PR #659, opened 2026-06-02
Fix Windows builds in GH Actions

jajik:
I noticed the discussion on the dev list about the WIndows CI failures. This was caused by the changes to the default `windows-latest` container. The new one uses VS 2026 with a newer CMake and a different path.

Similar changes were already implemented in [commons-daemon](https://github.com/apache/commons-daemon/commit/ddceb90fdff59144fb11249287aeb365e5fe7c19).

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -20,7 +20,6 @@ on:
 
 jobs:
   build:
-    if: false
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


jajik:
The job seems to work as expected in my fork: https://github.com/jajik/httpd/actions/runs/26805594043 (even with `windows-2025-vs2026` explicitly https://github.com/jajik/httpd/actions/runs/26806441320)

jajik:
Given 767d9550e7c9036fa1604c15f975f7c590d95752, this is no longer wanted/needed.

PR #660, opened 2026-06-02
CI: Update GH Actions before Node.js 20 deprecation + minor tweaks

jajik:
The current CI triggers warnings because in two weeks Node.js 20 actions will stop working. This PR updates all affected actions.

Additionally, it drops `shell: cmd` and uses PowerShell for the Windows workflow since it has already been introduced there (see 767d9550e7c9036fa1604c15f975f7c590d95752).

Finally, it removes the fix for Ubuntu 22.04 since it's no longer used.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -344,9 +344,6 @@ jobs:
         echo "${{ matrix.env }}" >> $GITHUB_ENV
         echo JOBID=`echo "OS=$ImageOS ${{ matrix.notest-cflags }} ${{ matrix.env }} ${{ matrix.config }}" \
            | md5sum - | sed 's/ .*//'` >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


notroj:
@jajik the third bit is conflicting when I try to apply it with patch, can you redo after recent trunk changes?

jajik:
@notroj Should be done now. It seems Ivan is trying to eat my lunch today!

jajik:
Another rebase after trunk changes.

@notroj I pushed a new commit that adds `cmake --version` to the Windows workflow. Given the recent issues, it could make things clearer.

Basically, the old Windows image used CMake 3.31.6, while the new one uses 4.3.2. That means that the 2.4.x branch will eventually need (apart from 767d9550e7c9036fa1604c15f975f7c590d95752) at least 3.5 in `CMAKE_MINIMUM_REQUIRED` in order to work (and this also affects Fedora 44, for example).

PR #665, opened 2026-06-04
Added support for running pyHttpd on Windows

PaulLodge:
This is an update that enables running pyHttpd on windows.

the build-win.bat, first installs all the needed dependencies, this is done using vcpkg and the script used is very similar in approach to the GHA, it then builds, compiles and installs httpd in an arbitrary directory, at the beginning of the install phase the config.ini is created, it's only at this time cmake knows for sure where the install directory is and therefore httpd.  

**_Prerequisites_**:
**curl:** this is needed in some of the tests, currently the curl included in windows isn't the same as it's Linux counterpart and as such needs to be pre-installed (from curl.se). once the install is complete the config_win.ini.in (curl_bin) will need updating with the full path, before running the above mentioned build-win.bat script.

**_Caveat_**:
**h2load:** currently this requires cygwin or similar to be installed, and all tests that run it are skipped (from the command line when pytest is run).

Tested on Linux to ensure that the changes didn't break anything 🤞🏻


jfclere (commented):
Why are you use vcpkg?
For example I have:
C:\cygwin64\home\Administrator\httpd-trunk\jfclere>cmake --version
cmake version 3.22.22022201-MSVC_2

CMake suite maintained and supported by Kitware (kitware.com/cmake).

C:\cygwin64\home\Administrator\httpd-trunk\jfclere>which cmake
/cygdrive/c/Program Files/Microsoft Visual Studio/2022/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/bin/cmake

C:\cygwin64\home\Administrator\httpd-trunk\jfclere>

PaulLodge:
Hi, @jfclere I used vcpkg simply for 2 reasons the first is to keep it consistent with the github actions https://github.com/apache/httpd/blob/trunk/.github/workflows/windows.yml#L66 and secondly these libraries are maintained by Microsoft.

notroj:
I think this is now done?

jfclere:
@notroj we can close it.

PR #666, opened 2026-06-05
http2: Log header limit violations only once

mkauf:
The previous fix in https://github.com/icing/mod_h2/pull/181 does not work anymore.


notroj:
@mkauf all your patches are obviously correct, please accept the standing invitation to commit rights so you can merge yourself.

PR #667, opened 2026-06-08
Configure GitHub workflows to use concurrency cancel-in-progress for

apupier:
pull requests

see recommended best practices at Apache
https://cwiki.apache.org/confluence/pages/viewpage.action?spaceKey=INFRA&title=GitHub+Actions+Recommended+Practices

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -26,6 +26,10 @@ env:
   # This will need updating as the ubuntu-latest image changes:
   PHP_FPM: "/usr/sbin/php-fpm8.3"
 
+concurrency:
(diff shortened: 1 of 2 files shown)


notroj:
Thanks for the PR! Are you able to tell if this "worked" somehow just looking at the runs?

apupier:
> Are you able to tell if this "worked" somehow just looking at the runs?

to check if it works after it is merged, you can create a PR, then force-pushed a chage right-after. if there are PRs checks in doubles still active, visible only in the Actions atb, it didn't work. If there is one active and one cancelled, it worked

apupier:
tested here https://github.com/apache/httpd/pull/668

PR #668, opened 2026-06-08
[DO NOT MERGE] Dummy change to test PR dependency

apupier:


apupier:
see that the previous trigger was cancelled: https://github.com/apache/httpd/actions?query=is%3Acancelled

EDIT: or skipped in case the job was still in queue: https://github.com/apache/httpd/actions?query=is%3Acancelled+is%3Askipped

image

notroj:
Perfect, thank you again!

PR #671, opened 2026-06-14
Hardening integer overflow guards in escape helper functions for atta…

SABITHSAHEB:
# server/util: Harden escape helpers against allocation size overflows

## Summary

This patch adds integer overflow validation to several escape helper functions in `server/util.c` before calculating allocation sizes.

These functions derive buffer sizes from input lengths using expressions such as `3 * len + 1` and `10 * len`. Without validation, extremely large values could cause arithmetic wraparound, resulting in an undersized allocation and potential memory corruption during subsequent writes.

## Changes

* Added overflow checks before allocation size calculations.
* Prevented allocations when the computed size would overflow `size_t`.
* Preserved existing behavior for valid inputs.
* Added regression tests covering boundary and overflow cases.

## Security Impact

This change hardens memory allocation logic against integer-overflow conditions that could otherwise lead to heap buffer overflows when processing unusually large inputs.

## Testing

* Built successfully with GCC and MSVC.
* Added regression tests for safe boundary values and overflow conditions.
* Verified normal behavior remains unchanged for valid inputs.

## Notes

* No API changes.
* No new dependencies.
* Backward compatible.

### DIFF
--- a/modules/dav/main/ms_wdv.c
+++ b/modules/dav/main/ms_wdv.c
@@ -94,7 +94,9 @@ static const char *mswdv_urlencode(request_rec *r, const char *str)
     char *output;
     char *op;
 
-    output = apr_palloc(r->pool, 3 * strlen(str) + 1);
+    apr_size_t slen = strlen(str);
+    ap_assert(slen <= (APR_SIZE_MAX - 1) / 3);
+    output = apr_palloc(r->pool, 3 * slen + 1);
     op = output;
 
     for (ip = str; *ip; ip++) {
--- a/modules/mappers/mod_rewrite.c
+++ b/modules/mappers/mod_rewrite.c
@@ -722,12 +722,17 @@ static char *escape_backref(apr_pool_t *p, const char *path,
                             const char *escapeme, const char *noescapeme,
                             int flags)
 {
-    char *copy = apr_palloc(p, 3 * strlen(path) + 1);
+    apr_size_t plen = strlen(path);
+    char *copy;
     const unsigned char *s = (const unsigned char *)path;
-    unsigned char *d = (unsigned char *)copy;
+    unsigned char *d;
     int noplus = (flags & RULEFLAG_ESCAPENOPLUS) != 0;
     int ctls = (flags & RULEFLAG_ESCAPECTLS) != 0;
     unsigned char c;
+
+    ap_assert(plen <= (APR_SIZE_MAX - 1) / 3);
+    copy = apr_palloc(p, 3 * plen + 1);
+    d = (unsigned char *)copy;
 
     while ((c = *s)) {
         if (((ctls ? !TEST_CHAR(c, T_VCHAR_OBSTEXT) : !escapeme)
--- a/modules/proxy/mod_proxy_ftp.c
+++ b/modules/proxy/mod_proxy_ftp.c
@@ -142,12 +142,15 @@ static const char *ftp_escape_globbingchars(apr_pool_t *p, const char *path, pro
 {
     char *ret;
     char *d;
+    apr_size_t plen;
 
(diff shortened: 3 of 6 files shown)


notroj:
Using `ap_assert` in these cases is correct, don't remove that, if these paths are reachable you are just introducing some new error path which would crash differently anyway. Don't add new such paths anywhere, just use `ap_assert` throughout.

SABITHSAHEB:
Makes sense, reverted to ap_assert across the board: restored the ones I'd removed and used it for the new checks too instead of the error-return paths. I also pulled the unrelated mod_core atoi changes and the cookie reverse-map sizing churn so this stays focused on the escape helpers. Pushed.

SABITHSAHEB:
any update?

notroj:
```
mod_rewrite.c: In function 'escape_backref':
mod_rewrite.c:727:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
  727 |     char *copy = apr_palloc(p, 3 * plen + 1);
      |     ^~~~
```

this doesn't compile in C90 in various places. Build with `./configure --enable-maintainer-mode`

SABITHSAHEB:
Fixed and pushed. The offenders were escape_backref in mod_rewrite.c and ftp_escape_globbingchars in mod_proxy_ftp.c, where the new declarations ended up after statements; moved them up into the declaration blocks. Rebuilt with --enable-maintainer-mode and the touched files compile clean now under -Wdeclaration-after-statement. The only failure left in my tree was the pre-existing SSL_get_handshake_rtt apr_uint64_t vs uint64_t mismatch on macOS with homebrew apr, unrelated to this patch.

notroj:
Squash it please.

SABITHSAHEB:
Done, squashed to a single commit.

SABITHSAHEB:
Squashed to a single commit and rebased onto current trunk, no content changes.

notroj:
Analysis by Claude:

> **The `proxy_util.c` guard cannot fire, and misses the overflow that is actually there.** `ap_proxy_canonenc_ex()` takes `int len`, so in `ap_assert(len <= (APR_SIZE_MAX - 1) / 3)` the left operand is promoted to `apr_size_t` and compared against ~6.1e18, which is larger than any `int` — the assertion is always true. A negative `len` also converts to a large unsigned value and passes. Meanwhile `3 * len + 1` is evaluated in `int` arithmetic, which is undefined for `len > INT_MAX / 3` (~715 million), and that is the case the guard is meant to cover. The check needs to stay in the `int` domain, e.g. `len >= 0 && len <= (INT_MAX - 1) / 3`.
>
> **The `ap_make_full_path()` check can wrap in exactly the way it is guarding against.** In `ap_assert(len1 <= APR_SIZE_MAX - len2 - 3)` both `len1` and `len2` are `apr_size_t`, so when `len2 > APR_SIZE_MAX - 3` the subexpression `APR_SIZE_MAX - len2 - 3` wraps to a large value and the assertion passes. `len2` needs bounding before it is subtracted.
>
> **The remaining guards are not reachable.** On 64-bit, `3 * len + 1` overflows only for a string longer than 6.1e18 bytes. The `ssl_engine_vars.c` cases are bounded far more tightly than that: `ciphers_len`, `extids_len`, `alpn_len` and the rest come from 16-bit TLS extension length fields inside a record capped around 16KB, so `* 2 + 1` cannot exceed a few hundred kilobytes. `escape_backref()`'s path comes from a URI bounded by `LimitRequestLine`, 8190 by default. None are reachable on 32-bit either, where the threshold is still around 1.4GB.
>
> The first two are worth correcting whatever is decided about the rest, since a guard that cannot fire reads as though the case has been handled.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #673, opened 2026-06-22
Fix pre-1.1.0 OpenSSL compat macros in ssl_private.h

crlorentzen:
Fixes pre-1.1.0 compat macros for X509_get0_notBefore/notAfter/serialNumber

The compat macros added in 83d565f for pre-1.1.0 OpenSSL used incorrect names (X509_get0_before/X509_get0_after) which don't match the actual API names used in the code (X509_get0_notBefore/X509_get0_notAfter). This caused link failures when building against OpenSSL 1.0.2.

Also add missing X509_get0_serialNumber compat macro which was used in ssl_engine_log.c but never shimmed for pre-1.1.0.

## Testing

Verified this fixes link failures when building httpd 2.4.68 against OpenSSL 1.0.2 (undefined references to X509_get0_serialNumber, X509_get0_notBefore, X509_get0_notAfter). Build completes successfully with this patch applied.


notroj:
Thanks!

PR #674, opened 2026-06-23
Removed modules/md/md_tailscale.c from CMakeList.txt so mod_md can be built on Windows.

PaulLodge:
The file modules/md/md_tailscale.c is no longer part of modules/md and causes the build to fail on Windows (using cmake), when building with all the prerequisites for mod_md are included.


notroj:
Thanks @PaulLodge

PR #677, opened 2026-07-02
fix off-by-one buffer overflow in proxy_send_dir_filter

arshsmith1:
proxy_send_dir_filter accumulates each FTP listing line into a fixed MAX_STRING_LEN buffer, and an over-long line from the backend fills it to the last byte with its newline truncated, so the ls -s1 fallback branch then strcat()s a newline two bytes past the array. A bare-LF line trips a matching underflow: the trailing-newline strip walks n down to 0 and reads ctx->buffer[n-1] before the buffer. Guard the strip with n > 0 and bound the re-appended newline so both stay inside ctx->buffer.


arshsmith1:
gentle ping

PR #679, opened 2026-07-06
mod_charset_lite: bound ctx->buf writes in finish_partial_char

arshsmith1:
1. set_aside_partial_char stops once a straddling char reaches sizeof(ctx->buf) (FATTEST_CHAR) and flags EES_LIMIT, but finish_partial_char appends to ctx->buf[ctx->saved] with no such limit.
2. if apr_xlate_conv_buffer keeps returning APR_INCOMPLETE while input remains, ctx->saved grows past the 8-byte buffer and the next write lands out of bounds.

Added the sizeof(ctx->buf) bound to the loop so an over-wide char takes the existing EES_LIMIT path, the same way set_aside_partial_char already refuses one.


arshsmith1:
any update?

PR #686, opened 2026-07-26
Fix ACME RewriteRule pattern in .htaccess recipe

deprrous:
The ACME exemption example is chained from the `.htaccess` HTTPS redirect recipe in this file.

In per-directory context, `RewriteRule` matches the path without a leading slash, so `^/\.well-known/acme-challenge/` does not match ordinary ACME challenge requests there. Dropping the leading slash makes the exemption line up with the `.htaccess` context described above it.

### DIFF
--- a/docs/manual/rewrite/remapping.xml
+++ b/docs/manual/rewrite/remapping.xml
@@ -188,7 +188,7 @@ RewriteRule "^(.*)" "https://%{SERVER_NAME}$1" [R=301,L]
 
 <highlight language="config">
 RewriteEngine On
-RewriteRule "^/\.well-known/acme-challenge/" - [L]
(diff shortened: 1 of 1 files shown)


notroj:
@deprrous appreciate all your contributions - since we have to merge these into SVN and these are all really small changes, could you stick them into a smaller number of PRs?  They would be much easier to review/merge in one go.

jfclere:
The rewrite directive in the actual doc is correct.
What you are suggesting won't work in a VirtualHost
I don't see how to get your example in an .htaccess or Directory
The ACME challenge is /.well-known/acme-challenge/token and mixing Directory and rewrite is looking for troubles.

PR #688, opened 2026-07-26
Use ProxySet for standalone worker example

deprrous:
The `ProxyPass "/notused" ...` example defines the worker, but it also maps a live local path.

`ProxySet` already has a documented worker-only form in this file and expresses the intended setup here without creating an extra reverse-proxy mapping.


jfclere:
AH00526: Syntax error on line 571 of /home/jfclere/APACHE/conf/httpd.conf:
ProxySet can not find 'http://backend.example.com' Worker.
httpd: abnormal exit 1
Your suggestion is wrong.

PR #690, opened 2026-07-26
Fix mod_asis handler names in mod_mime docs

deprrous:
The `mod_mime` examples currently use `send-asis` / `asis-handler`, but `mod_asis` registers `send-as-is`.

This updates the example and the later handler reference to use the actual handler name.


notroj:
Thanks for the PR.

PR #691, opened 2026-07-26
Fix mod_rewrite guide seealso link

deprrous:
The `mod_rewrite` module page already links to `../rewrite/` in the summary, but the `See also` entry still points at `rewrite/index.html` from inside `docs/manual/mod/`.

This switches the `See also` link to the same working relative target.


notroj:
Thanks, LGTM.

PR #692, opened 2026-07-26
Remove unsupported AllowCONNECT None syntax

deprrous:
`mod_proxy_connect` parses `AllowCONNECT` arguments as numeric ports or ranges only.

This removes the unsupported `None` alternative from the directive syntax and drops the sentence that recommends it.


notroj:
Looks like that was committed by accident but it's actually a good idea, proposed implementation in #719

PR #696, opened 2026-07-26
Fix CGI Python examples in howto docs

deprrous:
This fixes two small drift issues in the CGI Python examples.

The environment-dump example in both `cgi.xml` and `cgi.xml.fr` lost the indentation on the loop body, so the sample no longer parses as Python. The French page also still used `first.py` in the `chmod` example even though that page's sample script is named `premier.py`.

Checks:
- `xmllint --noout docs/manual/howto/cgi.xml`
- `xmllint --noout docs/manual/howto/cgi.xml.fr`
- `git diff --check`

### DIFF
--- a/docs/manual/howto/cgi.xml
+++ b/docs/manual/howto/cgi.xml
@@ -526,7 +526,7 @@ import os
 
 print("Content-type: text/html\n")
 for key, value in os.environ.items():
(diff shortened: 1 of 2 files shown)


notroj:
Thanks!

PR #715, opened 2026-08-26
mod_mime_magic: reject negative indirect offset in mget

arshsmith1:
mget() only checks the upper bound of the offset it uses to index the sniff buffer, but that offset is a signed long, and for indirect ">(...)" rules it is derived from a value read out of the file being sniffed (offset = p->l + m->in.offset, where p->l comes from a BELONG/LELONG conversion that sign-extends bytes >= 0x80). A crafted file whose field yields a small negative offset slips through the guard, because offset + sizeof(union VALUETYPE) promotes offset to size_t and wraps to a small positive value, so the following memcpy reads up to 64 bytes before buf in magic_process. Reject negative offsets in both bounds checks so the read stays inside the buffer; valid rules resolve to a non-negative in-buffer position and are unaffected.


notroj:
Thanks for the PR.

PR #733, opened 2026-09-02
try to fix failing test_008_restart

jfclere:
Co-Authored-By: Claude Opus 4.6


jfclere:
I am not able to reproduce it.

notroj:
I think it should be fixed with f723596 but we can try again if not!

PR #737, opened 2026-09-04
Fix flaky test_008_restart

notroj:
see also #733

### DIFF
--- a/test/pyhttpd/env.py
+++ b/test/pyhttpd/env.py
@@ -920,15 +920,31 @@ class HttpdTestEnv:
             log.warning("port still in use after stop")
         return 0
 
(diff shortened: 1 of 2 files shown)


notroj:
## Cause - from Claude

`HttpdTestEnv.apache_hard_restart()` ran `httpd -k restart`, which only
sends SIGHUP to the parent and returns, and then `is_live()`, which is
satisfied by the first answered request. The parent restarts
asynchronously (event.c `event_run`: reap the old children with SIGTERM,
`startup_children`, then log AH00489 "resuming normal operations"), so the
answer `is_live()` accepts can come from an old child which is about to be
killed. The test's own request then lands on an old child, which dies
mid-request: empty reply.

PR #749, opened 2026-09-09
Avoid UB trimming empty response header values

mcc0nnell:
Several response/header parsing paths trim trailing whitespace by first forming a pointer to `strlen(value) - 1`. For an empty string, that constructs a pointer before the start of the object before the loop condition can reject it.

This change adds `ap_cstr_stripws(char *)` in `server/util.c` and declares it in `include/httpd.h`. The helper uses an end-exclusive pointer, so empty and all-whitespace strings are handled without underflow.

The duplicated trailing-whitespace logic is replaced in:

- `mod_proxy_http`
- `mod_proxy_uwsgi`
- `mod_proxy_hcheck`
- `server/util_script.c`
- `mod_cern_meta`
- `cfg_trim_line()`
- `byterange_filter`
- `mod_negotiation`

Coverage includes unit cases for empty strings, all-whitespace strings, ordinary values, trailing whitespace, and preservation of leading whitespace. The proxy regression tests also retain backend responses containing a valid empty `X-Empty:` header for both HTTP proxy and uWSGI paths.

Validation on the current branch in a clean Ubuntu 24.04 container:

- full httpd build succeeds with the affected modules enabled
- `httpdunit` passes 60/60, including all seven `ap_cstr_stripws()` cases
- `test_03_response.py` and `test_05_uwsgi.py` pass 6/6, including the HTTP proxy and uWSGI empty-header regressions
- `git diff --check`: clean
- modified Python regression files compile cleanly with `py_compile`

I have not demonstrated a crash, out-of-bounds read/write, information disclosure, or other exploit primitive; this remains a narrow UB hardening change.


notroj:
Nice find. Could you add an `ap_cstr_stripws` to `server/util.c` (or maybe a better name comes to mind) - given the prevalence of the error?

mcc0nnell:
Thanks — added `ap_cstr_stripws()` in `server/util.c`, declared it in `httpd.h`, and switched the affected callers over to it. I also added focused unit coverage for empty/all-whitespace/trailing-whitespace cases and converted three additional instances of the same underflow-prone trimming idiom. The existing proxy regressions remain in place.

mcc0nnell:
Fresh validation on the current single-commit head is clean: full Ubuntu 24.04 build succeeds, `httpdunit` passes 60/60 (including all 7 `ap_cstr_stripws()` cases), and the HTTP proxy + uWSGI regression files pass 6/6 including both empty-header cases. `git diff --check` and `py_compile` are clean as well. I updated the PR description with the current results.

notroj:
Thanks, great stuff. I adjusted it here to return the end pointer and do some housekeeping.

https://github.com/apache/httpd/compare/trunk...notroj:httpd:pr749-copy

PR #754, opened 2026-09-11
varbuf: add optional -fbounds-safety annotations for struct ap_varbuf

LaptopsPlural:
## Summary
Secure-by-design memory-safety hardening. Annotates `struct ap_varbuf.buf` with optional Clang `-fbounds-safety` / sized-by macros tied to `avail`. Default builds unchanged (opt-in OFF).

Contributor: Jeff Bindel via `LaptopsPlural`. Not a vulnerability PoC.

## Test plan
- [ ] Default CMake/autotools build
- [ ] Optional bounds-safety ON with supporting Clang (maintainers)


notroj:
Very nice! I think this would be better in `ap_config.h` - toward the end - there are already some compiler-specific attribute handling things there.

LaptopsPlural:
@notroj Thanks — moved the inert macros into `ap_config.h` next to the existing attribute helpers and dropped the standalone `ap_bounds_safety.h`.

notroj:
Thanks!

Not trying to be awkward here but it looks like that LLVM feature is still a WIP? I was looking to see if there is something similar for GCC and found https://gcc.gnu.org/onlinedocs/gcc/Common-Attributes.html#index-counted_005fby - it's implied this would be used/usable under a UBSAN build already? (I wonder if we could try to trip it, we have UBSAN in CI already)

LaptopsPlural:
@notroj Good questions — you’re right on both counts.

Yes: Clang `-fbounds-safety` / `__sized_by*` is still experimental (Apple/LLVM WIP). That’s why this stays **opt-in OFF** by default and the `AP_*` macros in `ap_config.h` are inert unless someone explicitly turns `ENABLE_FBOUNDS_SAFETY` on with a toolchain that has `<ptrcheck.h>`.

GCC’s `counted_by` is the more mature cousin, and you’re also right that UBSAN can check it (so it’s attractive given httpd already has UBSAN in CI). Two practical mismatches for *this* tip though:

1. **Attribute shape.** `ap_varbuf.buf` is annotated as a **byte** bound (`avail + 1`, including the NUL), which maps to Clang `__sized_by_or_null`. GCC `counted_by` is an **element** count. For `char *` those coincide numerically, but it’s a different attribute family than what this PR wires today.

2. **Field order / ABI.** In the public `struct ap_varbuf`, `buf` is declared **before** `avail`. GCC’s `counted_by` wants the count field declared first. Flipping that order would be an ABI break, so we can’t just swap in `__attribute__((counted_by(avail + 1)))` on `buf` without a larger design change.

So I don’t think we can honestly “trip it in UBSAN CI” with a one-line GCC swap on this struct as it stands. What this PR is aiming for is: inert by default, optional Clang path for people with that toolchain, and capacity-before-pointer assigns so the invariants are correct if/when bounds checking is on.

Happy to trim the PR description so it doesn’t oversell the LLVM side, or adjust the macros/docs if you want the GCC/`counted_by` story called out more clearly for future work.

PR #756, opened 2026-09-14
remoteip: avoid pointer underflow on empty header token

mcc0nnell:
`remoteip_modify_request()` trims each `RemoteIPHeader` token by forming
`parse_remote + strlen(parse_remote) - 1` before checking whether the token is
empty. A trailing comma can make the token empty, for example:

```
X-Forwarded-For: 192.0.2.1,
```

In that case the old code forms a pointer one byte before the string object.
ASan invalid-pointer-pair instrumentation catches the old form; the
end-exclusive form in this patch is clean.

Use an end-exclusive pointer and inspect `eos[-1]` only after `eos >
parse_remote`. This keeps the existing literal-space trimming semantics and
avoids forming the invalid pointer.

The regression test exercises a trailing empty `RemoteIPHeader` token through
`mod_remoteip`.

Validation:
- `git diff --check`
- Python test module compiles
- patched `mod_remoteip` builds as a DSO with `apxs`
- throwaway httpd config loads the patched module and returns 200 for the
  trailing-empty-token request
- ASan `detect_invalid_pointer_pairs=2`: old primitive reports an invalid
  pointer pair; new end-exclusive primitive exits cleanly


test/modules/proxy/test_03_response.py
```diff
@@ -106,6 +107,16 @@ def test_proxy_03_004(self, env):
         assert "x-empty" in r.response["header"]
         assert r.response["body"] == b"Hello"
 
+    # a trailing empty RemoteIPHeader token must not underflow the trim pointer
+    def test_proxy_03_005(self, env):
```
notroj, line 111: I just noticed this (and the previous test added here) are not in sorted order in this file, can you reorder please?

notroj (changes requested):
LGTM apart from the test ordering nit.

notroj:
Thank you!

PR #764, opened 2026-09-17
modules/md/md_acme_acct.c (find_acct): Do not dereference a NULL md in the debug log

PavlNekrasov:
Fix: DEREF_AFTER_NULL

Problem:
find_acct() accepts a candidate account when ctx->md is NULL, and dereferences ctx->md in the very next statement to log it. The two lines contradict each other, so the NULL that the condition admits crashes the logging call.
https://github.com/apache/httpd/blob/ba3d76a275fad7559857774e01da57ad1cf92522/modules/md/md_acme_acct.c#L283-L287

Solution:
Log the CA only when there is an md, keeping the NULL that the condition already allows.

Signed-off-by: p.nekrasov@fobos-nt.ru
Signed-off-by: sovtouch@altlinux.org


notroj:
LGTM, thanks.

